PR #10, opened 2020-06-07
Jare

nightcrawlerstudios:
Wrapped Mesh so that user only needs to know the path to the mesh when creating one. 
Added helper methods to assign the mesh's filename to be the name of the mesh.
Added "SetMesh" and "SetMaterial" methods to Model class.


nc/graphics/Mesh.cpp
```diff
@@ -0,0 +1,67 @@
...
+#ifdef _WIN32
+    '\\';
+#else 
+    '/';
+#endif
+
+
+namespace nc::graphics::detail
+{
+    bool HasValidMeshExtension(std::string fileExtension)
```
McCallisterRomer, line 16: compare is const, can make const std::string&

nc/graphics/Mesh.cpp
```diff
@@ -0,0 +1,67 @@
...
+            return true;
+        }
+        if (fileExtension.compare("obj") == 0 || fileExtension.compare("OBJ") == 0)
+        {
+            return true;
+        }
+        return false;
+    }
+
+    std::string GetMeshFileName(std::string meshPath)
```
McCallisterRomer, line 29: same as before

nc/graphics/Mesh.h
```diff
@@ -0,0 +1,50 @@
+#pragma once
+
+#include <d3d11.h>
+#include <string>
+#include <vector>
+#include "graphics/DirectXMath/Inc/DirectXMath.h"
```
McCallisterRomer, line 6: Forward declare XMFLOAT3

PR #28, opened 2020-07-08
Graphics

nightcrawlerstudios:
Added functionality around diffuse texture loading/rendering.


nc/graphics/Mesh.cpp
```diff
@@ -1,7 +1,11 @@
 #include "debug\NcException.h"
 #include "graphics\d3dresource\GraphicsResourceManager.h"
 #include "graphics\Mesh.h"
-#include "utils\objloader\ObjLoader.h"
+#include <assimp\Importer.hpp>
```
McCallisterRomer, line 4: Convention is to use <> for compiler implementation includes, so this should be in quotes.

nc/graphics/Mesh.cpp
```diff
@@ -1,7 +1,11 @@
...
+#include <assimp\scene.h>
+#include <assimp\postprocess.h>
+#include <vector>
```
McCallisterRomer, line 7: I've been trying to separate source/language includes like so:

#include "thing.h"
#include "other.h"

#include <vector>
...
What do you think?

nc/graphics/Mesh.cpp
```diff
@@ -1,7 +1,11 @@
...
+#include "external\include\directx\math\DirectXMath.h"
```
McCallisterRomer, line 8: You should be able to ditch "external\include"

project/Scenes/InitialScene.cpp
```diff
@@ -30,22 +29,19 @@ void InitialScene::Load()
...
-    groundRenderer->SetMaterial(groundMaterial);
+    NCE::AddEngineComponent<Renderer>(groundHandle, groundMesh, groundMaterial);
 
     //Ship
-    auto shipHandle = NCE::CreateEntity(Vector3::Zero(), Vector3::Zero(), Vector3::One(), "Ship");
+    auto shipHandle = NCE::CreateEntity(Vector3::Zero(), Vector3::Zero(), Vector3::One() * 10, "Ship");
     NCE::AddUserComponent<Ship>(shipHandle);
-    auto shipRend = NCE::AddEngineComponent<Renderer>(shipHandle, wormMesh);
-    shipRend->SetMaterial(shipMaterial);
+    NCE::AddEngineComponent<Renderer>(shipHandle, beeMesh, beeMaterial);
```
McCallisterRomer, line 44: slick

PR #47, opened 2020-11-11
Cal/vnext

McCallisterRomer:
Added an Exists method to GraphicsResourceManager so the Mesh constructor can query if the needed buffers exist before invoking AssImp. Vertex was moved to its own file so that including Mesh.h doesn't also include DirectXMath.

### DIFF
--- a/nc/source/graphics/d3dresource/GraphicsResource.cpp
+++ b/nc/source/graphics/d3dresource/GraphicsResource.cpp
@@ -1,5 +1,6 @@
 #include "GraphicsResource.h"
 #include "graphics/Graphics.h"
+#include "graphics/Model.h"
 #include "graphics/WICTextureLoader.h"
 #include "directx/math/DirectXMath.h"
 #include "component/Transform.h"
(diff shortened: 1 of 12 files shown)


nc/source/graphics/d3dresource/GraphicsResource.cpp
```diff
@@ -1,5 +1,6 @@
 #include "GraphicsResource.h"
 #include "graphics/Graphics.h"
+#include "graphics/Model.h"
```
nightcrawlerstudios, line 3: Why does GraphicsResource have to know about Model? That's pretty wonky we should redesign if this needs to be here

nc/source/graphics/d3dresource/GraphicsResource.h
```diff
@@ -2,17 +2,21 @@
...
 #include <stdint.h>
+#include <memory>
 
-namespace nc::graphics { class  Graphics; }
-namespace DirectX      { struct XMMATRIX; }
-
+namespace nc::graphics
+{
+    class Graphics;
+    class Model;
```
nightcrawlerstudios, line 17: GraphicsResource should not know about Model

nightcrawlerstudios (approved):

PR #52, opened 2020-11-17
Cal/on resize subscribing

McCallisterRomer:
-Added ability to subscribe to resizing through IOnResizeReceiver
-Added IUIFixed and IUIFixedElement


project/source/ui/AddServerSelectableUIElement.cpp
```diff
@@ -1,10 +1,15 @@
 #include "AddServerSelectableUIElement.h"
 
+namespace
+{
+    const auto ELEMENT_SIZE = ImVec2{ 300, 100 };
```
nightcrawlerstudios, line 5: Can this be a property instead of a free const like this?

nightcrawlerstudios (approved):

PR #59, opened 2020-12-04
Cal/concepts

McCallisterRomer:
-Replaced usage of enable_if with concepts.
-Fixed Assimp deprecated-copy warnings(it uses implicitly-defined copy assignment for types with user-defined copy constructors)


nightcrawlerstudios (approved):
neat-o!!!

PR #60, opened 2020-12-09
Cal/const correctness

McCallisterRomer:
-added a few missing consts
-updated build scripts to use ninja and handle different build types


nightcrawlerstudios (approved):
Nice!

PR #62, opened 2020-12-22
shared ptrs to unique

McCallisterRomer:
-GfxResManager creates unique_ptrs and returns non-owning * from Acquire instead of shared_ptr.
-All AcquireUnique methods removed from GraphicsResource.h and replaced with calls to make_unique. This seems simpler and clearly indicates that the caller is responsible for the resource.

### DIFF
--- a/nc/source/component/PointLightManager.cpp
+++ b/nc/source/component/PointLightManager.cpp
@@ -8,6 +8,13 @@ namespace nc
 {
     const uint32_t PointLightManager::MAX_POINT_LIGHTS;
 
+    PointLightManager::PointLightManager()
+        : m_currentIndex{0u},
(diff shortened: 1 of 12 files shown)


nc/source/component/PointLightManager.cpp
```diff
@@ -27,9 +34,7 @@ namespace nc
 
     void PointLightManager::Bind()
     {
-        using namespace nc::graphics::d3dresource;
-
-        auto pointLightArray = PixelConstantBuffer<PointLightsArrayCBuf>::AcquireUnique(PointLightManager::m_pointLightsArrayConstBufData, 0u);
```
nightcrawlerstudios, line 32: Nice catch

nightcrawlerstudios (approved):

PR #63, opened 2020-12-31
Reorganized GraphicsResources into separate classes and added ability to edit Technique properties.

nightcrawlerstudios:


nc/source/graphics/techniques/PhongShadingTechnique.h
```diff
@@ -3,6 +3,7 @@
 #include "graphics/techniques/Technique.h"
 #include "graphics/techniques/TechniqueType.h"
 #include "graphics/materials/MaterialProperties.h"
+#include "graphics/d3dresource/ConstantBufferResources.h"
```
McCallisterRomer, line 6: PixelConstantBuffer is forward declared in this file. Can this include be in the cpp? It may require the default destructor in cpp trick.

nc/source/graphics/techniques/TechniqueManager.h
```diff
@@ -2,6 +2,7 @@
 
 #include "graphics/techniques/Technique.h"
 #include "graphics/techniques/PhongShadingTechnique.h"
+#include "graphics/techniques/WireframeTechnique.h"
```
McCallisterRomer, line 5: It looks like this doesn't need PhongShadingTechnique.h or WireframeTechnique.h.

McCallisterRomer (approved):

PR #72, opened 2021-01-30
Cal/telemetry

McCallisterRomer:
-Added function profiling
-Removed some methods from Editor and moved most imgui code to another file

### DIFF
--- a/nc/source/component/Transform.h
+++ b/nc/source/component/Transform.h
@@ -25,7 +25,7 @@ namespace nc
             Quaternion GetRotation() const;
             Vector3 GetScale() const;
 
-            DirectX::XMMATRIX GetTransformationMatrix() const;
+            DirectX::FXMMATRIX GetTransformationMatrix() const;
             DirectX::XMMATRIX GetTransformationMatrixEx(Vector3 additionalScale) const;
             DirectX::XMMATRIX GetViewMatrix() const;
 
--- a/nc/source/debug/Profile.cpp
+++ b/nc/source/debug/Profile.cpp
@@ -1,0 +1,85 @@
+#ifdef NC_EDITOR_ENABLED
+
+#include "Profile.h"
+
+#include <algorithm>
+#include <numeric>
+#include <stdexcept>
+#include <vector>
+
+namespace nc::debug::profile
+{
+    namespace internal
+    {
+        FuncId currentId = 0u;
+        ProfileData data{};
+        std::vector<FuncId> openProfiles{};
+    }
+
+    FunctionMetrics::FunctionMetrics(std::string name, Filter profileFilter)
+        : functionName{std::move(name)},
+          callHistory{historySize},
+          timeHistory{historySize},
+          timer{},
+          callCount{0u},
+          execTime{0.0f},
+          filter{profileFilter},
+          active{false}
+    {
+    }
+
+    void FunctionMetrics::UpdateHistory()
+    {
+        std::shift_left(callHistory.begin(), callHistory.end(), 1);
+        callHistory.back() = static_cast<float>(callCount);
+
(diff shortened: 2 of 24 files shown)


nc/source/debug/Profile.cpp
```diff
@@ -0,0 +1,118 @@
...
+    void Push(FuncId id)
+    {
+        internal::openProfiles.push_back(id);
+        try
+        {
+            auto& data = internal::data.at(id);
+            ++data.callCount;
+            data.timer.Start();
+        }
+        catch(...)
```
McCallisterRomer: Can catch std::exception here instead of ...

nc/source/ui/editor/EditorControls.h
```diff
@@ -0,0 +1,212 @@
...
+            ImGui::Text("Scene Graph");
+            ImGui::Spacing();
+            filter.Draw("##filterlabel", 2.0f * DefaultItemWidth);
+            ImGui::Separator();
+            ImGui::Separator();
+            ImGui::Spacing();
+
+            if(ImGui::BeginChild("EntityList"))
+            {
+                auto uidScope = 0u;
```
McCallisterRomer: Should be possible to use handle as the id here.

jaremieromer:
Can we change Profile.h to Profiler.h? Even knowing what this pull request was for I was confused by the contents of Profile (maybe because the filter values are User and Engine) and didn't click that it was a profiler until I looked at how it was being used.

nc/source/component/Transform.h
```diff
@@ -25,7 +25,7 @@ namespace nc
             Quaternion GetRotation() const;
             Vector3 GetScale() const;
 
-            DirectX::XMMATRIX GetTransformationMatrix() const;
+            DirectX::FXMMATRIX GetTransformationMatrix() const;
```
jaremieromer, line 28: Why FXMMATRIX?

nc/source/ui/editor/Editor.cpp
```diff
@@ -1,223 +1,67 @@
...
-    const auto MainWindowFlags = ImGuiWindowFlags_NoBackground |
+    constexpr auto MainWindowFlags = ImGuiWindowFlags_NoBackground |
                                  ImGuiWindowFlags_MenuBar |
                                  ImGuiWindowFlags_NoResize |
                                  ImGuiWindowFlags_NoScrollbar |
                                  ImGuiWindowFlags_NoMove;
-
-    void DrawComponentControl(nc::ComponentBase* comp)
+    
+    namespace hotkey
```
jaremieromer, line 16: This is nice

nc/source/ui/editor/EditorControls.h
```diff
@@ -0,0 +1,212 @@
...
+#ifdef NC_EDITOR_ENABLED
+#include "Profile.h"
+#include "Ecs.h"
+#include "graphics/d3dresource/GraphicsResourceManager.h"
+#include "imgui/imgui.h"
+
+
+namespace nc::ui::editor::controls
+{
+    const auto TitleBarHeight = 40.0f;
```
jaremieromer, line 11: Not sure if worth it for IMGUI, but we should eventually not hardcode these sizes.

jaremieromer (approved):

nc/source/component/Transform.h
McCallisterRomer: It is a typedef for const & or const depending on platform.

nc/source/ui/editor/EditorControls.h
McCallisterRomer: Agreed. Once we've settled on a permanent solution, we'll figure out a proper way to do layout and scaling.

PR #73, opened 2021-02-03
Cal/api

McCallisterRomer:
-Moved all headers required to use the engine to include directory
-Changed api functions from static members to free functions


nc/include/Entity.h
```diff
@@ -43,7 +43,7 @@ namespace nc
             template<std::derived_from<Component> T>
             bool RemoveUserComponent() noexcept;
 
-            const std::vector<std::unique_ptr<Component>> & GetUserComponents() const noexcept;
+            const std::vector<std::unique_ptr<Component>>& GetUserComponents() const noexcept;
```
McCallisterRomer: It feels wrong to have unique_ptr in the return type here. Since the performance of this method is not important, I'm thinking of constructing a vector<Component*> to return on each call.

nc/include/MainCamera.h
```diff
@@ -1,26 +1,17 @@
...
+    /** Get the transform of the current main camera */
+    [[nodiscard]] Transform* GetMainCameraTransform();
 
-        private:
-            friend class ::nc::engine::ApiBinder;
-            static std::function<void(Camera*)> Set_;
-            static std::function<Transform*()> GetTransform_;
-    };
+    /** @note used internally - move somewhere else? */
+    void ClearMainCamera();
```
McCallisterRomer: Forgot to move this. Will do before merge.

nc/include/graphics/Material.h
```diff
@@ -31,8 +31,8 @@ namespace nc::graphics
     };
 
     template<>
-    Material Material::CreateMaterial<TechniqueType::PhongShadingTechnique>(const std::vector<std::string>& texturePaths, MaterialProperties& materialProperties);
+    Material Material::CreateMaterial<TechniqueType::PhongShading>(const std::vector<std::string>& texturePaths, MaterialProperties& materialProperties);
```
jaremieromer, line 34: This is better without Technique at the end. nice

jaremieromer (approved):
I think this is great. My only suggestion if we do keep this organization (and I think we should) is to name folders that implement sub-'modules' of a specific header the same as the header. The only change here would be /net -> /network to correspond with Network.h

PR #77, opened 2021-02-07
Initial Vulkan implementation.

jaremieromer:
Initial Vulkan implementation, prior to graphics pipeline.
Added a USE_VULKAN flag to CMAKE/build script and integrated it for conditionally building against Vulkan or DirectX.
To build, must have Vulkan SDK installed and build with ninja -C .\build\Engine\Debug-Vulkan.


nc/source/graphics/vulkan/Device.h
```diff
@@ -0,0 +1,82 @@
...
+#include "math/Vector2.h"
+
+#define VK_USE_PLATFORM_WIN32_KHR
+#include "vulkan/vulkan.hpp"
+
+namespace nc::graphics::vulkan
+{
+    class Instance;
+
+    static const std::vector<const char*> m_deviceExtensions = { VK_KHR_SWAPCHAIN_EXTENSION_NAME };
```
McCallisterRomer: I don't think this is what you want to do here. A static declaration at namespace scope specifies internal linkage or "local to this translation unit." You probably want this in the cpp, not the header. On a style note, the rest of the codebase uses static only for members and block scope variables. Anonymous namespaces are used for internal linkage to reduce some of the overloading of the static keyword(also they can have type declarations).

Somewhere in Device.cpp
namespace
{
    const auto DeviceExtensions = std::vector<const char*>{ VK_KHR_SWAPCHAIN_EXTENSION_NAME };
}

nc/source/graphics/vulkan/Instance.h
```diff
@@ -0,0 +1,33 @@
+#pragma once
+
+#include <vector>
+
+#define VK_USE_PLATFORM_WIN32_KHR
+#include "vulkan/vulkan.hpp"
+
+namespace nc::graphics::vulkan
+{
+    static const std::vector<const char*> m_globalExtensions = { VK_KHR_WIN32_SURFACE_EXTENSION_NAME, VK_KHR_SURFACE_EXTENSION_NAME };
```
McCallisterRomer: See Device.h

nc/source/graphics/vulkan/Instance.h
```diff
@@ -0,0 +1,33 @@
...
+        public:
+            Instance(HWND hwnd, HINSTANCE hinstance);
+            const vk::SurfaceKHR* GetSurface() const noexcept;
+            const vk::Instance* GetInstance() const noexcept;
+
+        private:
+            vk::UniqueInstance m_instance;
+            vk::UniqueSurfaceKHR m_surface;
+
+            bool CheckValidationLayerSupport();
```
McCallisterRomer: These 3 member functions are not dependent on Instance's state, so they should not be members. Given that they are also private, they don't need to be declared in the header at all. I would put them in an anonymous namespace in the cpp along with the variables at the top of this file.

nc/source/graphics/vulkan/Device.h
```diff
@@ -0,0 +1,82 @@
...
+            {
+                throw std::runtime_error("QueueFamilyIndices::GetQueueFamilyIndex() - QueueFamilies incomplete.");
+            }
+
+            switch (type)
+            {
+                case QueueFamilyType::GraphicsFamily:
+                    return graphicsFamily.value();
+                case QueueFamilyType::PresentFamily:
+                    if (isSeparatePresentQueue)
```
McCallisterRomer: I slightly prefer a ternary here; your choice.

nc/source/graphics/vulkan/Device.h
```diff
@@ -0,0 +1,82 @@
...
+
+    static const std::vector<const char*> m_deviceExtensions = { VK_KHR_SWAPCHAIN_EXTENSION_NAME };
+
+    enum class QueueFamilyType : uint8_t
+    {
+        GraphicsFamily,
+        PresentFamily
+    };
+
+    struct QueueFamilyIndices
```
McCallisterRomer: see discussion

nc/source/graphics/vulkan/Device.cpp
```diff
@@ -0,0 +1,369 @@
...
+
+            if (indices.IsComplete())
+            {
+                break;
+            }
+
+            i++;
+        }
+
+        indices.isSeparatePresentQueue = (indices.presentFamily != indices.graphicsFamily || !indices.presentFamily.has_value()) ? true : false;
```
McCallisterRomer: Check logic here - equals true when there is no presentFamily value.

nc/source/graphics/vulkan/Device.cpp
```diff
@@ -0,0 +1,369 @@
...
+            {
+                indices.presentFamily = i;
+            }
+
+            if (indices.IsComplete())
+            {
+                break;
+            }
+
+            i++;
```
McCallisterRomer: When one uses pre-increment, regarding temporaries, one mustn't lament.

nc/source/graphics/vulkan/Device.h
```diff
@@ -0,0 +1,82 @@
...
+            vk::Queue m_graphicsQueue;
+            vk::Queue m_presentQueue;
+            vk::SwapchainKHR m_swapChain;
+            std::vector<vk::Image> m_swapChainImages;
+            vk::Format m_swapChainImageFormat;
+            vk::Extent2D m_swapChainExtent;
+            std::vector<vk::ImageView> m_swapChainImageViews;
+            Vector2 m_dimensions;
+
+            SwapChainSupportDetails QuerySwapChainSupport(const vk::PhysicalDevice& device, const vk::SurfaceKHR* surface);
```
McCallisterRomer: Another free function candidate.

nc/source/graphics/vulkan/Device.cpp
```diff
@@ -0,0 +1,369 @@
...
+                        break;
+                    }
+                }
+
+                if (!extensionFound)
+                {
+                    extensionsSupported = false;
+                }
+            }
+            extensionsSupported = true;
```
McCallisterRomer: Left over?

nc/source/graphics/vulkan/Device.cpp
```diff
@@ -0,0 +1,369 @@
...
+            }
+
+            bool extensionsSupported;
+            std::vector<vk::ExtensionProperties> availableExtensions(extensionCount);
+            if (device.enumerateDeviceExtensionProperties(nullptr, &extensionCount, availableExtensions.data()) != vk::Result::eSuccess)
+            {
+                throw std::runtime_error("Device::IsPhysicalDeviceExtensionSupported() - could not enumerate device extensions.");
+            }
+
+            for (const char* extensionName : m_deviceExtensions)
```
McCallisterRomer: Setting extensionsSupported would be more succinct with a nested all_of + any_of. 
Pseudo: result = all_of(extensions, any_of(available, return strcmp(...)))

nc/source/graphics/vulkan/Device.cpp
```diff
@@ -0,0 +1,369 @@
...
+
+        uniqueQueueFamilies.emplace(indices.graphicsFamily.value());
+
+        if (indices.isSeparatePresentQueue)
+        {
+            uniqueQueueFamilies.emplace(indices.presentFamily.value());
+        }
+
+        float queuePriority = 1.0f;
+        for (auto queueFamily : uniqueQueueFamilies)
```
McCallisterRomer: std::transform is the idiomatic way to do this.

nc/source/graphics/vulkan/Device.cpp
```diff
@@ -0,0 +1,369 @@
...
+        deviceCreateInfo.setPEnabledFeatures(nullptr);
+        deviceCreateInfo.setEnabledExtensionCount(static_cast<uint32_t>(m_deviceExtensions.size()));
+        deviceCreateInfo.setPpEnabledExtensionNames(m_deviceExtensions.data());
+        deviceCreateInfo.setEnabledLayerCount(0);
+
+        try
+        {
+            m_device = m_physicalDevice.createDevice(deviceCreateInfo);
+        }
+        catch (std::exception& error)
```
McCallisterRomer: catch by const ref

nc/source/graphics/vulkan/Device.cpp
```diff
@@ -0,0 +1,369 @@
...
+        {
+            std::string message = "Device::Create() - Failed to create device: ";
+            message.append(error.what());
+            throw std::runtime_error(message);
```
McCallisterRomer: throw std::throw_with_nested(std::runtime_error("..."));

nc/source/graphics/vulkan/Device.cpp
```diff
@@ -0,0 +1,369 @@
...
+        }
+
+        /**************
+         * SWAP CHAIN *
+         * ************/
+        auto swapChainSupport = Device::QuerySwapChainSupport(m_physicalDevice, vkSurface);
+
+        auto surfaceFormat = swapChainSupport.formats[0];
+        for (const auto& availableFormat : swapChainSupport.formats)
```
McCallisterRomer, line 151: Fingers crossed we get a FirstOrDefault in c++23 - can't keep giving wins to Linq! If this is a super common pattern, I wouldn't mind making a wrapper around find_if for that purpose.

nc/source/graphics/vulkan/Device.cpp
```diff
@@ -0,0 +1,369 @@
...
+        {
+            // https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkPresentModeKHR.html
+            if (availablePresentMode == vk::PresentModeKHR::eMailbox)
+            {
+                presentMode = availablePresentMode;
+                break;
+            }
+        }
+
+        vk::Extent2D extent;
```
McCallisterRomer, line 171: I wouldn't worry about it here, but I usually try to handle these conditionally initialized variables with a lamda. Also, who uses UINT32_MAX when you could run out the clock typing std::numeric_limits<uint32_t>::max()? We live in a society.

nc/source/graphics/vulkan/Device.cpp
```diff
@@ -0,0 +1,369 @@
...
+#include <string>
+
+namespace nc::graphics::vulkan
+{
+    Device::Device(const vulkan::Instance* instance, Vector2 dimensions)
+    : m_device{},
+      m_physicalDevice{},
+      m_graphicsQueue{},
+      m_presentQueue{},
+      m_swapChain{},
```
McCallisterRomer, line 28: missing m_swapChainImages

nc/source/graphics/vulkan/Device.cpp
```diff
@@ -0,0 +1,369 @@
...
+
+        // Set swapchain images        
+        uint32_t swapChainImageCount;
+        auto countResult = m_device.getSwapchainImagesKHR(m_swapChain, &swapChainImageCount, nullptr);
+        if (countResult != vk::Result::eSuccess)
+        {
+            throw std::runtime_error("Error getting swapchain images count.");
+        }
+
+        auto result = m_device.getSwapchainImagesKHR(m_swapChain, &swapChainImageCount, m_swapChainImages.data());
```
McCallisterRomer: It doesn't look like there is any reason to store result (or countResult above) - just put the calls in the if like you did with devices and extension properties.

It also looks like some UB here. m_swapChainImages is never initialized(see previous comment), so passing the .data() ptr is dangerous. Even with it default initialized though, you'd still be passing nullptr. My bet is that you want to do: resize(swapChainImageCount).

nc/source/graphics/vulkan/Device.cpp
```diff
@@ -0,0 +1,369 @@
...
+        swapChainComponents.setA(vk::ComponentSwizzle::eIdentity);
+
+        auto swapChainSubresourceRange = vk::ImageSubresourceRange{};
+        swapChainSubresourceRange.setAspectMask(vk::ImageAspectFlagBits::eColor);
+        swapChainSubresourceRange.setBaseMipLevel(0);
+        swapChainSubresourceRange.setLevelCount(1);
+        swapChainSubresourceRange.setBaseArrayLayer(0);
+        swapChainSubresourceRange.setLayerCount(1);
+
+        for (size_t i = 0; i < m_swapChainImages.size(); i++)
```
McCallisterRomer: Intentions are lacking, muddled, bereft; if only the increment had gone on the left.

nc/source/graphics/vulkan/Instance.cpp
```diff
@@ -0,0 +1,109 @@
...
+        vk::InstanceCreateInfo instanceCreateInfo ( {}, &applicationInfo );
+
+        EnableValidationLayers(instanceCreateInfo);
+        SetGlobalExtensions(instanceCreateInfo);
+
+        try
+        {
+            m_instance = createInstanceUnique(instanceCreateInfo);
+        }
+        catch (std::exception& error)
```
McCallisterRomer: See previous - catch by const&, and rethrow with std::throw_with_nested

nc/source/graphics/vulkan/Instance.cpp
```diff
@@ -0,0 +1,109 @@
...
+        else 
+        {
+            instanceCreateInfo.enabledLayerCount = 0;
+        }
+    }
+
+    bool Instance::CheckValidationLayerSupport()
+    {
+        uint32_t layerCount;
+        auto layerResult = vk::enumerateInstanceLayerProperties(&layerCount, nullptr);
```
McCallisterRomer: layerResult and layerDataResult don't need to be stored

nc/source/graphics/vulkan/Instance.cpp
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        std::vector<vk::LayerProperties> availableLayers(layerCount);
+        auto layerDataResult = vk::enumerateInstanceLayerProperties(&layerCount, availableLayers.data());
+        if (layerDataResult != vk::Result::eSuccess)
+        {
+            return false;
+        }
+
+        for (const char* layerName : m_validationLayers)
```
McCallisterRomer: See previous comment about all_of->any_of

McCallisterRomer (approved):
That's a lotta code, nice job! Can't wait to see a triangle!

nc/source/graphics/vulkan/Device.h
McCallisterRomer: Github seems to be highlighting the case statements - I meant inside the PresentFamily case.

nc/source/graphics/vulkan/Device.h
jaremieromer: Done, thanks

nc/source/graphics/vulkan/Instance.h
jaremieromer: Done

PR #81, opened 2021-02-12
point light ctor takes properties

McCallisterRomer:
-PointLight constructor (and AddComponent) can take in PointLight::Properties rather than having to call Set after construction.
-Added a few CreateEntity overloads
-Fixed Worms scene only creating 1 worm
-Minor organization+comments in sample scenes

### DIFF
--- a/nc/source/ecs/EcsApiImpl.cpp
+++ b/nc/source/ecs/EcsApiImpl.cpp
@@ -19,9 +19,19 @@ namespace nc
         return CreateEntity(Vector3::Zero(), Quaternion::Identity(), Vector3::One(), internal::DefaultEntityTag);
     }
 
+    EntityHandle CreateEntity(Vector3 pos)
+    {
+        return CreateEntity(pos, Quaternion::Identity(), Vector3::One(), internal::DefaultEntityTag);
+    }
+
     EntityHandle CreateEntity(std::string tag)
     {
         return CreateEntity(Vector3::Zero(), Quaternion::Identity(), Vector3::One(), std::move(tag));
+    }
+
+    EntityHandle CreateEntity(Vector3 pos, std::string tag)
+    {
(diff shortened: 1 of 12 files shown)


jaremieromer (approved):
Do do do doooooo!

nc/source/ecs/EcsApiImpl.cpp
```diff
@@ -19,11 +19,21 @@ namespace nc
         return CreateEntity(Vector3::Zero(), Quaternion::Identity(), Vector3::One(), internal::DefaultEntityTag);
     }
 
+    EntityHandle CreateEntity(Vector3 pos)
+    {
+        return CreateEntity(pos, Quaternion::Identity(), Vector3::One(), internal::DefaultEntityTag);
```
jaremieromer, line 24: Doesn't strictly pertain to this review, and Github does not let me leave review comments on unmodified code, but it would be a fun side project to create a little helper function that generates a friendly default name per Entity type. (ie, internal::DefaultEntityTag -> Point Light)

project/source/spawn_test/SpawnTest.cpp
```diff
@@ -39,11 +40,10 @@ namespace nc::sample
             .attConst = 1.0f,
             .attLin = 0.0001f
         };
-        auto lvHandle = CreateEntity(Vector3::Zero(), Quaternion::Identity(), Vector3::One(), "Point Light");
-        auto pointLight = AddComponent<PointLight>(lvHandle);
-        pointLight->Set(lightProperties);
 
-        // Collider that destroys anything leaving its bounded area
+        AddComponent<PointLight>(CreateEntity("Point Light"), lightProperties);
```
jaremieromer, line 44: So much code was condensed and removed in this review. Nice

jaremieromer (approved):

PR #82, opened 2021-02-14
Cal/more colliders

McCallisterRomer:
-Created CollisionSystem - no new functionality, but this will make more sense once additional data structures need to be managed
-Created EntityInfo  for Entity initialization
-Added IsStatic to Entity (not used yet)
-Moved all entity types from /include to /include/entity

### DIFF
--- a/nc/source/ecs/EcsApiImpl.cpp
+++ b/nc/source/ecs/EcsApiImpl.cpp
@@ -14,20 +14,10 @@ namespace nc
         }
     }
 
-    EntityHandle CreateEntity()
+    EntityHandle CreateEntity(EntityInfo info)
     {
-        return CreateEntity(Vector3::Zero(), Quaternion::Identity(), Vector3::One(), internal::DefaultEntityTag);
-    }
-
-    EntityHandle CreateEntity(std::string tag)
-    {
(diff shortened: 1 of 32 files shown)


nc/source/ecs/EcsApiImpl.cpp
```diff
@@ -14,20 +14,10 @@ namespace nc
...
-    EntityHandle CreateEntity(std::string tag)
-    {
-        return CreateEntity(Vector3::Zero(), Quaternion::Identity(), Vector3::One(), std::move(tag));
-    }
-
-    EntityHandle CreateEntity(Vector3 pos, Quaternion rot, Vector3 scale, std::string tag)
-    {
-        IF_THROW(scale == Vector3::Zero(), "CreateEntity - scale cannot be Vector3::Zero");
-        return internal::g_impl->CreateEntity(pos, rot, scale, std::move(tag));
+        IF_THROW(info.scale == Vector3::Zero(), "CreateEntity - scale cannot be Vector3::Zero");
```
jaremieromer, line 19: Not sure about best practices with structs, but I would like this type checking to be done at struct creation time, so we aren't passing around potentially invalid structs. Maybe a CTOR or explicit setter?

jaremieromer (approved):

PR #85, opened 2021-02-16
Completed second half of initial Vulkan implementation. Rendering a triangle to the screen.

jaremieromer:
Added ability to render triangle to the screen.
Added swapchain recreation on resize.


nc/source/graphics/vulkan/Commands.cpp
```diff
@@ -0,0 +1,76 @@
...
+      m_presentReadySemaphores{presentReadySemaphores},
+      m_framesInFlightFences{framesInFlightFences},
+      m_imagesInFlightFences{imagesInFlightFences}
+    {
+        // Create the command buffers.
+        m_commandBuffers.resize(device->GetSwapChainImageViews()->size()); // Need to have one command buffer per frame buffer, which have the same count as the image views.
+        vk::CommandBufferAllocateInfo allocInfo{};
+        allocInfo.setCommandPool(*(device->GetCommandPool()));
+        allocInfo.setLevel(vk::CommandBufferLevel::ePrimary); // Primary means the command buffer can be submitted to a queue for execution, but not called from other command buffers. Alternative is Secondary, which cant be submitted directly but can be called from other primary command buffers.
+        allocInfo.setCommandBufferCount((uint32_t)m_commandBuffers.size());
```
McCallisterRomer: Does this not give a narrowing warning?

nc/source/graphics/vulkan/Commands.cpp
```diff
@@ -0,0 +1,76 @@
...
+        m_commandBuffers = device->GetDevice()->allocateCommandBuffers(allocInfo);
+    }
+
+    void Commands::RecordRenderPass(const vulkan::Device* device, const vulkan::RenderPass* renderPass, const vulkan::FrameBuffers* frameBuffers, const vulkan::GraphicsPipeline* pipeline)
+    {
+        vk::ClearValue const clearValues[1] = { vk::ClearColorValue(std::array<float, 4>({{0.2f, 0.2f, 0.2f, 0.2f}})) };
```
McCallisterRomer: west const is best const

nc/source/graphics/vulkan/Commands.h
```diff
@@ -0,0 +1,30 @@
...
+
+namespace nc::graphics::vulkan
+{
+    class Device; class RenderPass; class FrameBuffers; class GraphicsPipeline;
+
+    class Commands
+    {
+        public:
+            Commands(const vulkan::Device* device, 
+                     const std::vector<vk::Semaphore>* renderReadySemaphores,
```
McCallisterRomer: Unless there is a reason I'm overlooking, these vectors would be better as const&. It would get rid of the strange indexing into the semaphores in the cpp.

nc/source/graphics/vulkan/FrameBuffers.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+            vk::FramebufferCreateInfo framebufferInfo{};
+            framebufferInfo.setRenderPass(*(renderpass->GetRenderPass()));
+            framebufferInfo.setAttachmentCount(1);
+            framebufferInfo.setPAttachments(attachments);
+            framebufferInfo.setWidth(device->GetSwapChainExtentDimensions().x);
+            framebufferInfo.setHeight(device->GetSwapChainExtentDimensions().y);
+            framebufferInfo.setLayers(1);
+
+            m_framebuffers[i] = std::move(device->GetDevice()->createFramebufferUnique(framebufferInfo));
```
McCallisterRomer: No need to std::move on a temporary.

nc/source/graphics/vulkan/FrameBuffers.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+        for (size_t i = 0; i < swapChainImageViewsCount; i++)
+        {
+            vk::ImageView attachments[] = { device->GetSwapChainImageViews()->at(i) };
+
+            vk::FramebufferCreateInfo framebufferInfo{};
+            framebufferInfo.setRenderPass(*(renderpass->GetRenderPass()));
+            framebufferInfo.setAttachmentCount(1);
+            framebufferInfo.setPAttachments(attachments);
+            framebufferInfo.setWidth(device->GetSwapChainExtentDimensions().x);
```
McCallisterRomer: Save the return value from GetSwapChainExtentDimensions so you don't have to call it twice.

nc/source/graphics/vulkan/GraphicsPipeline.cpp
```diff
@@ -0,0 +1,207 @@
...
+#include "RenderPass.h"
+#include "config/Config.h"
+
+#include <vector>
+#include <string>
+#include <fstream>
+
+namespace
+{
+    static std::vector<char> ReadShader(const std::string& filename)
```
McCallisterRomer: anonymous namespace is already internal linkage - no static

nc/source/graphics/vulkan/GraphicsPipeline.cpp
```diff
@@ -0,0 +1,207 @@
...
+    {
+        std::ifstream file(filename, std::ios::ate | std::ios::binary);
+        if (!file.is_open())
+        {
+            throw std::runtime_error("Failed to open file.");
+        }
+        auto fileSize = (size_t)file.tellg();
```
McCallisterRomer: I would highly suggest getting in the habbit of using c++ style casts (static_cast, etc.). The C way attempts different types of casts until something succeeds.

nc/source/graphics/vulkan/GraphicsPipeline.cpp
```diff
@@ -0,0 +1,207 @@
...
+        vertexInputInfo.setPVertexBindingDescriptions(nullptr);
+        vertexInputInfo.setVertexAttributeDescriptionCount(0);
+        vertexInputInfo.setPVertexAttributeDescriptions(nullptr);
+
+        /******************
+         * INPUT ASSEMBLY *
+         * ****************/
+        // @todo: Look into using element buffers here to resuse vertices
+        vk::PipelineInputAssemblyStateCreateInfo inputAssembly{};
+        inputAssembly.setTopology(vk::PrimitiveTopology::eTriangleList);
```
McCallisterRomer, line 86: A familiar land!

nc/source/graphics/vulkan/GraphicsPipeline.cpp
```diff
@@ -0,0 +1,207 @@
...
+    const vk::Pipeline* GraphicsPipeline::GetPipeline() const
+    {
+         return &(m_pipeline.get());
+    }
+
+    vk::ShaderModule GraphicsPipeline::CreateShaderModule(const std::vector<char>& code, const vulkan::Device* device)
+    {
+        vk::ShaderModuleCreateInfo createInfo{};
+        createInfo.setCodeSize(code.size());
+        createInfo.setPCode(reinterpret_cast<const uint32_t*>(code.data()));
```
McCallisterRomer: We need to look into the safety of this. uint32_t has stricter alignment requirements than char, so typically this cast is unsafe, but I feel like vector has to allocate on a word boundary.

nc/source/graphics/Graphics2.h
```diff
@@ -9,37 +9,52 @@
...
-            
-            void FrameBegin();
-            void DrawIndexed(UINT count);
-            void FrameEnd();
-
-            #ifdef NC_EDITOR_ENABLED
-            uint32_t GetDrawCallCount() const;
-            #endif
-            
+        Graphics2(HWND hwnd, HINSTANCE hinstance, Vector2 dimensions);
```
McCallisterRomer: Where have all the tabs gone?

nc/source/graphics/Graphics2.cpp
```diff
@@ -1,24 +1,70 @@
...
+        m_renderPass{ nullptr },
+        m_frameBuffers{ nullptr },
+        m_commands{ nullptr },
+        m_dimensions{ dimensions },
+        m_isMinimized{ false },
+        m_isFullscreen{ false },
+        m_viewMatrix{},
+        m_projectionMatrix{}
     {
         m_instance = std::make_unique<Instance>(hwnd, hinstance);
```
McCallisterRomer: This stuff can be set in the member initializer list

nc/source/graphics/vulkan/Commands.cpp
McCallisterRomer: It does not. New learnin everyday.

nc/source/graphics/vulkan/GraphicsPipeline.cpp
```diff
@@ -0,0 +1,207 @@
...
+        std::ifstream file(filename, std::ios::ate | std::ios::binary);
+        if (!file.is_open())
+        {
+            throw std::runtime_error("Failed to open file.");
+        }
+        auto fileSize = (size_t)file.tellg();
+        std::vector<char> buffer(fileSize);
+        file.seekg(0);
+        file.read(buffer.data(), fileSize);
+        file.close();
```
McCallisterRomer, line 36: Good practice to check fail and bad bits with file.fail()

McCallisterRomer (approved):
Would ya look at that triangle

nc/source/graphics/vulkan/Commands.cpp
McCallisterRomer: Cracked the case! We don't pass -Wconversion in CMAKE

PR #86, opened 2021-02-18
Cal/broad phase collision

McCallisterRomer:
-Management of Colliders is done through a new class, ColliderSystem, which is owned by CollisionSystem. There will be more significant changes here in the future, but for now there are just two instances of ComponentSystem<Collider> - for static and dynamic Colliders.
-Added position offset to Collider
-Added a broad phase pass to collision detection using spheres as bounding volume estimates
-Temporarily removed properties from collider widget - will add back once changes to design settle down


jaremieromer (approved):
Hooray!!!!!!!!!!!!

PR #87, opened 2021-02-20
Struct of Arrays Collider Representation

McCallisterRomer:
-Added ColliderInfo for initializing Colliders
-Changed Collider to maintain Collider objects through a ComponentSystem but keep collision-specific data separately.
-General CollisionSystem tinkering


nc/include/component/Collider.h
```diff
@@ -50,11 +57,12 @@ namespace nc
             #endif
 
         protected:
```
jaremieromer: Why final and protected?

nc/source/component/Collider.cpp
```diff
@@ -7,29 +7,37 @@
...
+}
+#endif
+
 namespace nc
 {
     #ifdef NC_EDITOR_ENABLED
-    Collider::Collider(EntityHandle handle, ColliderType type, Vector3 offset, Vector3 scale)
+    Collider::Collider(EntityHandle handle, ColliderInfo info)
         : ComponentBase(handle),
+          m_type{info.type},
```
jaremieromer, line 25: Curious about this: shouldn't the Collider have a ColliderInfo as well? (Rather than decomposed ColliderInfo members?)

nc/source/component/Collider.cpp
```diff
@@ -7,29 +7,37 @@
...
-          m_boundingVolume{collider_detail::CreateBoundingVolume(type, offset, scale)},
-          m_type{type},
-          m_widgetModel{collider_detail::CreateWireframeModel(type)},
+          m_boundingVolume{collider_detail::CreateBoundingVolume(info.type, info.offset, info.scale)},
+          m_widgetModel{collider_detail::CreateWireframeModel(info.type)},
           m_selectedInEditor{false}
     {
-        IF_THROW(!HasNoZeroElement(scale), "Collider::Collider - Invalid scale(elements cannot be 0)");
-        IF_THROW(type == ColliderType::Sphere && scale - scale != Vector3::Zero(), "Collider::Collider - Sphere colliders do not support nonuniform scaling");
+        IF_THROW(!HasNoZeroElement(info.scale), "Collider::Collider - Invalid scale(elements cannot be 0)");
```
nightcrawlerstudios, line 32: Since this call is inverted so many times, it would be more efficient to create a HasAnyZeroElement that can jump out on the first FloatEqual(0.0f)

nc/source/physics/CollisionSystem.cpp
```diff
@@ -2,130 +2,185 @@
 #include "debug/Profiler.h"
 #include "directx/math/DirectXCollision.h"
 
+namespace
+{
+    DirectX::BoundingSphere EstimateBoundingVolume(const nc::physics::ColliderSoA::CenterExtentPair& volumeData, const DirectX::XMMATRIX* transform)
```
jaremieromer, line 7: Did we ever come up with a plan to extend this for shapes like the castle wall example we had discussed? (Where a sphere is not appropriate)

jaremieromer (approved):

PR #90, opened 2021-03-01
Index buffer. Prior to Buffer abstraction.

jaremieromer:
Added index buffer support. 
After uniform buffers (next CR) I will be abstracting the concept of buffers to avoid DRY violations present here.

### DIFF
--- a/nc/source/graphics/vulkan/Device.cpp
+++ b/nc/source/graphics/vulkan/Device.cpp
@@ -540,6 +540,14 @@ namespace nc::graphics::vulkan
         m_allocator.unmapMemory(m_buffers.at(bufferId).second);
     }
 
+    void Device::MapMemory(uint32_t bufferId, std::vector<uint32_t> indices, size_t size)
+    {
+        void* mappedData;
+        m_allocator.mapMemory(m_buffers.at(bufferId).second, &mappedData);
+        memcpy(mappedData, indices.data(), size);
+        m_allocator.unmapMemory(m_buffers.at(bufferId).second);
+    }
+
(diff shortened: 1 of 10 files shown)


nc/source/graphics/vulkan/Device.cpp
```diff
@@ -540,6 +540,14 @@ namespace nc::graphics::vulkan
         m_allocator.unmapMemory(m_buffers.at(bufferId).second);
     }
 
+    void Device::MapMemory(uint32_t bufferId, std::vector<uint32_t> indices, size_t size)
+    {
+        void* mappedData;
+        m_allocator.mapMemory(m_buffers.at(bufferId).second, &mappedData);
+        memcpy(mappedData, indices.data(), size);
+        m_allocator.unmapMemory(m_buffers.at(bufferId).second);
```
McCallisterRomer: In both MapMemory functions you're searching the map twice.  Just grab the value once.

nc/source/graphics/vulkan/IndexBuffer.cpp
```diff
@@ -0,0 +1,61 @@
...
+{
+    IndexBuffer::IndexBuffer(Device& device, Commands& commands, std::vector<uint32_t> indices)
+    : m_device { device },
+      m_commands { commands },
+      m_id { 0 },
+      m_size { 0 },
+      m_indexBuffer { nullptr },
+      m_stagingBufferId { 0 },
+      m_stagingBuffer { nullptr },
+      m_indices { indices }
```
McCallisterRomer, line 15: std::move

McCallisterRomer (approved):

PR #91, opened 2021-03-04
AssetBuilder tool

McCallisterRomer:
-Moved AssImp into a separate tool so I don't get bedsores waiting for the debugger.
-Meshes resources are now to be loaded prior to creating meshes (at launch/scene load time) to reduce random file i/o.
-Changed GfxResManager according to mesh changes. Load just loads necessary data, Acquire gets already loaded data, and AcquireOnDemand loads (if needed) and acquires the data (old behavior). 
-Removed some unnecessary members and parameters from GfxResources and GetUID functions.
-Fixed CMakeLists so that vulkan/glfw stuff is totally ignored if USE_VULKAN == OFF.


tools/asset_builder/AssetBuilder.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+    return 0;
+}
+
+void Usage()
+{
+    std::cout << "Usage: build.exe [options]\n"
+              << "Options:\n"
+              << "  -h or --help            Display this information\n"
+              << "  -f <file>               Read targets from <file>\n"
```
jaremieromer: I can definitely see myself wanting to specify a single mesh rather than having to write the mesh path to a file first. 
ie. build.exe -m "C:\Temp\MyMesh.fbx" -o "C:\Source\nc\project\assets\mesh"

jaremieromer (approved):

PR #103, opened 2021-03-16
JobSystem

McCallisterRomer:
-Added a job system and thread pool
-Updated CollisionSystem to use JobSystem for a few things


nc/source/engine/Engine.cpp
```diff
@@ -69,12 +69,13 @@ namespace nc::core
     Engine::Engine(HINSTANCE hInstance)
         : m_isRunning{ false },
           m_frameDeltaTimeFactor{ 1.0f },
+          m_jobSystem{2},
```
jaremieromer, line 72: Can this be read from a config? If not, we should extract to a constant

jaremieromer (approved):

PR #107, opened 2021-03-27
Generic SoA

McCallisterRomer:
Removed ColliderSoA and added a generic type to avoid future code repetition.


nc/source/physics/CollisionSystem.cpp
```diff
@@ -58,7 +58,15 @@ namespace nc::physics
 
     void CollisionSystem::FetchEstimates()
     {
-        m_colliderSystem.GetDynamicSoA()->CalculateEstimates(&m_dynamicEstimates);
+        auto* soa = m_colliderSystem.GetDynamicSoA();
+        auto properties = soa->View<VolumeProperties>();
```
jaremieromer: Make this and the one below const auto

nc/source/ecs/SoA.h
```diff
@@ -0,0 +1,137 @@
...
+
+    template<class... Properties>
+    bool SoA<Properties...>::IsFull() const
+    {
+        return m_nextFree >= m_size && m_gaps.empty();
+    }
+
+    template<class... Properties>
+    template<class Property>
+    std::span<Property> SoA<Properties...>::View() const
```
jaremieromer: View functions should be readonly

jaremieromer (approved):

PR #111, opened 2021-03-29
Cal/inspect components

McCallisterRomer:
Added functionality to the editor for inspecting component systems.


nc/source/ecs/ComponentSystem.h
```diff
@@ -15,9 +15,10 @@ namespace nc::ecs
     class ComponentSystem
     {
         using Allocator = alloc::PoolAllocator<T>;
-        using ContainerType = std::vector<std::unique_ptr<T, alloc::basic_deleter<Allocator>>>;
 
         public:
+            using ContainerType = std::vector<std::unique_ptr<T, alloc::basic_deleter<Allocator>>>;
```
jaremieromer, line 20: What does moving this into public do?

nc/source/engine/Engine.cpp
```diff
@@ -33,6 +33,20 @@ namespace
             .worldspaceExtent = config.physics.worldspaceExtent
         };
     }
+
+    #ifdef NC_EDITOR_ENABLED
```
jaremieromer, line 40: Can you explain why this is only here for the editor? I'm missing something here

jaremieromer (approved):

PR #113, opened 2021-04-04
Collider management moved from physics to ecs

McCallisterRomer:
-ColliderSystem is now a part of EntityComponentSystem instead of PhysicsSystem. This makes things like clearing all components a little clearer.
-Config split into multiple smaller structs
-Made logging, quit and shutdown operations noexcept


nc/include/config/Config.h
```diff
@@ -4,45 +4,45 @@
 
 namespace nc::config
 {
-    struct Config
+    struct ProjectSettings
```
jaremieromer, line 7: This looks nice!

jaremieromer (approved):

PR #118, opened 2021-05-06
Nested Transforms

McCallisterRomer:
-Transforms are now hierarchical
-Transform::GetXXX methods split into GetXXX and GetLocalXXX as they now have both local and world matrices. Translate was also split and the Space enum removed. This is more consistent with the Get methods and saves a branch. Set methods only operate on the local matrix and were not split. This can change in the future, if needed.
-Added a new sample scene demonstrating/testing nested transforms.

Minor Changes:
-Added function to set the color frames are cleared to
-View matrix calculation moved from transform to camera
-Added Quaternion::Multiply (needed for unit testing rotation propagation through hierarchy)


nc/include/component/Transform.h
```diff
@@ -5,51 +5,64 @@
...
-    {
-        World,
-        Local
-    };
-
     class Transform final : public ComponentBase
     {
         public:
-            Transform(EntityHandle handle, const Vector3& pos, const Quaternion& rot, const Vector3& scale);
+            Transform(EntityHandle handle, const Vector3& pos, const Quaternion& rot, const Vector3& scale, Transform* parent);
```
jaremieromer, line 16: I've always preferred a second constructor rather than a single constructor with an optional parameter.

nc/include/component/Transform.h
```diff
@@ -5,51 +5,64 @@
...
             void Set(const Vector3& pos, const Quaternion& quat, const Vector3& scale);
             void Set(const Vector3& pos, const Vector3& angles, const Vector3& scale);
             void SetPosition(const Vector3& pos);
             void SetRotation(const Quaternion& quat);
             void SetRotation(const Vector3& angles);
             void SetScale(const Vector3& scale);
-
-            void Translate(Vector3 vec, Space space = Space::Local);
+            void Translate(Vector3 vec);
+            void TranslateLocalSpace(Vector3 vec);
```
jaremieromer, line 43: What's going on with this method and the above comment? Does Translate(Vector3 vec) operate in world space then? If so, we should name all other local space setters XLocalSpace or XLocal

nc/include/math/Quaternion.h
```diff
@@ -19,6 +19,9 @@ namespace nc
         static Quaternion FromAxisAngle(const Vector3& axis, float angle);
     };
 
+    // note: computes lhs rotation followed by rhs (the product rhs*lhs)
```
jaremieromer: Can you add a note as to why it is in this order

nc/source/component/Transform.cpp
```diff
@@ -28,24 +28,48 @@ namespace
...
-    Transform::Transform(EntityHandle handle, const Vector3& pos, const Quaternion& rot, const Vector3& scale)
+    Transform::Transform(EntityHandle handle, const Vector3& pos, const Quaternion& rot, const Vector3& scale, Transform* parent)
         : ComponentBase(handle),
-          m_matrix{ComposeMatrix(scale, rot, pos)}
+          m_localMatrix{ComposeMatrix(scale, rot, pos)},
+          m_worldMatrix{m_localMatrix},
+          m_parent{parent},
+          m_children{}
     {
         IF_THROW(HasAnyZeroElement(scale), "Transform::Transform - Invalid scale(elements cannot be 0)");
```
jaremieromer, line 38: Might be worthwhile to not call ComposeMatrix in the initializer list, in the case where the IF_THROW would throw, we would not have to perform that matrix multiplication

jaremieromer: Call it in this scope after the IF_THROW instead

nc/source/component/Transform.cpp
```diff
@@ -128,82 +150,167 @@ namespace nc
     void Transform::Set(const Vector3& pos, const Quaternion& quat, const Vector3& scale)
     {
         IF_THROW(HasAnyZeroElement(scale), "Transform::Set - Invalid scale(elements cannot be 0)");
-        m_matrix = ToScaleMatrix(scale) * ToRotMatrix(quat) * ToTransMatrix(pos);
+        m_localMatrix = ToScaleMatrix(scale) * ToRotMatrix(quat) * ToTransMatrix(pos);
```
jaremieromer, line 153: In the essence of DRY - create a private function that takes in a local matrix and sets it, then propagates it via calling UpdateWorldMatrix(); and UpdateChildren().

eg:

SetAndApplyTransform(XMMatrix localMatrix)
{
    m_localMatrix = std::move(localMatrix);
    UpdateWorldMatrix();
    UpdateChildren();
}

Then, in each of these setters you can just call SetAndApplyTransform(<some operation that generates the local transform>)

nc/source/engine/Engine.cpp
```diff
@@ -255,6 +255,7 @@ namespace nc::core
             // No UI for us until we tackle integrating IMGUI with Vulkan
         #else
             m_window.BindGraphicsOnResizeCallback(std::bind(graphics::Graphics::OnResize, &m_graphics, _1, _2, _3, _4, _5));
+            m_window.BindGraphicsSetClearColorCallback(std::bind(graphics::Graphics::SetClearColor, &m_graphics, _1));
```
jaremieromer, line 258: do we need this much ado around the clear color? Any reason it can't be set as a const once in some file and forgotten about?

nc/source/math/Quaternion.cpp
```diff
@@ -55,4 +55,14 @@ namespace nc
         DirectX::XMStoreQuaternion(&out, quat_v);
         return out;
     }
+
+    Quaternion Multiply(const Quaternion& lhs, const Quaternion& rhs)
+    {
+        auto lhs_v = DirectX::XMVectorSet(lhs.x, lhs.y, lhs.z, lhs.w);
```
jaremieromer, line 61: Will you add the comment about order importance here as well?

jaremieromer (approved):

PR #125, opened 2021-05-13
Enabled depth/stencil testing.

jaremieromer:
Enabled depth/stencil testing and proper culling.


McCallisterRomer:
How nice and short!

PR #126, opened 2021-05-14
Cal/random

McCallisterRomer:
Moved rng so that particles and the sample scene don't have to duplicate code.


nc/source/random/Random.cpp
```diff
@@ -0,0 +1,52 @@
...
+    {
+        public:
+            FloatDistribution()
+                : m_device{},
+                  m_gen{m_device()},
+                  m_distribution{-1.0f, std::nextafter(1.0f, std::numeric_limits<float>::max())}
+            {
+            }
+
+        float Generate()
```
jaremieromer, line 17: With the same input does this generate the same output? (is there no behind the scenes seed?) If it does, we should introduce a seed param.

jaremieromer (approved):

PR #128, opened 2021-05-15
PoolAdapter

McCallisterRomer:
Pulled most of the functionality out of ComponentSystem into PoolAdapter, which manages a contiguous range of pointers mapping to active pool allocations. The primary purpose for this is to allow ComponentSystem-like management of memory pools with types that do not derive from Component. ComponentSystem now uses a PoolAdapter instead of directly dealing with the allocator.


jaremieromer:
Awesome

jaremieromer (approved):

PR #130, opened 2021-05-15
EntitySystem

McCallisterRomer:
-Added EntitySystem to handle most entity-related tasks previously handled by EntityComponentSystem.
-Added EntityExists to Ecs.h
-DestroyEntity in Ecs.h now throws on failure rather than checking if the entity is valid.

### DIFF
--- a/nc/source/ecs/EntityComponentSystem.h
+++ b/nc/source/ecs/EntityComponentSystem.h
@@ -53,7 +53,7 @@ namespace nc::ecs
             std::span<Entity*> GetActiveEntities() noexcept;
 
             EntityHandle CreateEntity(EntityInfo info);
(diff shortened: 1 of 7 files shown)


nc/source/ecs/EntityComponentSystem.h
```diff
@@ -38,40 +30,33 @@ namespace nc::ecs
     {
         public:
             #ifdef USE_VULKAN
-            EntityComponentSystem();
+            EntityComponentSystem(const config::MemorySettings& memSettings,
+                                  const config::PhysicsSettings& physSettings);
```
jaremieromer, line 34: Very nice re-org of this class

jaremieromer (approved):

PR #131, opened 2021-05-22
Vulkan lighting

jaremieromer:
Modified the rendering system by adding a global renderer (Renderer.cpp) that is responsible for binding the different pipelines and the data required to render them, then recording the commands.

Added a point lights component system (PointLightSystem.cpp) and added the Vulkan data in GraphicsResources.h (GraphicsResources/PointLightsData).

Updated the Vulkan scene to show the new point lights.


Cube.nca
```diff
@@ -0,0 +1,38 @@
+24
```
McCallisterRomer: Delete this file

nc/include/component/vulkan/PointLight.h
```diff
@@ -0,0 +1,52 @@
...
+        float attLin = 0.22f;
+        float attQuad = 0.20f;
+        bool isInitialized = 0;
+        float padding[3] = {0.0f, 0.0f, 0.0f};
+    };
+
+    class PointLight final : public Component
+    {
+        public:
+            PointLight(EntityHandle parentHandle);
```
McCallisterRomer: You shouldn't need two constructors for a component. There should only be one place where they are created.

nc/source/component/vulkan/PointLight.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+namespace nc::vulkan
+{
+    Attenuation GetAttenuationFromRange(float range)
+    {
+        Attenuation att = {};
+        att.constant = 1.0f;
+        att.linear = 4.5f / range;
+        att.quadratic = 75.0f / (range * range);
+        return att;
```
McCallisterRomer, line 17: You can use aggregate initialization here. You can also use designators (.member = xxx) if you still want to see the member names.

Also prefer to use RVO when possible as copy elision is mandatory, although what you have here would likely also be elided (it is a simple case for NRVO).

return Attenuation
{
    .constant = 1.0f,
    ...,
    ...
};

nc/include/component/vulkan/PointLight.h
McCallisterRomer: On further inspection, I don't think there is a definition for this c'tor.

nc/source/component/vulkan/PointLight.cpp
```diff
@@ -0,0 +1,86 @@
...
+                ui::editor::floatWidget("Intensity", "difintensity", &m_info.diffuseIntensity, dragSpeed,  0.0f, 600.0f, "%.2f");
+                ui::editor::floatWidget("Range", "range", &m_range, dragSpeed,  0.0f, 600.0f, "%.2f");
+            ImGui::Unindent();
+        ImGui::Unindent();
+    }
+    #endif
+
+    PointLight::PointLight(EntityHandle parentHandle, PointLightInfo info)
+    : Component{parentHandle},
+      m_range{13},
```
McCallisterRomer, line 49: Is this meant to be hardcoded?

nc/source/component/vulkan/PointLight.cpp
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    void PointLight::Update()
+    {
+        auto transformPos = m_transform->GetPosition();
+        auto attenuation = GetAttenuationFromRange(m_range);
+        m_info.attConst = attenuation.constant;
+        m_info.attLin = attenuation.linear;
+        m_info.attQuad = attenuation.quadratic;
+        m_info.pos = Vector4(transformPos.x, transformPos.y, transformPos.z, 1.0);
```
McCallisterRomer, line 71: I would use 1.0f for your homogeneous coordinate as without it, this is a narrowing conversion. It doesn't logically affect anything, but Wconversion or whatever it is will probably complain about this if we ever turn it back on.

nc/include/component/vulkan/PointLight.h
```diff
@@ -0,0 +1,52 @@
...
+    {
+        Vector4 pos = Vector4::Zero();
+        Vector4 ambient = Vector4{0.2f, 0.2f, 0.2f, 1.0f};
+        Vector4 diffuseColor = Vector4{0.75f, 0.75f, 0.75f, 1.0f};
+        Vector4 specularColor = Vector4{0.75f, 0.75f, 0.75f, 1.0f};
+        float diffuseIntensity = 10.0f;
+        float attConst = 1.0f;
+        float attLin = 0.22f;
+        float attQuad = 0.20f;
+        bool isInitialized = 0;
```
McCallisterRomer, line 30: I haven't pieced together why exactly this is here, but I don't love the bool. Its costing 16 bytes with the padding and doesn't look like it is really needed. Maybe I haven't gone far enough to reveal its true purpose, but it feels icky vicky.

nc/source/component/vulkan/PointLight.cpp
```diff
@@ -0,0 +1,86 @@
...
+        m_info.isInitialized = true;
+    }
+
+    void PointLight::Set(const PointLightInfo& info)
+    {
+        m_info.pos = info.pos;
```
McCallisterRomer, line 77: Can you just do m_info = info? Also diffuseIntensity is missing.

nc/source/graphics/vulkan/Resources/WriteableBuffer.h
```diff
@@ -0,0 +1,103 @@
...
+    void WriteableBuffer<T>::Map(const std::vector<T>& dataToMap)
+    {
+        auto allocation = *(m_base->GetBufferAllocation(m_memoryIndex));
+        
+        void* dataContainer;
+        m_base->GetAllocator()->mapMemory(allocation, &dataContainer);
+
+        T* mappedContainerHandle = (T*)dataContainer;
+
+        for (uint32_t i = 0; i < static_cast<uint32_t>(dataToMap.size()); ++i)
```
McCallisterRomer, line 96: Can you memcpy here?

nc/source/graphics/vulkan/Base.cpp
```diff
@@ -307,6 +309,16 @@ namespace nc::graphics::vulkan
...
     }
 
+    vma::Allocator* Base::GetAllocator() noexcept
+    {
+        return &m_allocator;
+    }
+
+    vma::Allocation* Base::GetBufferAllocation(uint32_t index) noexcept
+    {
+        return &(m_buffers.at(index)).second;
```
McCallisterRomer, line 319: unordered_map::at() isn't noexcept

project/source/shared/WasdController.h
```diff
@@ -26,6 +26,6 @@ namespace nc::sample
     inline void WasdController::FrameUpdate(float dt)
     {
         auto [leftRight, frontBack] = input::GetAxis() * m_speed * dt;
-        m_transform->TranslateLocalSpace(Vector3{leftRight, 0.0f, frontBack});
+        m_transform->Translate(Vector3{leftRight, 0.0f, frontBack});
```
McCallisterRomer, line 29: I think we want this back in local space.

tools/asset_builder/targets.txt
```diff
@@ -4,4 +4,6 @@ project/models/beeper.fbx
 project/models/coin.fbx
 project/models/table.fbx
 project/models/token.fbx
-project/models/worm.fbx
\ No newline at end of file
+project/models/worm.fbx
+project/models/Cube.fbx
```
McCallisterRomer, line 8: Is this a different "cube" than the one in nc/resources? I forgot to move the plane out of the project dir, but I'd like to keep the primitives in the nc/resources directory because the engine needs cube/sphere/plane regardless of the project.

McCallisterRomer (approved):
Nice job - lighting makes a huge difference in how nice the scene looks!

PR #132, opened 2021-05-27
Cal/sparse set

McCallisterRomer:
-Removed ComponentSystem
-New type Registry owns storage for each component type + the EntitySystem
-Changed EntityHandle to uint64_t made from concatenating multiple smaller integers
-All places where component pointers were caches were changed to lookups using EntityHandle
-Added a to add staging area for EntitySystem
-HandleManager can recycle old handles
-Fixed bug in CollisionBenchmark where adding/removing in the middle of rendering could crash 
-Transform unit tests are intentionally commented out - they are broken with new changes, and Transform is likely changing soon


nc/source/ecs/ColliderTree.h
```diff
@@ -15,6 +15,7 @@ namespace nc
     {
         using allow_trivial_destruction = std::true_type;
         using sort_dense_storage_by_address = std::true_type;
+        // ignore callbacks
```
jaremieromer, line 18: Is this a todo?

nc/source/ecs/Registry.h
```diff
@@ -0,0 +1,347 @@
...
+        auto requiredSize = storage.componentPool.size() + additionalRequiredCount;
+        storage.componentPool.reserve(requiredSize);
+        storage.entityPool.reserve(requiredSize);
+    }
+
+    template<class... Ts>
+    template<class T>
+    void Registry<Ts...>::RegisterOnAddCallback(typename SystemCallbacks<T>::on_add_type func)
+    {
+        static_assert(StoragePolicy<T>::requires_on_add_callback::value, "Cannot register an OnAdd callback unless specified in the StoragePolicy");
```
jaremieromer, line 299: It's neat that these are compile-time validated.

project/source/spawn_test/SpawnTest.cpp
```diff
@@ -68,7 +68,7 @@ namespace nc::sample
         staticCubeSpawner->SetPrefab(prefab::Resource::CubeRed);
         staticCubeSpawner->Spawn(5);
 
-        // Fixed interval spawner for moving cubes
+        //Fixed interval spawner for moving cubes
```
jaremieromer, line 71: Why is this changed? I prefer one space, and the others in this file are one space.

jaremieromer (approved):

project/source/spawn_test/SpawnTest.cpp
McCallisterRomer: no idea

nc/source/ecs/ColliderTree.h
McCallisterRomer: No, just commenting that I've intentionally left out aliases that won't be used.

PR #134, opened 2021-06-08
Registry Passed to Scene

McCallisterRomer:
-Global Ecs stuff mostly removed in favor of passing a registry pointer around.
-Entity removed
-Entity handle renamed to Entity
-Tag added as a component
-AutoComponentGroup added to replace the vector of component* previously in Entity


nc/source/component/AutoComponentGroup.cpp
```diff
@@ -0,0 +1,54 @@
...
+            comp->FixedUpdate();
+    }
+
+    void AutoComponentGroup::SendOnDestroy()
+    {
+        for(auto& comp : m_components)
+            comp->OnDestroy();
+    }
+
+    void AutoComponentGroup::SendOnCollisionEnter(Entity hit)
```
nightcrawlerstudios, line 37: I don't like the collision specific methods here. Will these be moved in future PRs?

jaremieromer (approved):

PR #135, opened 2021-06-08
Component Sorting

McCallisterRomer:
-Registry now gets its types from RegistryTypeList
-Some Registry functionality moved into PerComponentStorage
-Added ViewGroup to Registry - more or less a ViewAll<T, U> that sorts a range within pools T & U by entity and returns spans over that range.

### DIFF
--- a/nc/include/component/AutoComponentGroup.h
+++ b/nc/include/component/AutoComponentGroup.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "component/Component.h"
+#include "debug/Utils.h"
 
 #include <concepts>
 #include <memory>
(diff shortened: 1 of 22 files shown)


nc/include/component/AutoComponentGroup.h
```diff
@@ -45,17 +46,15 @@ namespace nc
     };
 
     template<std::derived_from<AutoComponent> T, class ... Args>
-    T* AutoComponentGroup::Add(Args&& ... args) noexcept
+    T* AutoComponentGroup::Add(Args&& ... args)
```
jaremieromer, line 49: Always nice to throw instead of return null IMO, nice

nc/include/component/Registry.h
```diff
@@ -211,306 +161,253 @@ namespace nc::ecs
...
-        auto& [emplacedEntity, emplacedComponent] = storage.stagingPool.emplace_back(entity, T{entity, std::forward<Args>(args)...});
-
-        if constexpr(StoragePolicy<T>::requires_on_add_callback::value)
-        {
-            storage.callbacks.OnAdd(emplacedComponent);
-        }
-
-        return &emplacedComponent;
+        IF_THROW(!Contains<Entity>(entity), std::string{"Bad Entity\n   "} + __PRETTY_FUNCTION__);
+        return GetStorageFor<T>().Add(entity, std::forward<Args>(args)...);
```
jaremieromer, line 198: I'm a big fan of this - it's much easier to see what's going on

jaremieromer (approved):

PR #139, opened 2021-06-25
Intermediary Particle System 

jaremieromer:
Converted particle system from DX to Vulkan.
Increased roughness on planets.
Added a build mode for Debug w/out Vulkan validation to speed up debug build. (For when Vulkan is stable, and we are working in debug mode on other aspects of the engine.)


nc/source/graphics/vulkan/Base.cpp
```diff
@@ -13,7 +13,7 @@ namespace
     const std::vector<const char*> GlobalExtensions = { VK_KHR_WIN32_SURFACE_EXTENSION_NAME, VK_KHR_SURFACE_EXTENSION_NAME };
     const std::vector<const char*> ValidationLayers = { "VK_LAYER_KHRONOS_validation" };
 
-    #ifdef NC_DEBUG_BUILD
+    #ifdef VALIDATION
```
McCallisterRomer, line 16: Can we rename this to something like NC_VULKAN_VALIDATION?

nc/source/graphics/vulkan/techniques/ParticleTechnique.h
```diff
@@ -0,0 +1,56 @@
+#pragma once
+
+#include "component/Component.h"
```
McCallisterRomer, line 3: why Component.h?

McCallisterRomer (approved):

PR #142, opened 2021-06-28
Converted prefabs and worms scene to vulkan

jaremieromer:
Converted Prefabs.h/.cpp to Vulkan.
Added a second point light to the worms scene as the lighting model is a little different for Vulkan.
Note: the planets are currently manually added to the solar_system scene, so they are not in the Prefabs.


project/source/shared/Prefabs.cpp
```diff
@@ -219,226 +164,157 @@ Entity Create_(registry_type*, EntityInfo);
 template<> Entity Create_<Resource::Beeper>(registry_type* registry, EntityInfo info)
 {
     auto handle = registry->Add<Entity>(std::move(info));
-    registry->Add<Renderer>(handle, mesh::Beeper, material::Beeper);
+    registry->Add<vulkan::MeshRenderer>(handle, mesh::meshPaths[0], material::Beeper, nc::graphics::vulkan::TechniqueType::PhongAndUi);
```
McCallisterRomer, line 169: Indexing the vector like this seems brittle and confusing. I don't think we have to fix it right at this moment, but I'd like to find a better way.

McCallisterRomer (approved):

PR #145, opened 2021-07-18
Split CollisionSystem into Free Functions

McCallisterRomer:
-Removed CollisionSystem. Its responsibilities are handled by free functions in a few new files:
- CollisionCache: simple struct for maintaining collision state.
- CollisionPhases: setup, broad and narrow phase detection.
- CollisionNotification: notifies collision/trigger events.

-Moved some free functions in PhysicsSystem.cpp to their own file Dynamics.h/cpp
-Added default constructor for Entity. Most of the casting to underlying_type was because containers of objects with no default c'tor are difficult. This gets rid of a lot of noise.
-Trigger colliders work. Added OnTriggerXXX methods to AutoComponent.
-Fixed small bug in Gjk.
-Fixed Polytope Initialize so it doesn't unnecessarily reallocate vectors.
-Solver.cpp shouldn't need to be looked at too closely, even though there were many changes. I just moved the internal functions into an anonymous namespace.


nc/source/engine/Engine.cpp
```diff
@@ -180,14 +180,12 @@ namespace nc::core
 
     void Engine::FixedStepLogic(float dt)
     {
-        NC_PROFILE_BEGIN(debug::profiler::Filter::Physics);
```
jaremieromer, line 183: Why removed?

nc/source/physics/collision/CollisionCache.h
```diff
@@ -0,0 +1,83 @@
...
+
+#include "Manifold.h"
+#include "ecs/ColliderTree.h"
+
+#include <cstdint>
+#include <vector>
+
+namespace nc::physics
+{
+    /** Handling collisions requires either a physics response and collision notification
```
jaremieromer, line 11: Great comments

jaremieromer (approved):

PR #147, opened 2021-07-28
DX11 -> Vulkan cutover for click and collision events scenes

jaremieromer:
Converted the ClickEvents and CollisionEvents scenes from DirectX11 to Vulkan.
Added a DebugWidget component that allows for rendering of a wireframe mesh.
The Collider component will (if the editor is enabled) now add/remove a DebugWidget component to the parent entity.
The DebugWidget can be enabled/disabled.


nc/include/ECS.h
```diff
@@ -10,12 +10,13 @@
...
 #include "component/Tag.h"
 #include "component/Transform.h"
+#include "component/vulkan/DebugWidget.h"
 #include "component/vulkan/MeshRenderer.h"
 #include "component/vulkan/PointLight.h"
 
 namespace nc
 {
-    using registry_type_list = ecs::RegistryTypeList<AutoComponentGroup, Collider, NetworkDispatcher, ParticleEmitter, PointLight, Renderer, Tag, Transform, vulkan::MeshRenderer, vulkan::PointLight>;
+    using registry_type_list = ecs::RegistryTypeList<AutoComponentGroup, Collider, NetworkDispatcher, ParticleEmitter, PointLight, Renderer, Tag, Transform,  vulkan::DebugWidget, vulkan::MeshRenderer, vulkan::PointLight>;
```
McCallisterRomer, line 19: What if we're in release mode? We shouldn't need to allocate storage for debug only types.

nc/include/component/vulkan/DebugWidget.h
```diff
@@ -0,0 +1,57 @@
+#pragma once
+
+#include "component/Component.h"
+#include "directx/math/DirectXMath.h"
+
+#include <string>
+
+namespace nc { class Transform; }
```
McCallisterRomer, line 8: I may be missing it, but I don't think the forward decl is needed.

nc/include/component/vulkan/DebugWidget.h
```diff
@@ -0,0 +1,57 @@
...
+
+const auto CubeMeshPath = std::string{"project/assets/mesh/cube.nca"};
```
McCallisterRomer, line 10: Move these into a namespace

McCallisterRomer: Actually, can these go in the cpp in an anonymous namespace?

nc/source/component/vulkan/DebugWidget.cpp
```diff
@@ -0,0 +1,79 @@
...
+#ifdef NC_EDITOR_ENABLED
+#include "ui/editor/Widgets.h"
+#endif
+
+namespace nc::vulkan
+{
+    DebugWidget::DebugWidget(Entity entity, WidgetShape shape)
+        : ComponentBase{entity}
+    {
+        switch(shape)
```
McCallisterRomer, line 13: I prefer having the switch in a function. I feel like it makes the constructor clearer. Then you could just do m_meshUid{GetMeshUid(shape)}

nc/source/ecs/ColliderSystem.cpp
```diff
@@ -70,4 +79,39 @@ namespace nc::ecs
         m_dynamicSoA.Clear();
         m_staticTree.Clear();
     }
+
+    #ifdef NC_EDITOR_ENABLED
+    void ColliderSystem::AddDebugWidget(const Collider& collider)
```
McCallisterRomer, line 84: Yikes, I deleted this class in the PR you just reviewed.

nc/source/graphics/vulkan/MeshManager.cpp
```diff
@@ -111,8 +111,20 @@ namespace nc::graphics::vulkan
         std::vector<Vertex> allVertices = {};
         std::vector<uint32_t> allIndices = {};
         std::unordered_map<std::string, Mesh> meshes;
+        std::vector<std::string> pathsToLoad;
 
-        for (auto& path : meshPaths)
+        auto meshesExist = ResourceManager::HasMeshes();
```
McCallisterRomer, line 116: There are a few things to look at here.

The first argument to insert on line 120 was probably meant to be pathsToLoad.end()

GetMeshPaths copies the vector 3 times (at least one of them should get optimized away). ResourceManager::GetMeshPaths, ResourceManager::GetMeshPaths_, and MeshesData::GetMeshPaths should probably be returning by const& to avoid this.

Also note that you are going to copy meshPaths to pathsToLoad at some point. Either through insert on line 120 or copy initialization on 124. Because you are taking the parameter by const&, there is no other choice. If you took by value instead, you can give the caller the choice to copy or move then you can directly modify the parameter (insert, move somewhere else, whatever). This also allows you to get rid of the else statement and pathsToLoad variable entirely. If ResourceManager has meshes, you can insert those meshes directly into the meshPaths.

McCallisterRomer: Well, maybe ignore that second point. I thought MeshesData had a vector not a map.

nc/source/graphics/vulkan/techniques/WireframeTechnique.cpp
```diff
@@ -121,6 +134,28 @@ namespace nc::graphics::vulkan
...
         pushConstants.viewProjection = viewMatrix * projectionMatrix;
+
+        #ifdef NC_EDITOR_ENABLED
+        for (auto& [meshUid, debugWidgets] : m_debugWidgets)
+        {
+            const auto meshAccessor = ResourceManager::GetMeshAccessor(meshUid);
+            
+            for (auto handle : debugWidgets)
+            {
+                auto* widget = ActiveRegistry()->Get<nc::vulkan::DebugWidget>(handle);
```
McCallisterRomer, line 145: Hoist ActiveRegistry() out of both for loops.

McCallisterRomer (approved):

PR #149, opened 2021-08-19
HullCollider Inertia Tensor

McCallisterRomer:
-HullColliders now have better inertia tensors. The asset builder finds the x/y/z extents of the input vertices and writes them to the .nca file. Those values are then used by PhysicsBody to estimate a tensor based a bounding box.
-Found typo in Capsule inertia code causing the classic "jumping bean" bug.
-Fixed SphereCollider inertia tensor being based on a hollow volume instead of a solid one. Oops.
-Consolidated some comments scattered throughout various source files into the docs.
-Fixed an issue in Polytope with failed barycentric projection. In the past, that code was returning NaN occasionally, so that check will remain until I'm confident that there aren't lingering issues.
-Changed [] style indexing to at() in Polytope. There is one more issue to fix there, and I thought it may be an out of range thing. It isn't.
-Added caching of the broad collision pair count giving us a better guess for vector reserving.


nc/source/physics/dynamics/Solver.cpp
```diff
@@ -316,19 +316,19 @@ namespace
...
+            pA = XMLoadVector3(&position) + joint.rA;
 
-            auto* transformB = registry->Get<Transform>(joint.entityB);
-            auto positionB = transformB->GetPosition();
-            auto rotationB = transformB->GetRotation();
-            auto rotMatrix = XMMatrixRotationQuaternion(XMLoadQuaternion(&rotationB));
+            transform = registry->Get<Transform>(joint.entityB);
+            position = transform->GetPosition();
+            rotation = transform->GetRotation();
+            rotMatrix = XMMatrixRotationQuaternion(XMLoadQuaternion(&rotation));
```
jaremieromer, line 329: ha - this double declare of rotMatrix popped up on my merge this morning. I see its fixed now.

jaremieromer (approved):

PR #152, opened 2021-08-30
Concave Collisions

McCallisterRomer:
-Added component and asset for MeshCollider.
-Added BspTree to manage MeshColliders and make them a little cheaper to test against.
-Changed CollisionVolumes.h to Geometry.h and removed 'collider' from some type names. There are some cases where we want to use these but aren't explicitly dealing with colliders. Added plane and triangle types.
-Replaced HullColliderManager with AssetManager. It deals with both hull and mesh collider assets.
-Changed how Gjk/Epa are called to allow adding optimized special cases. So far, only sphere vs sphere is done.


nc/include/component/MeshCollider.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include "Component.h"
+
+namespace nc
+{
+    class MeshCollider final : public ComponentBase
```
jaremieromer: I've changed my mind. I prefer ConcaveCollider over MeshCollider. Years down the road, when we've forgotten about the fact that this should only be a concave collider and have performance issues, we're going to wish we'd named it such.

jaremieromer (approved):

PR #153, opened 2021-09-03
Vulkan more robust gpu handling logic

jaremieromer:


nc/source/graphics/vulkan/Renderer.h
```diff
@@ -42,6 +39,7 @@ namespace nc::graphics::vulkan
             void DeregisterRenderable(Entity entity);
 
         private:
+            std::pair<std::span<nc::vulkan::MeshRenderer>, std::span<nc::vulkan::MeshRenderer>> GetMeshRenderersByTechnique();
```
McCallisterRomer: I think you deleted this.

nc/source/graphics/vulkan/Resources/GraphicsResources.h
```diff
@@ -90,4 +91,23 @@ namespace nc::graphics::vulkan
...
             vk::UniqueDescriptorSetLayout m_descriptorSetLayout;
     };
+
+    class ObjectsData
+    {
+        public:
+            ObjectsData() = default;
+            ObjectsData(Graphics2* graphics);
+            ObjectsData(ObjectsData&&) = default;
+            ObjectsData& operator=(ObjectsData&&) = default;
```
McCallisterRomer, line 100: Will you explicitly delete the copy operations? If they don't exist, I like to make it obvious that it is intentional.

nc/include/component/vulkan/MeshRenderer.h
```diff
@@ -1,28 +1,45 @@
...
     class MeshRenderer : public ComponentBase
     {
         public:
             MeshRenderer(Entity entity, std::string meshUid, nc::graphics::vulkan::Material material, nc::graphics::vulkan::TechniqueType techniqueType);
-            Transform* GetTransform();
-            const std::string& GetMeshUid() const;
+            const nc::graphics::vulkan::Mesh& GetMesh() const;
             nc::graphics::vulkan::Material& GetMaterial();
             nc::graphics::vulkan::TechniqueType GetTechniqueType() const;
+            ObjectData GetObjectData() const;
```
McCallisterRomer: I know that your calls to this want to make a copy, but it should return const& so the caller has the option of not making a copy.

nc/source/component/vulkan/MeshRenderer.cpp
```diff
@@ -9,30 +10,42 @@ namespace nc::vulkan
...
+        return m_techniqueType;
     }
 
     nc::graphics::vulkan::Material& MeshRenderer::GetMaterial()
     {
         return m_material;
     }
 
-    nc::graphics::vulkan::TechniqueType MeshRenderer::GetTechniqueType() const
+    void MeshRenderer::Update(const  DirectX::FXMMATRIX& viewMatrix, const  DirectX::FXMMATRIX& projectionMatrix)
```
McCallisterRomer: I eventually want to get rid of ActiveRegistry(). The more places it is used, the harder it becomes, so can you pass the registry to this method? The system that calls this already has it.

nc/source/graphics/Graphics2.h
```diff
@@ -66,6 +65,7 @@ namespace nc::graphics
             std::unique_ptr<vulkan::Swapchain> m_swapchain;
             std::unique_ptr<vulkan::Commands> m_commands;
             vulkan::Renderer* m_renderer;
+            std::unique_ptr<vulkan::ResourceManager> m_resourceManager;
```
McCallisterRomer: This doesn't have to be a unique_ptr; we just did that because ResourceManager has mostly (all?) inline functions in its header. Will you put a todo here and in ResourceManager to fix these things?

nc/source/graphics/vulkan/Base.cpp
```diff
@@ -576,7 +576,8 @@ namespace nc::graphics::vulkan
         auto buffer = m_buffers.find(id);
         if (buffer == m_buffers.end())
         {
-            throw std::runtime_error("The given buffer ID was not present in the dictionary.");
+            return;
+            //throw std::runtime_error("The given buffer ID was not present in the dictionary.");
```
McCallisterRomer: This is a legit bug, right?

nc/source/graphics/vulkan/Renderer.cpp
```diff
@@ -11,19 +11,23 @@
 #include "graphics/vulkan/Swapchain.h"
 #include "graphics/vulkan/Initializers.h"
 
+#include <span>
+#include <iostream>
```
McCallisterRomer: I think this is left over from me.

nc/source/graphics/vulkan/Resources/GraphicsResources.h
```diff
@@ -7,6 +7,7 @@
 #include "graphics/vulkan/Mesh.h"
 #include "graphics/vulkan/Texture.h"
 #include "component/vulkan/PointLight.h"
+#include "component/vulkan/MeshRenderer.h"
 
 #include <vector>
 #include <unordered_map>
```
McCallisterRomer: It doesn't look like you're using unordered_map here.

McCallisterRomer (approved):

PR #156, opened 2021-09-08
Fixed time step

McCallisterRomer:
Added fixed time step because we can't use the frame dt if frame rate isn't capped. This new method forces every physics step to operate on the same time step (specified in the config), but it doesn't necessarily get called with this frequency. During each game loop iteration, physics may run several times (or not at all) in an attempt to sync to the desired rate. This solves the 'exploding' problem caused by dt spikes by running physics multiple times instead of once with a large dt. This introduces another problem, however. In intensive scenes with a high update frequency, the frame rate drop is worsened due to physics being 'behind' and running extra times to try to catch up. I've added a maximum iteration count to try and mitigate this.


nc/source/time/NcTime.cpp
```diff
@@ -13,38 +13,21 @@ namespace nc::time
     using std::chrono::nanoseconds;
 
     Time::Time() noexcept
+        : m_lastTime{Clock::now()},
```
jaremieromer, line 16: We should eventually switch over and normalize using UTC time, especially when we get into networking.

jaremieromer (approved):

PR #157, opened 2021-09-09
AudioSystem

McCallisterRomer:
-Added AudioSystem and AudioSource
-AssetManager can now load wave files.
-Added RtAudio and AudioFile library. AudioFile has UB warnings, so I'm not sure if we'll fix those or write our own wave reader.
-Missing licenses for various libraries


nc/source/component/AudioSource.cpp
```diff
@@ -0,0 +1,103 @@
...
+        : ComponentBase{entity},
+          m_leftChannel{},
+          m_rightChannel{},
+          m_samplesPerChannel{},
+          m_currentSampleIndex{0u},
+          m_properties{properties},
+          m_playing{false}
+    {
+        const auto& soundClip = AssetManager::AcquireSoundClip(path);
+        m_leftChannel = std::span<const double>{soundClip.leftChannel};
```
jaremieromer, line 40: Be consistent here with curlies vs parentheses

jaremieromer (approved):

PR #160, opened 2021-09-10
PerFrameRenderState

McCallisterRomer:
PerFrameRenderState is a copy of the registry data needed for rendering. It can be created after all writes in a frame have finished, allowing rendering to run on another thread without the need for synchronization.


nc/source/graphics/PerFrameRenderState.h
```diff
@@ -0,0 +1,22 @@
...
+#include "Ecs.h"
+#include "resources/ObjectData.h"
+
+namespace nc::graphics
+{
+    struct PerFrameRenderState
+    {
+        PerFrameRenderState(registry_type* registry, bool isPointLightSystemDirty);
+
+        DirectX::XMMATRIX viewMatrix;
```
nightcrawlerstudios, line 12: Will you change these to cameraViewMatrix and cameraProjectionMatrix?

jaremieromer (approved):

PR #163, opened 2021-09-15
Simplify Entity

McCallisterRomer:
Entity was unnecessarily complicated due to a mistake I made at some point regarding x64 calling conventions. As long as Entity is 64 bits or less, it can be passed by register, so all the bit shifting stuff isn't needed. Also, version isn't used anymore, so it can be removed.

### DIFF
--- a/nc/include/component/PerComponentStorage.h
+++ b/nc/include/component/PerComponentStorage.h
@@ -131,7 +131,7 @@ namespace nc::ecs
     template<Component T>
     bool PerComponentStorage<T>::Contains(Entity entity) const
     {
-        if(Entity::NullIndex != sparseArray.at(entity.Index()))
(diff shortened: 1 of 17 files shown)


nc/include/component/PerComponentStorage.h
```diff
@@ -131,7 +131,7 @@ namespace nc::ecs
     template<Component T>
     bool PerComponentStorage<T>::Contains(Entity entity) const
     {
-        if(EntityTraits::NullIndex != sparseArray.at(EntityUtils::Index(entity)))
+        if(Entity::NullIndex != sparseArray.at(entity.Index()))
```
jaremieromer: nit but will you reverse order of operands here?

jaremieromer (approved):

PR #164, opened 2021-09-17
Taskflow

McCallisterRomer:
-Added Taskflow library to replace our thread pool/job system.
-Added TaskGraph wrapper around taskflow for exceptions.
-Added a task graph for the physics step & some main loop jobs.
-Changed physics functions to use out params instead of return values. While I like this less, it plays nicer with the task graph.
-Made some changes to Gjk that prevent NaN issues when passing a bad simplex to Epa.


nc/source/physics/collision/CollisionCache.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "Manifold.h"
+#include "../dynamics/Constraint.h"
```
jaremieromer: change path

nc/source/physics/collision/IntersectionQueries.cpp
```diff
@@ -462,9 +456,18 @@ namespace nc::physics
             XMStoreVector3(&localSupportB, supports.localB);
 
             stateOut->simplex.PushFront(supportCSO, worldSupportA, worldSupportB, localSupportA, localSupportB);
-
+
```
jaremieromer: nit but, remove whitespace here

nc/source/task/Task.h
```diff
@@ -0,0 +1,116 @@
...
+
+        private:
+            tf::Taskflow m_tasks;
+            TaskExceptionState m_exceptionState;
+    };
+
+    inline void TaskExceptionState::StoreException(std::exception_ptr e)
+    {
+        std::lock_guard lock{m_mutex};
+        if(m_exception != nullptr)
```
jaremieromer: Just for style/DRY reasons, call HasException here

jaremieromer (approved):

PR #165, opened 2021-10-06
Changes to accommodate adding textures multiple times.

jaremieromer:
Refactored resources to enable adding textures after textures have initially been added.

### DIFF
--- a/nc/source/graphics/TextureManager.cpp
+++ b/nc/source/graphics/TextureManager.cpp
@@ -16,13 +16,13 @@ namespace nc::graphics
 {
     void LoadTextures(const std::vector<std::string>& paths)
     {
-        IF_THROW(!impl, "graphics::LoadMeshAsset - impl is not set");
+        IF_THROW(!impl, "graphics::LoadTextures - impl is not set");
         impl->LoadTextures(paths);
     }
     
     void LoadTexture(const std::string& path)
     {
-        IF_THROW(!impl, "graphics::LoadMeshAsset - impl is not set");
+        IF_THROW(!impl, "graphics::LoadTexture - impl is not set");
         impl->LoadTexture(path);
     }
     
@@ -52,7 +52,7 @@ namespace nc::graphics
 
     void TextureManager::LoadTexture(const std::string& path)
     {
-        auto sampler = m_graphics->GetBasePtr()->CreateTextureSampler();
+        auto sampler = m_graphics->GetBasePtr()->CreateTextureSampler(); /** @todo: Address this as part of refactor of Base and Resources */
 
         // Parse the file into a Vulkan image object
         if (ResourceManager::TextureExists(path)) return;
--- a/nc/source/graphics/TextureManager.cpp
+++ b/nc/source/graphics/TextureManager.cpp
@@ -1,7 +1,6 @@
 #include "graphics/TextureManager.h"
 #include "graphics/Initializers.h"
 #include "graphics/resources/ResourceManager.h"
-#include "graphics/resources/GraphicsResources.h"
 #include "debug/Utils.h"
 
 #define STB_IMAGE_IMPLEMENTATION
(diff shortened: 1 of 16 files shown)


nc/source/graphics/resources/GraphicsResources.cpp
```diff
@@ -122,14 +127,79 @@ namespace nc::graphics
...
+            m_accessors.emplace(texture.uid, index);
+            m_textures.push_back(std::move(texture));
+        }
+
+        UpdateTextures(graphics);
+    }
+
+    void TexturesData::UpdateTextures(Graphics* graphics)
+    {
+        m_sampler = graphics->GetBasePtr()->CreateTextureSampler();
```
McCallisterRomer, line 169: Do we need to create each time?

nc/source/graphics/resources/GraphicsResources.cpp
```diff
@@ -122,14 +127,79 @@ namespace nc::graphics
...
+
+        std::array<vk::WriteDescriptorSet, 2> writes;
+        vk::DescriptorImageInfo samplerInfo = {};
+        samplerInfo.sampler = m_sampler.get();
+
+        if (!m_texturesInitialized)
+        {
+            m_imageInfos.reserve(m_maxTexturesCount);
```
McCallisterRomer: Why not use the count + value vector constructor here instead of reserve/push_back?

m_imageInfos = std::vector<whatever>(m_maxTexturesCount, m_textures.at(0).imageInfo);

McCallisterRomer: Also, why can't this be done in the constructor? Maybe I'm missing something with the m_textures[0].imageInfo bit...

nc/source/graphics/resources/GraphicsResources.cpp
```diff
@@ -122,14 +127,79 @@ namespace nc::graphics
         return m_accessors.at(uid);
     }
 
+    void TexturesData::AddTexture(Graphics* graphics, Texture texture)
```
McCallisterRomer, line 131: I think these Add methods need to enforce the max size isn't exceeded. Without it there's UB in UpdateTextures.

nc/source/graphics/resources/GraphicsResources.cpp
```diff
@@ -122,14 +127,79 @@ namespace nc::graphics
...
+
+        writes[0].setDstBinding(0);
+        writes[0].setDstArrayElement(0);
+        writes[0].setDescriptorType(vk::DescriptorType::eSampler);
+        writes[0].setDescriptorCount(1);
+        writes[0].setDstSet(m_descriptorSet.get());
+        writes[0].setPBufferInfo(0);
+        writes[0].setPImageInfo(&samplerInfo);
+
+        std::transform(m_textures.cbegin(), m_textures.cend(), m_imageInfos.begin(), [](const auto& texture)
```
McCallisterRomer, line 189: Adding to the comment above about limiting the size of m_textures, the output iterator will write past the end of m_imageInfos if m_textures is larger than m_imageInfos. I know it isn't supposed to be, but I don't see anything making that guarantee. It is probably appropriate to fail in that case, but we should fail by exception, not segfault.

nc/source/graphics/TextureManager.cpp
```diff
@@ -29,94 +34,50 @@ namespace nc::graphics
...
-        {
-            pathsToLoad = ResourceManager::GetTexturePaths();
-            pathsToLoad.insert(paths.end(), paths.begin(), paths.end());
-        }
-        else
-        {
-            pathsToLoad = paths;
-        }
-
         auto sampler = m_graphics->GetBasePtr()->CreateTextureSampler();
```
McCallisterRomer, line 37: I'm not sure what the common practice is with these? Should they be created and thrown away as needed or created once and reused?

McCallisterRomer (approved):
Neat, can't wait to get this into my branch!

PR #175, opened 2021-10-20
Removed global state

McCallisterRomer:
* Replaced the majority of global pointers with abstract classes accessible from the NcEngine base class.
  * Internal classes had to be changed slightly to implement the new interfaces. Some names were also changed to fit a consistent scheme: XXXSystem refers to an API facing type and XXXSystemImpl refers to the internal implementer.
  * Scene::Load was changed to take NcEngine* instead of registry_type*.
  * There are at least two remaining cases to fix: ActiveRegistry() and g_instance for WindowImpl.
* Most nested namespaces were removed from /include. These make sense in /source where there are substantial, self-contained modules like nc::physics, but those lines aren't as clear in /include.
* Moved view/projection/frustum calculations into the Camera component.


ClickerMonkey:
Thats a great improvement!

nc/source/engine/Engine.cpp
```diff
@@ -76,11 +43,47 @@ namespace nc
...
+    }
+
+    void Engine::Shutdown() noexcept
+    {
+        V_LOG("Shutdown NcEngine");
+        ClearState();
+        debug::internal::CloseLog();
+    }
+
+    auto Engine::Audio()       noexcept -> AudioSystem*     { return & m_audioSystem;     }
```
jaremieromer: extra space

project/source/shared/EdgePanCamera.cpp
```diff
@@ -43,6 +40,7 @@ namespace nc::sample
             translation = Vector3::Zero();
         m_lastFrameTranslation = translation;
         translation = translation * dt;
-        camera::GetMainCameraTransform()->Translate(translation);
+        //GetMainCameraTransform()->Translate(translation);
```
jaremieromer: Remove

project/source/shared/SceneNavigationCamera.h
```diff
@@ -35,7 +35,8 @@ namespace nc::sample
 
     inline void SceneNavigationCamera::FrameUpdate(float dt)
     {
-        auto* transform = camera::GetMainCameraTransform();
+        auto* transform = ActiveRegistry()->Get<Transform>(GetParentEntity());
+        //auto* transform = GetMainCameraTransform();
```
jaremieromer: Remove

jaremieromer (approved):

PR #177, opened 2021-10-22
Asset unloading and tests

### LINKED ISSUE #98: GfxResourceManager Improvements
-There should be Unload and potentially UnloadAll members.
-There are lots of string creations/comparisons happening here. At the same time, the strings are very useful for debugging.

McCallisterRomer:
Added some missing stuff to all AssetService implementers
* Implemented missing bulk Load() methods and changed to take span instead of vector
* Implemented all Unload() methods
* Added UnloadAll() method
* Exposed return values through Asset.h functions
* Added tests
  * Didn't add tests for TextureAssetManager. It requires mocking out a good deal of the graphics layer. Another sign we need some more refactoring.

Also, I added a vertexCount member to MeshView. While it can be calculated from the indices, I tried with some questionable data and decided I don't ever want to debug that kind of nonsense.


jaremieromer (approved):
SO GREAT!

PR #180, opened 2021-10-23
Allow user-specified components in registry

McCallisterRomer:
Changed Registry to allow a game to specify its own components to be pooled without rebuilding the library:
* For this to work, Registry cannot be a template type or you get fun ABI breakage like I had the other day, so registry_type_list had to be removed.
* PerComponentStorageBase was added to allow homogeneously storing PerComponentStorage<T>s. There was also some "for all storage do X" things that were previously handled with fold expressions. Because the templates are gone, this is done through virtual functions in PerComponentStorageBase.
* A templated static pointer (non-owning) was added to Registry for locating the correct PerComponentStorage given some T.
* Added Registry::RegisterComponentType<T>(). This creates a storage object for any type satisfying the Component concept and can be called from within a project. All of the engine types are registered in the Registry's constructor.

To make this easier to review, I left the registry_type alias in. I'll do a find + replace on another branch and merge it in to this one.


nc/include/component/Registry.h
```diff
@@ -155,32 +145,25 @@ namespace nc::ecs
...
+            std::vector<std::unique_ptr<PerComponentStorageBase>> m_registeredStorage;
             std::vector<Entity> m_active;
             std::vector<Entity> m_toAdd;
             std::vector<Entity> m_toRemove;
             HandleManager m_handleManager;
+            size_t m_maxEntities;
 
             template<Component T>
-            auto GetStorageFor() -> PerComponentStorage<T>& { return std::get<PerComponentStorage<T>>(m_storage); }
+            inline static PerComponentStorage<T>* m_typedStoragePtr = nullptr;
```
ClickerMonkey, line 161: And this works? This is pretty neat...

McCallisterRomer: Yes. Compiling the project with a new type instantiates the new template, but because it is static, it doesn't affect the binary layout of the Registry object. No need to recompile any engine source.

jaremieromer (approved):

PR #195, opened 2021-12-07
Jare/skyboxes

jaremieromer:
Added ability to add skyboxes to scenes.
Added environment struct that currently holds the selected skybox texture index (if any) and the camera position. Modified shaders to consume new environment struct for camera position.
Added environmental reflections.
Added new buffer type "UniformBuffer" for frequent CPU writes and frequent GPU reads.

Note: Ended up reverting the "Buffer per Frame In Flight" approach. We still use single buffers.


nc/source/graphics/PerFrameRenderState.cpp
```diff
@@ -60,6 +61,17 @@ namespace nc::graphics
             meshes.push_back(renderer.GetMesh());
         }
 
+        environment->Set(cameraPosition);
```
McCallisterRomer: Here you already have set useSkybox above. Also, since the camera position was fetched above, it would be more efficient to load the 'cameraPosition' member variable than to get the position by fetching the camera again.

skyboxMatrix.r[3] = DirectX::XMLoadVector3(&cameraPosition);

nc/include/graphics/Environment.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include "Assets.h"
+
+namespace nc
+{
+    const std::string SkyboxMeshPath = "skybox.nca";
```
McCallisterRomer: I think this can be in the cpp, unless I missed a load.

nc/include/graphics/Environment.h
```diff
@@ -0,0 +1,29 @@
...
+
+    struct EnvironmentData
```
McCallisterRomer: I don't think this is part of the API. If you structure Environment like the other user-facing systems, I think you can keep everything hidden:

// here in include/
class Environment : public SystemBase
{
    public:
        virtual void SetCubeMap(const CubeMapFaces&) = 0;
};

// somewhere in source/
class EnvironmentImpl : public Environment
{
    ...
};

McCallisterRomer: ahh my formatting

McCallisterRomer (approved):

PR #203, opened 2021-12-29
MSVC support + Submodules

McCallisterRomer:
* Moved some folders from /external to separate repos. They are now included as git submodules.
* Fixed all build errors for msvc. Still to do:
  * Figure out which msvc flags and warnings we want.
  * Fix warnings.
  * Visual Studio doesn't clone recursively by default, so cloning must be done on command line.
  * Single solution for engine/editor/sample?
  * CMakePresets.json for configure/build tasks?


nc/source/engine/Engine.cpp
```diff
@@ -128,7 +128,7 @@ namespace nc
         {
             OPTICK_FRAME("Main Thread");
 
-            m_dt = m_frameDeltaTimeFactor * m_time.UpdateTime();
+            m_dt = m_frameDeltaTimeFactor * static_cast<float>(m_time.UpdateTime());
```
McCallisterRomer: Change this so time returns a float. We expect floats pretty much everywhere.

nc/include/debug/Serialize.h
```diff
@@ -1,7 +1,7 @@
 #pragma once
 
 #include "math/Vector.h"
-#include "directx/math/DirectXMath.h"
+#include "directx/Inc/DirectXMath.h"
```
jaremieromer: Why is this folder capitalized?

jaremieromer (approved):

PR #210, opened 2022-01-09
Cal/tests

McCallisterRomer:
* Various tests
* Added ClientObjectProperties to replace the ColliderInteractionType enum. We now look up interactions from a table.
* Added license


nc/include/math/Math.h
```diff
@@ -13,19 +13,10 @@ namespace nc::math
 
     template<std::totally_ordered T>
     constexpr T Min(T a, T b) noexcept { return a < b ? a : b; }
-    
+
     template<std::totally_ordered T>
     constexpr T Max(T a, T b) noexcept { return a > b ? a : b; }
 
-    template<std::floating_point T>
```
jaremieromer, line 20: Why get rid of these?

jaremieromer (approved):
Nice!!!!

PR #236, opened 2022-02-08
Remove AutoComponent

McCallisterRomer:
Removed AutoComponent and added FrameLogic, FixedLogic, and CollisionLogic components as replacements.


docs/EngineComponents.md
```diff
@@ -12,6 +12,50 @@ Engine Components
 * [Camera](#camera)
 * [ParticleEmitter](#particleemitter)
 * [NetworkDispatcher](#networkdispatcher)
+* [FreeComponentGroup](#freecomponentgroup)
+
+### Logic
+---------
+There are three Components designed for quickly hooking up custom logic. They each each hold a callable type which is invoked under specific conditions:
```
McCallisterRomer: 'each' repeated

nc/include/ecs/HandleManager.h
```diff
@@ -3,6 +3,7 @@
 #include "Entity.h"
 
 #include <algorithm>
+#include <span>
```
McCallisterRomer, line 7: pretty sure iterator will be required

nc/include/type/EngineTypes.h
```diff
@@ -13,7 +13,7 @@ namespace nc
 
     /** @todo Only registering names of these: */
     REGISTER_TYPE(AudioSource);
-    REGISTER_TYPE(AutoComponentGroup);
+    REGISTER_TYPE(FreeComponentGroup);
```
McCallisterRomer, line 16: Need to add logic types too.

nc/source/ecs/component/Component.cpp
```diff
@@ -5,9 +5,9 @@
 
 namespace nc
 {
-    void AutoComponent::ComponentGuiElement()
+    void FreeComponent::ComponentGuiElement()
     {
-        ImGui::Text("User AutoComponent");
+        ImGui::Text("Unknown FreeComponent");
```
jaremieromer, line 10: When would this be hit?

jaremieromer (approved):

nc/source/ecs/component/Component.cpp
McCallisterRomer: For any user-defined FreeComponent that doesn't override it. It still shows up in the editor as a component, but we don't know how to present more data than "something exists here."

PR #238, opened 2022-02-13
Cal/registry groups

McCallisterRomer:
* Added a view object for iterating a component pool (this will eventually replace Registry::ViewAll()).
* Added multi_view to iterate entities matching a set of components.

misc:
* Refactored parts of registry and friends to cut down on the number of registry functions.
* Removed two unused entries from StoragePolicy.
* Added a detail directory and namespace for "internal" ecs stuff that still needs to ship in nc/include/.

Also, the new types use snake_case since we're switching. It feels out of place - any thoughts on doing this incrementally vs. all in one go?


nc/include/ecs/detail/EntityStorage.h
```diff
@@ -0,0 +1,79 @@
+#pragma once
+
+#include "HandleManager.h"
+
+#include <algorithm>
+#include <vector>
+
+namespace nc::detail
+{
+    class EntityStorage
```
jaremieromer: Why not snake case?

jaremieromer (approved):

PR #240, opened 2022-02-17
Cal/handle

McCallisterRomer:
Changed return type of Registry::Add<Entity> to a new type, handle, which is more convenient to pass around than an entity + registry* and gives an OO-ish interface for dealing with entities and their components.


McCallisterRomer:
On second thought, the registry shouldn't be returning a handle - it should be left entirely up to the user.

PR #242, opened 2022-02-18
Cal/assert

McCallisterRomer:
Changed name of IF_THROW to NC_ASSERT and enabled it by default in all builds. It can now be disabled through NC_ASSERT_ENABLED.


nc/source/window/WindowImpl.cpp
```diff
@@ -20,19 +20,19 @@ namespace nc::window
     /* Api Function Implementation */
     Vector2 GetDimensions()
     {
-        IF_THROW(!g_instance, "window::GetDimensions - g_instance is not set");
+        NC_ASSERT(g_instance, "window::GetDimensions - g_instance is not set");
```
jaremieromer, line 23: Get rid of these function names

jaremieromer (approved):

PR #246, opened 2022-02-28
Merged descriptor sets from one set per SRS to one combined set per bind frequency.

### LINKED ISSUE #227: Refactor descriptor sets to combine based on bind frequency

jaremieromer:
Shader resource services are no longer one-to-one with descriptor sets, allowing for scale as many integrated devices only allow for four concurrently bound sets.

Now, a single set can have multiple descriptors. Sets are differentiated by bind frequencies.

Shader resource services no longer manage descriptor sets. Now, a new shader resource service just has to register as a descriptor and update the descriptor when changed.


nc/source/graphics/resources/ShaderResourceService.h
```diff
@@ -20,7 +20,7 @@ namespace nc
 
 namespace vk
 {
-    class DescriptorSet;
+    class descriptor_set;
```
McCallisterRomer: I don't think we need to forward declare this.

nc/source/graphics/resources/UniformBuffer.h
```diff
@@ -18,20 +19,50 @@ namespace nc::graphics
...
             void Clear();
 
         private:
             Base* m_base;
             uint32_t m_memoryIndex;
             vk::Buffer m_uniformBuffer;
     };
+
+    template <typename T>
+    void UniformBuffer::Bind(Graphics* graphics, const T& data)
```
McCallisterRomer: Since we only use T for memcpy, you could just take const void* and a size here instead of const T&. Then we can avoid having to include graphics here.

nc/source/graphics/Graphics.cpp
```diff
@@ -105,11 +105,17 @@ namespace nc::graphics
...
+    {
+        return m_shaderResources.get();
+    }
+
     const Vector2 Graphics::GetDimensions() const noexcept
     {
         return m_dimensions;
     }
 
+
```
McCallisterRomer: ws

McCallisterRomer (approved):

PR #270, opened 2022-03-27
Runtime ValidationLayers Enabling

### LINKED ISSUE #268: Specify validation layers at runtime
We have unnecessary tasks and definitions around enabling the Vulkan validation layers. It can all be done at runtime (we end up passing a bool anyway). I propose we add it to the config file and send the value to Base (or whatever it ends up being called).

jaremieromer:
Made enabling Vulkan validation layers a graphics config setting instead of a build flag.
Closes #268


McCallisterRomer (approved):
This is nice! I prefer defaulting use_validation_layers to 0 in all the config files (except maybe sample?).

PR #272, opened 2022-03-27
rename directories

McCallisterRomer:
Just moving/renaming directories.

Unrelated change: Added another constructor to NcError for taking in a detail messaged. (some day std::format will to that for us)


jaremieromer:
Looks good! We should fix the naming of the nested texture folders eventually

jaremieromer (approved):

PR #276, opened 2022-04-03
Initial piece of asset-ized shaders. 

jaremieromer:
Shader assets can now be read into in-memory metadata via a new class, ShaderAssetManager.
Merging as a separate request to keep things organized.
This metadata is not consumed by graphics yet, but the ShaderAssetManager has been fully unit tested.


engine/include/Assets.h
```diff
@@ -114,6 +115,42 @@ namespace nc
         std::string leftPath;
     };
 
+    enum class DescriptorType : uint8_t
```
jaremieromer, line 119: Looking for advice on where to put these. They can't easily go into ShaderAssetManager.h because it would cause circular dependencies, but they shouldn't bloat this guy like they are.

engine/include/Assets.h
```diff
@@ -114,6 +115,42 @@ namespace nc
...
+    };
+
+    enum class ShaderStages : uint8_t
+    {
+        None     = 0,
+        Vertex   = 1,
+        Fragment = 2
+    };
+
+    inline ShaderStages operator|(ShaderStages a, ShaderStages b)
```
McCallisterRomer: There is a macro in utility/EditorUtilities.h for defining bitwise operators. See EngineInitFlags (in NcEngine.h) for an example.

engine/include/Assets.h
```diff
@@ -114,6 +115,42 @@ namespace nc
...
+
+    struct DescriptorManifest
+    {
+        uint32_t setIndex;
+        uint32_t slotIndex;
+        DescriptorType descriptorType;
+        ShaderStages shaderStages;
+    };
+
+    struct ShaderView
```
McCallisterRomer, line 142: Using vectors instead of spans here makes this less of a 'view' and more of the actual asset data. Each copy of this will own its own memory, but the pattern for other assets is the actual memory is owned by the asset manager, and the views just 'points' to that data.

Also, do we need the string here?

engine/source/assets/ShaderAssetManager.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include "assets/AssetService.h"
+
+#include <string>
+#include <vector>
+
+namespace nc
+{    
+    class ShaderAssetManager final : public IAssetService<ShaderView, std::string>
```
McCallisterRomer, line 18: Adding to my comment on the `ShaderView`: I think this file should define a type `ShaderAssetFlyweight` which has the actual data (vectors and such). Then m_shaderViews should be a `std::vector<ShaderAssetFlyweight>` and `Acquire()` still returns a `ShaderView`, which now only has non-owning members (spans and such). Basically what `ConcaveColliderAssetManager` does.

engine/source/assets/ShaderAssetManager.cpp
```diff
@@ -0,0 +1,239 @@
...
+        {
+            file.close();
+            throw nc::NcError("The file was wonky. (failbit set)", path);
+        }
+
+        file.close();
+        return buffer;
+    }
+
+    auto HasValidAssetExtension(const std::string& path) -> bool
```
McCallisterRomer: This should be a shared utility function. Mesh/CubeMap managers also define it.

engine/source/assets/ShaderAssetManager.cpp
```diff
@@ -0,0 +1,239 @@
...
+        auto descriptorType = nc::DescriptorType::None;
+        if (descriptorTypeFromFile == "UniformBuffer")
+        {
+            descriptorType = nc::DescriptorType::UniformBuffer;
+        }
+        else if (descriptorTypeFromFile == "StorageBuffer")
+        {
+            descriptorType = nc::DescriptorType::StorageBuffer;
+        }
+        else descriptorType = nc::DescriptorType::CombinedImageSampler;
```
McCallisterRomer: This should be else if, otherwise the if check below can never be true.

engine/source/assets/ShaderAssetManager.cpp
```diff
@@ -0,0 +1,239 @@
...
+        file >> fragmentShaderPath;
+        fragmentShaderPath = (std::filesystem::path(path) / std::filesystem::path(fragmentShaderPath)).string();
+
+        auto vertexByteCode = ReadSpirvIntoBytes(vertexShaderPath);
+        auto fragmentByteCode = ReadSpirvIntoBytes(fragmentShaderPath);
+
+        auto shaderView = nc::ShaderView
+        {
+            .uid ="",
+            .vertexByteCode = vertexByteCode,
```
McCallisterRomer: We want to move here or just directly call `ReadSpirvIntoBytes()`.

engine/source/assets/ShaderAssetManager.cpp
```diff
@@ -0,0 +1,239 @@
...
+        for (uint32_t i = 0; i < descriptorCount; ++i)
+        {
+            shaderView.descriptors.emplace_back(ReadDescriptor(file));
+        }
+
+        shaderView.uid = uid;
+
+        return shaderView;
+    }
+
```
McCallisterRomer: ws

engine/source/assets/ShaderAssetManager.cpp
```diff
@@ -0,0 +1,239 @@
...
+    {
+    }
+
+    bool ShaderAssetManager::Load(const std::string& path, bool isExternal)
+    {
+        const auto fullPath = isExternal ? path : m_assetDirectory + path;
+
+        auto shaderView = ReadShader(fullPath, path);
+
+        if (IsLoaded(fullPath))
```
McCallisterRomer: Check this before calling `ReadShader()`.

engine/source/assets/ShaderAssetManager.cpp
```diff
@@ -0,0 +1,239 @@
...
+        {
+            return false;
+        }
+
+        for (const auto& manifest : m_shaderViews)
```
McCallisterRomer: Maybe I'm missing the point here, but isn't this just checking `IsLoaded()` again?

engine/source/assets/ShaderAssetManager.cpp
```diff
@@ -0,0 +1,239 @@
...
+                return false;
+            }
+        }
+
+        if (shaderView.vertexByteCode.empty() || shaderView.fragmentByteCode.empty())
+        {
+            return false;
+        }
+
+        m_shaderViews.emplace_back(std::move(shaderView));
```
McCallisterRomer: Although they both work, I would argue `push_back()` is more appropriate here.

engine/source/assets/ShaderAssetManager.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+        return true;
+    }
+
+    bool ShaderAssetManager::Load(std::span<const std::string> paths, bool isExternal)
+    {
+        auto atLeastOneFailure = false;
+        for (const auto& path : paths)
+        {
+            if (IsLoaded(path))
```
McCallisterRomer: No need to check `IsLoaded()` here since the single path version also has to check.

engine/source/assets/ShaderAssetManager.cpp
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    bool ShaderAssetManager::Unload(const std::string& path)
+    {
+        if (!IsLoaded(path))
+        {
+            return false;
+        }
+
+        auto shaderViews = std::vector<ShaderView>();
```
McCallisterRomer: We shouldn't have to move the whole vector to a temp then copy it back. You can use `std::ranges::find_if()` and `m_shaderViews.erase()`.

McCallisterRomer: Well a uid is... unique, right?

engine/source/assets/ShaderAssetManager.cpp
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    void ShaderAssetManager::UnloadAll()
+    {
+        m_shaderViews.clear();
+    }
+
+    auto ShaderAssetManager::Acquire(const std::string& path) const -> ShaderView
+    {
+        for (const auto& shaderView : m_shaderViews)
```
McCallisterRomer: `std::ranges::find_if()` is probably the droid you're looking for.

test/unit/assets/CMakeLists.txt
```diff
@@ -122,4 +122,33 @@ target_link_libraries(AudioClipAssetManager_unit_tests
...
+
+target_compile_options(ShaderAssetManager_unit_tests
+    PUBLIC
+        ${WARNING_FLAGS}
+)
+
+target_link_libraries(ShaderAssetManager_unit_tests
+    PRIVATE
+        gtest
+)
```
McCallisterRomer, line 156: You'll need an `add_test` at the end here.

test/unit/assets/ShaderAssetManager_unit_tests.cpp
```diff
@@ -0,0 +1,80 @@
+#include <array>
+#include <filesystem>
+
+#include "gtest/gtest.h"
+
+#include "assets/ShaderAssetManager.h"
+
+using namespace nc;
+
+class ShaderAssetManager_unit_tests : public ::testing::Test
```
McCallisterRomer, line 10: Nice!!

McCallisterRomer (approved):

engine/include/Assets.h
McCallisterRomer: I'd leave them here for now. We're going to end up with an include/assets/ directory, at which point we can start breaking this stuff into separate files where necessary.

engine/source/assets/ShaderAssetManager.h
McCallisterRomer: Also, I don't remember from a previous conversion if flyweight is the right word here, or if flyweight is synonymous with the view - I'm fine with whatever, but the current pattern is flyweight=owning +internal and view=nonowning + api facing.

PR #288, opened 2022-07-06
Refactored Assets to remove graphics dependency

jaremieromer:
Refactored and renamed Assets and related classes to no longer depend on Graphics.
Note: This PR lays down the groundwork and implements the changes through just one of the AssetServices (MeshAssetManager).
The rest of the services will go in the next review.
Used new Signal pattern to separate Assets from Graphics.


sdk/engine/include/assets/AssetUtilities.h
```diff
@@ -73,6 +75,12 @@ namespace nc
         float maxExtent;
     };
 
+    struct MeshAsset
```
McCallisterRomer, line 78: I'm reasonably sure this won't be needed in the API. I think the only two places that need to know about it are the parsing code and the receiver of the parsed data. I vote to move this to the same file as `GpuAccessorChannels`.

sdk/engine/include/assets/Assets.h
```diff
@@ -0,0 +1,39 @@
+#pragma once
```
McCallisterRomer, line 1: Leaving some notes from our discussion here:

This all should be moved to `source/assets/`. We will eventually add some sort of interface for assets here to replace the free functions and service locator stuff, but that's a story for another story.

I'd also prefer if we undid the `AssetUtilities.h` name change that this file caused.

sdk/engine/include/NcEngine.h
```diff
@@ -69,6 +70,12 @@ namespace nc
              * @return SceneModule*
              */
             virtual auto Scene() noexcept -> SceneModule* = 0;
+
+            /**
+             * @brief Get a pointer to the global Assets instance.
+             * @return Assets*
+             */
+            virtual auto Assets() noexcept -> nc::Assets* = 0;
```
McCallisterRomer, line 78: This can be removed.

sdk/engine/source/assets/AssetServices.h
```diff
@@ -15,7 +15,6 @@ namespace nc
         AssetServices(graphics::Graphics* graphics, const config::ProjectSettings& projectSettings, uint32_t maxTextures)
             : concaveManager{projectSettings.concaveCollidersPath},
               hullManager{projectSettings.hullCollidersPath},
-              meshManager{graphics, projectSettings.meshesPath},
```
McCallisterRomer, line 18: One down!

sdk/engine/source/assets/Assets.cpp
```diff
@@ -1,126 +1,23 @@
...
-    void UnloadAllCubeMapAssets()
-    {
-        return AssetService<CubeMapView>::Get()->UnloadAll();
-    }
-
-} // namespace nc
\ No newline at end of file
+        Assets::Assets(const config::ProjectSettings& projectSettings, const config::MemorySettings& memorySettings)
+            : m_meshManager{std::make_unique<MeshAssetManager>(projectSettings.meshesPath)},
+              m_gpuAccessorChannels{m_meshManager.get()->OnMeshAdd()}
```
McCallisterRomer, line 8: Call to `get()` shouldn't be necessary.

sdk/engine/source/assets/Assets.cpp
```diff
@@ -1,126 +1,23 @@
...
+        {
+        }
+
+        Assets::~Assets() {}
```
McCallisterRomer, line 12: Pop a noexcept on that bad boy. The header is correct. Kind of hoping MSVC would warn on mismatched implicit vs explicit specifiers.

sdk/engine/source/assets/Assets.cpp
```diff
@@ -1,126 +1,23 @@
...
+
+        GpuAccessorChannels* Assets::GpuAccessorChannels()
```
McCallisterRomer, line 14: This could be noexcept, just for kicks.

sdk/engine/source/assets/MeshAssetManager.cpp
```diff
@@ -118,13 +108,20 @@ namespace nc
         };
 
         m_accessors.emplace(path, mesh);
-        UpdateBuffers();
+
+        MeshAsset assetData
```
McCallisterRomer, line 112: Since we're sending the whole boat on each call to `Emit`, I'd prefer this to be named something else. It isn't really a mesh it is all of the meshes.

sdk/engine/source/assets/MeshAssetManager.cpp
```diff
@@ -118,13 +108,20 @@ namespace nc
...
+        {
+            m_vertexData.vertices,
```
McCallisterRomer, line 114: Both vectors get copied here, which is likely to be quite expensive. Maybe the easiest way around that is to use spans instead of vectors inside `MeshAsset` (or whatever it ends up being called). If the receiver needs to copy, so be it, but lets leave it up to them.

sdk/engine/source/assets/MeshAssetManager.cpp
```diff
@@ -196,7 +207,6 @@ namespace nc
         m_accessors.clear();
         m_vertexData.vertices.clear();
         m_indexData.indices.clear();
```
McCallisterRomer, line 209: My first thought here was "shouldn't we call emit here?" - the old comment suggest we should not. It looks like this is just to avoid doing a bunch of set up only to do 0 byte memcpy. Anyways, I like having a comment here to signal that it was intentionally left out.

sdk/engine/source/assets/MeshAssetManager.h
```diff
@@ -1,49 +1,43 @@
...
-        graphics::ImmutableBuffer buffer;
         std::vector<uint32_t> indices;
     };
 
     class MeshAssetManager : public IAssetService<MeshView, std::string>
     {
         public:
-            MeshAssetManager(graphics::Graphics* graphics, const std::string& assetDirectory);
-            ~MeshAssetManager() noexcept;
+            MeshAssetManager(const std::string& assetDirectory);
```
McCallisterRomer, line 25: Oo I love seeing Graphics pointers die.

sdk/engine/source/engine/Runtime.h
```diff
@@ -25,13 +29,16 @@ namespace nc
...
+            time::Time m_time;
+            nc::Random m_random;
+            nc::Assets m_assets;
             Modules m_modules;
             Executor m_executor;
             float m_dt;
             float m_dtFactor;
             bool m_isRunning;
-            unsigned m_currentPhysicsIterations; /** @todo should go in physics_module_impl */
+            unsigned m_currentPhysicsIterations; /** @todo should go in PhysicsModuleImpl */
```
McCallisterRomer, line 41: Thanks

sdk/engine/source/graphics/resources/ImmutableBuffer.cpp
```diff
@@ -57,7 +57,7 @@ namespace nc::graphics
     {
     }
 
-    ImmutableBuffer::ImmutableBuffer(Base* base, GpuAllocator* allocator, const std::vector<Vertex>& data)
+    ImmutableBuffer::ImmutableBuffer(Base* base, GpuAllocator* allocator, const std::vector<nc::graphics::Vertex>& data)
```
McCallisterRomer, line 60: Redundant namespace qualification

sdk/engine/source/graphics/techniques/WireframeTechnique.cpp
```diff
@@ -23,7 +23,6 @@ namespace nc::graphics
       m_pipelineLayout{nullptr}
     {
         CreatePipeline(renderPass);
```
McCallisterRomer, line 25: Curious why this was removed. Its former existence was pretty gross, so I'm not complaining. Does wireframe rendering blow up now if we don't have a cube loaded?

sdk/engine/source/graphics/AssetsStorage.h
```diff
@@ -0,0 +1,51 @@
...
+
+        std::vector<uint32_t> indices;
+        ImmutableBuffer buffer;
+    };
+
+    class AssetsStorage
+    {
+        public:
+            AssetsStorage(Base* base, GpuAllocator* allocator, nc::GpuAccessorChannels* gpuAccessorChannels);
+            ~AssetsStorage() = default;
```
McCallisterRomer, line 32: An implicit destructor should be work here.

sdk/engine/source/assets/Assets.cpp
```diff
@@ -1,126 +1,23 @@
...
+        {
+            return &m_gpuAccessorChannels;
```
McCallisterRomer, line 16: I'm not sure there is a reason to cache this struct as a member. Seems like a 'create once and pass down' sort of deal. Can this function just create on and return by value?

sdk/engine/source/graphics/AssetsStorage.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include "graphics/GpuAllocator.h"
+
+namespace nc::graphics
+{   
+    AssetsStorage::AssetsStorage(Base* base, GpuAllocator* allocator, nc::GpuAccessorChannels* gpuAccessorChannels)
+        : m_base{base},
+          m_allocator{allocator},
+          m_vertexData{m_base, m_allocator},
+          m_indexData{m_base, m_allocator},
+          m_gpuAccessorChannels{gpuAccessorChannels},
```
McCallisterRomer, line 13: I don't see a need to keep this as a member.

sdk/engine/source/graphics/AssetsStorage.cpp
```diff
@@ -0,0 +1,65 @@
+#include "assets/Assets.h"
```
McCallisterRomer, line 1: Some wacky ws on this file

sdk/engine/source/graphics/AssetsStorage.cpp
```diff
@@ -0,0 +1,65 @@
...
+        m_indexData.buffer = ImmutableBuffer(m_base, m_allocator, indices);
+    }
+
+    void AssetsStorage::UpdateMeshBuffer(const MeshAsset& meshAsset)	
+    {	
+        SetVertexData(meshAsset.vertices);	
+        SetIndexData(meshAsset.indices);	
+    }
+
+    VertexData::VertexData(Base* base, GpuAllocator* allocator)
```
McCallisterRomer, line 46: Use a member initializer list here and in `IndexData`.

sdk/engine/source/graphics/Graphics.h
```diff
@@ -13,7 +13,11 @@
 #include <memory>
 #include <mutex>
 
-namespace nc { struct AssetServices; }
+namespace nc
```
McCallisterRomer, line 16: ws at the end of lines 16-19

sdk/engine/source/graphics/Renderer.cpp
```diff
@@ -96,7 +97,7 @@ namespace nc::graphics
         m_renderPasses->RegisterTechnique<UiTechnique>(RenderPassManager::LitShadingPass);
     }
 
-    vk::CommandBuffer* Renderer::BeginFrame(Commands* commands, AssetServices* assetServices, uint32_t currentSwapChainImageIndex)
+    vk::CommandBuffer* Renderer::BeginFrame(Commands* commands, AssetServices* assetServices, AssetsStorage* assetsStorage, uint32_t currentSwapChainImageIndex)
```
McCallisterRomer, line 100: It looks like `AssetsStorage` can be const here and above in `Record`.

sdk/engine/source/graphics/Renderer.cpp
```diff
@@ -107,9 +108,11 @@ namespace nc::graphics
         SetViewportAndScissor(cmd, m_dimensions);
 
         vk::DeviceSize offsets[] = { 0 };
-        auto vertexBuffer = assetServices->meshManager.GetVertexBuffer();
+        auto& vertexData = assetsStorage->GetVertexData();
```
McCallisterRomer, line 111: vertexData and indexData should be const

sdk/test/unit/assets/MeshAssetManager_unit_tests.cpp
```diff
@@ -1,28 +1,15 @@
 #include "gtest/gtest.h"
```
McCallisterRomer, line 1: Oh my god, I don't see a single graphics include.

McCallisterRomer (changes requested):

sdk/engine/source/assets/MeshAssetManager.cpp
McCallisterRomer: Clarification: I'm referring to the `MeshAsset` type not the variable. Maybe that's already obvious.

sdk/engine/source/graphics/techniques/WireframeTechnique.cpp
jaremieromer: Removed because in previous version not using signals, the graphics module was constructed before the assets utility in the runtime. It was causing problems trying to load a mesh asset in the ctor here.
WireframeTechnique does not need it here anyway

sdk/test/unit/assets/MeshAssetManager_unit_tests.cpp
jaremieromer: huzzah

PR #297, opened 2022-07-31
ModuleRegistry

McCallisterRomer:
Many files changed, but most of them are minor usage updates. All interesting stuff is under `sdk/engine`.

Added `ModuleRegistry` as a more generic way to deal with modules. There is a bit more to do before a user can properly make a custom module, but this allows replacing hard-coded queries with templated ones(`NcEngine::Physics()` becomes `ModuleRegistry::Get<PhysicsModule>()`).

Also, `Scene::Load(NcEngine*)` was changed to `Scene::Load(Registry*, ModuleProvider)`, where `ModuleProvider` is a wrapper for `ModuleRegistry::Get<T>()`. The `NcEngine` pointer was only passed to give access to the "get module x" functions, which now live in `ModuleRegistry`.


sdk/editor/source/framework/ProjectManager.cpp
```diff
@@ -9,6 +9,9 @@
 #include "serialize/SceneWriter.h"
 #include "EditorScene.h"
```
jaremieromer, line 11: Extra newline here

sdk/engine/include/NcEngine.h
```diff
@@ -36,39 +33,13 @@ namespace nc
 
             /**
              * @brief Get a pointer to the active registry.
-             * @return Registry*
              */
-            virtual auto Registry() noexcept -> nc::Registry* = 0;
+            virtual auto GetRegistry() noexcept -> Registry* = 0;
```
jaremieromer, line 37: We keep going back and forth between prefixing our getters with "Get". I thought we'd last decided to leave out the "Get". Why change now?

sdk/sample/source/scenes/ClickEvents.cpp
```diff
@@ -48,18 +51,19 @@ namespace
...
 {
-    void ClickEvents::Load(NcEngine* engine)
+    void ClickEvents::Load(Registry* registry, ModuleProvider modules)
     {
-        auto* registry = engine->Registry();
-
         // Setup
-        m_sceneHelper.Setup(engine, true, false, Widget);
+        m_sceneHelper.Setup(registry, modules, true, false, Widget);
+        auto* graphics = modules.Get<GraphicsModule>();
```
jaremieromer, line 58: This is nice

jaremieromer (approved):
This is such a big step in the right direction!

Are any modules required (going forward) to prevent a crash?
(aka - can the user unregister every module?)

 There could be user modules dependent on one another (think Graphics requiring GpuAccessorSignals from Assets now, even though Assets is not currently a module), and if so - how do we want to represent that and enforce the relationship?

PR #313, opened 2022-09-14
Logging

McCallisterRomer:
* Logging improvements: thread safety, format strings, buffering, more log levels
* Started using `std::format` here and there - MSVC requires latest flag (which is not the same c++20), so I've updated CMake to require c++23.
* Split `NcError` and logging stuff into separate files.


include/utility/detail/LogInternal.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#include <string_view>
+
+namespace nc::utility::detail
+{
+void InitializeLog(std::string_view path);
```
jaremieromer, line 7: "Initialize" and "Close" is sufficient here, "..Log" is redundant IMO.

source/engine/scene/SceneManager.cpp
```diff
@@ -1,6 +1,6 @@
 #include "SceneManager.h"
 #include "module/ModuleProvider.h"
-
+#include "utility/Log.h"
 namespace nc::scene
```
jaremieromer, line 5: Add a newline after this include

jaremieromer (approved):
Looks good! I'm not sure why the tests are failing though.

PR #315, opened 2022-09-26
Jare per frame gpu context 2 of 2

jaremieromer:
### Goal
Complete removal of all of the constructs spread throughout the graphics classes that exist on a per-frame basis and extract them into a single `PerFrameGpuContext` class. 

### This PR
This PR adds a `CommandBuffer` and a `CommandPool` to the `PerFrameGpuContext`. 
To facilitate that work, a few related concepts were pulled out of `Base` and into `GpuAllocator`, `Commands` and a new struct called `Engine`. 

 - Concepts related to executing commands against a `CommandBuffer` were moved out of `Base` and into `Commands`. 
 - Concepts related to memory management with buffers were moved out of `Base` and into `GpuAllocator`.
 - The four fundamental building blocks of Vulkan calls (`vk::Device`, `vk::PhysicalDevice`, `vk::Instance`, `vk::SurfaceKHR`) were moved out of `Base` and into `Engine`.

### Future PR
In order to reduce the amount of PR noise, I left the `GetDevice()`, `GetPhysicalDevice()`, `GetSurface()` and `GetInstance()` calls in `Base`. A future PR will remove all of these and update all references.

I also really dislike the way ImGui is initialized (now in `Renderer`) and want to find a place for it.

Big picture, I want to move towards `Graphics.h` being a GPU framework agnostic interface, with pointers to the different areas of graphics (`FrameManager`, `Renderer`, `ShaderResources`, etc) and have those classes be Vulkan or DX specific.


source/engine/graphics/vk/Engine.cpp
```diff
@@ -0,0 +1,222 @@
...
+#include "config/Config.h"
+#include "utility/NcError.h"
+#include "QueueFamily.h"
+#include "Swapchain.h"
+
+#include <set>
+
+namespace
+{
+const std::vector<const char*> DeviceExtensions = { VK_KHR_SWAPCHAIN_EXTENSION_NAME };
```
McCallisterRomer: I know these are just copy-pasted, but I think they're worth cleaning up. They can all be constexpr std::arrays and AAA (almost always auto) declaration style is sort of the default in the engine:

`constexpr auto DeviceExtensions = std::array<const char*>{...};`

source/engine/graphics/vk/Engine.cpp
```diff
@@ -0,0 +1,222 @@
...
+const std::vector<const char*> GlobalExtensions = { VK_KHR_WIN32_SURFACE_EXTENSION_NAME, VK_KHR_SURFACE_EXTENSION_NAME };
+const std::vector<const char*> ValidationLayers = { "VK_LAYER_KHRONOS_validation" };
```
McCallisterRomer: Not a big deal, is there not a macro for this one?

source/engine/graphics/vk/ImmutableImage.cpp
```diff
@@ -1,6 +1,8 @@
 #include "ImmutableImage.h"
 #include "graphics/Base.h"
 
+typedef unsigned char stbi_uc;
```
McCallisterRomer: I'm not sure there's a good reason to use this typedef. I would just use `unsigned char*` instead, like the header does.

source/engine/graphics/vk/PerFrameGpuContext.h
```diff
@@ -4,26 +4,32 @@
...
     public:
-        PerFrameGpuContext(vk::Device device);
+        PerFrameGpuContext(vk::Device logicalDevice, vk::PhysicalDevice physicalDevice, vk::SurfaceKHR surface);
 
         void Wait(); // Waits until the fence has signaled
         void Reset() noexcept; // Resets the fence
         vk::Fence Fence() const noexcept { return m_inFlightFence.get(); }
-        vk::Semaphore ImageAvailableSemaphore() const noexcept { return m_imageAvailableSemaphore.get(); }
-        vk::Semaphore RenderFinishedSemaphore() const noexcept { return m_renderFinishedSemaphore.get(); }
+        vk::Semaphore* ImageAvailableSemaphore() noexcept { return &(m_imageAvailableSemaphore.get()); }
```
McCallisterRomer: We briefly discussed this. I'd like to over a bit more.

McCallisterRomer: pointers to these handles, that is

source/engine/graphics/FrameManager.h
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include "vk/PerFrameGpuContext.h"
```
McCallisterRomer, line 3: I feel like this type is trying to go in two different directions. It both includes and forward declares PerFrameGpuContext. It is outside the vk folder and uses unique_ptr to try and hide that stuff, but it takes vk types in its constructor. If it is trying to hide all vulkan types, the vector may need to be passed in through the constructor to avoid exposing those vk:: params.

source/engine/graphics/Renderer.cpp
```diff
@@ -59,10 +91,13 @@ Renderer::~Renderer() noexcept
     m_renderPasses.reset();
 }
 
-void Renderer::Record(Commands* commands, const PerFrameRenderState& state, const MeshStorage& meshStorage, uint32_t currentSwapChainImageIndex)
+void Renderer::Record(PerFrameGpuContext* currentFrame, const PerFrameRenderState& state, const MeshStorage& meshStorage, uint32_t currentSwapChainImageIndex)
 {
+    auto cmd = currentFrame->CommandBuffer();
+
     OPTICK_CATEGORY("Renderer::Record", Optick::Category::Rendering);
```
McCallisterRomer: Try to keep optick macros at the very top of functions so they don't miss anything

source/engine/graphics/Graphics.cpp
```diff
@@ -192,32 +156,46 @@ namespace nc::graphics
...
     {
         OPTICK_CATEGORY("Graphics::Draw", Optick::Category::Rendering);
         if (m_isMinimized) return;
 
-        m_renderer->Record(m_commands.get(), state, m_gpuAssetsStorage.get()->meshStorage, m_imageIndex);
+        m_renderer->Record(m_frameManager->CurrentFrameContext(), state, m_gpuAssetsStorage.get()->meshStorage, m_imageIndex);
 
         // Executes the command buffer to render to the image
-        RenderToImage(m_imageIndex);
+        OPTICK_CATEGORY("Graphics::RenderToImage", Optick::Category::Rendering);
```
McCallisterRomer: We should only have one optick tag per scope. We don't want 3 different profiling events that fall under the same function. I think (?) we're fine just profiling `Draw()` here, but if we really want to profile the 3 sections, this strongly suggests the function should be split up into 3 functions.

McCallisterRomer (approved):

source/engine/graphics/vk/Engine.cpp
jaremieromer: There is not, no

PR #318, opened 2022-11-13
Graphics cleanup

jaremieromer:
### Goal
Slim down and further detangle graphics classes by removing `GpuOptions` (previously `Base`) and `Graphics` from it's position in many CTOR/function parameters.

### This PR
Removed all `Graphics` pointers, no more passing around of the mega class.
`Base`is no more, `GpuOptions` only passed where GPU options data is needed.
Fixed the ImGui resize bug introduced a few PRs ago.
 Progress towards better flow direction of data and params (removed many "GetXClassPointer" methods).
 Removed the `SetClearColor(f, f, f, f)`functionality as it was a lot of code and would never be exercised in a real game.


source/engine/graphics/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 target_sources(${NC_ENGINE_LIB}
     PRIVATE
-        Base.cpp
+        GpuOptions.cpp
```
McCallisterRomer: I'd prefer to keep alphabetical order here

source/engine/graphics/GpuOptions.cpp
```diff
@@ -0,0 +1,86 @@
...
+      m_depthFormat{},
+      m_samplesCount{},
+      m_samplesInitialized{false},
+      m_gpuProperties{physicalDevice.getProperties()}
+{
+}
+
+vk::SampleCountFlagBits GpuOptions::GetMaxSamplesCount()
+{
+    if (m_samplesInitialized == true)
```
McCallisterRomer: Can we do the computation below in the constructor, and get rid of the bool gate here?

source/engine/graphics/GpuOptions.cpp
```diff
@@ -0,0 +1,86 @@
...
+    else if (counts & vk::SampleCountFlagBits::e4)  m_samplesCount = vk::SampleCountFlagBits::e4;
+    else if (counts & vk::SampleCountFlagBits::e2)  m_samplesCount = vk::SampleCountFlagBits::e2;
+    else m_samplesCount = vk::SampleCountFlagBits::e1;
+
+    return m_samplesCount;
+}
+
+const vk::Format& GpuOptions::GetDepthFormat() noexcept
+{
+    if (m_depthInitialized == true)
```
McCallisterRomer: Same as above

source/engine/graphics/GpuOptions.h
```diff
@@ -0,0 +1,25 @@
+#pragma once
+
+#include "vulkan/vk_mem_alloc.hpp"
+
+namespace nc::graphics
+{
+    /** This class deals with rendering options that are dependent on properties of the GPU (the physical device). */
+    class GpuOptions
```
McCallisterRomer: Unindent class

source/engine/graphics/GpuOptions.h
```diff
@@ -0,0 +1,25 @@
...
+    {
+        public:
+            GpuOptions(vk::PhysicalDevice physicalDevice);
+
+            const vk::Format& GetDepthFormat() noexcept;
```
McCallisterRomer: If the bools go away, both `GetDepthFormat()` and `GetMaxSamplesCount()` can be const + noexcept.

McCallisterRomer: Since `vk::Format` is just an enum, there's probably not a reason to return by const&.

source/engine/graphics/GpuOptions.cpp
```diff
@@ -0,0 +1,86 @@
...
+#include "config/Config.h"
+#include "utility/NcError.h"
+
+#include <string>
+
+namespace
+{
+vk::Format QueryDepthFormatSupport(vk::PhysicalDevice physicalDevice)
+{
+    std::vector<vk::Format> depthFormats = { vk::Format::eD32SfloatS8Uint, vk::Format::eD32Sfloat, vk::Format::eD24UnormS8Uint, vk::Format::eD16UnormS8Uint, vk::Format::eD16Unorm };
```
McCallisterRomer: I know just a copy+paste, but this could be `constexpr auto depthFormats = std::array<...>{...}`.

source/engine/graphics/GpuOptions.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+    for (auto& format : depthFormats)
```
McCallisterRomer: No reason to grab by reference for trivial types. Also, consider `std::ranges::find_if()`.

source/engine/graphics/Graphics.cpp
```diff
@@ -171,19 +125,20 @@ namespace nc::graphics
     // Note: All calls below are asynchronous fire-and-forget methods. A maximum of MaxFramesInFlight sets of calls will be running at any given time.
     void Graphics::Draw(const PerFrameRenderState& state)
     {
+        auto* currentFrame = m_frameManager->CurrentFrameContext();
```
McCallisterRomer: Try to keep optick macros at the top of functions otherwise stuff doesn't get profiled.

source/engine/graphics/Graphics.h
```diff
@@ -64,24 +45,20 @@ namespace nc::graphics
             void RecreateSwapchain(Vector2 dimensions);
 
             camera::MainCamera* m_mainCamera;
-            std::unique_ptr<Core> m_engine;
-            std::unique_ptr<Base> m_base;
+            std::unique_ptr<Core> m_core;
+            std::unique_ptr<GpuOptions> m_gpuOptions;
```
McCallisterRomer, line 49: Hey, that's p. neat.

source/engine/graphics/Renderer.h
```diff
@@ -19,21 +19,18 @@ class ParticleRenderer;
...
-class RenderPassManager;
 
 class Renderer
 {
     public:
-        Renderer(graphics::Graphics* graphics,
-                 vk::Device device, 
-                 ShaderResourceServices* shaderResources,
-                 Vector2 dimensions);
+        Renderer(vk::Device device, Swapchain* swapchain, GpuOptions* gpuOptions, GpuAllocator* gpuAllocator, ShaderResourceServices* shaderResources, Vector2 dimensions);
```
McCallisterRomer, line 33: Nice!

source/engine/graphics/ShaderUtilities.cpp
```diff
@@ -1,18 +1,17 @@
...
+    vk::ShaderModule CreateShaderModule(vk::Device device, const std::vector<uint32_t>& code)
     {
         vk::ShaderModuleCreateInfo createInfo{};
         createInfo.setCodeSize(code.size()*sizeof(uint32_t));
         createInfo.setPCode(code.data());
         vk::ShaderModule shaderModule;
-        if (base->GetDevice().createShaderModule(&createInfo, nullptr, &shaderModule) != vk::Result::eSuccess)
+        if (device.createShaderModule(&createInfo, nullptr, &shaderModule) != vk::Result::eSuccess)
         {
             throw NcError("Failed to create shader module.");
```
McCallisterRomer, line 16: For the future: it would be nice to have some sort of metadata about what 'code' failed to pack into this error.

source/engine/graphics/resources/RenderPassManager.h
```diff
@@ -18,13 +20,15 @@ namespace nc::graphics
         vk::UniqueFramebuffer frameBuffer;
     };
 
+    constexpr std::array<float, 4> ClearColor = {0.1f, 0.1f, 0.1f, 0.1f};
```
McCallisterRomer: Can this go in the cpp? Not sure if it is references elsewhere.

source/engine/graphics/techniques/ShadowMappingTechnique.cpp
```diff
@@ -108,8 +81,26 @@ namespace nc::graphics
         pipelineCreateInfo.setBasePipelineHandle(nullptr); // Graphics pipelines can be created by deriving from existing, similar pipelines. 
         pipelineCreateInfo.setBasePipelineIndex(-1); // Similarly, switching between pipelines from the same parent can be done.
 
-        m_pipeline = m_base->GetDevice().createGraphicsPipelineUnique(nullptr, pipelineCreateInfo).value;
-        m_base->GetDevice().destroyShaderModule(vertexShaderModule, nullptr);
+        m_pipeline = device.createGraphicsPipelineUnique(nullptr, pipelineCreateInfo).value;
+        device.destroyShaderModule(vertexShaderModule, nullptr);
+    }
+
+    ShadowMappingTechnique::~ShadowMappingTechnique() noexcept
```
McCallisterRomer, line 87: I see all of the techniques have destructors like this now. The vk handle types will free these automatically. I'm assuming the order of destruction matters here, but that should be handled by reordering within the class, if possible.

source/engine/graphics/techniques/WireframeTechnique.cpp
```diff
@@ -102,9 +73,34 @@ namespace nc::graphics
...
+        m_pipeline.reset();
+        m_pipelineLayout.reset();
+    }
+
+    bool WireframeTechnique::CanBind(const PerFrameRenderState& frameData)
+    {
+        return frameData.colliderDebugWidget.has_value() 
+        #ifndef NC_DEBUG_RENDERING_ENABLED
+        ;
+        #else
```
McCallisterRomer, line 94: I forgot about this gem

McCallisterRomer (approved):

source/engine/graphics/techniques/ShadowMappingTechnique.cpp
McCallisterRomer: You're right, these have been around for a while.

PR #319, opened 2022-11-17
Graphics folder and subfolder restructuring

jaremieromer:
### Goal
Restructure graphics classes into logical groups to allow for easier reading and finding of classes. Remove the vk folder.

### This PR
This PR is solely name changing and file/folder moving. 
The subfolders in the graphics folder are now:

 - buffers
 - components
 - cubemaps
 - meshes
 - renderpasses
 - shaders
 - techniques
 - textures

All classes have been moved into subfolders (or kept in the main graphics folder) based on their functionality. 

**Example:** `ImmutableImage` is a piece of a `TextureBuffer`, and not used anywhere else. It has been moved out of resources (which no longer exists) and into textures.
**Example:**  All the implementors of IShaderResourceService were scattered around and called `XManager`. (ie `CubeMapManager`). These classes are renamed to `XShaderResource` and moved into the shaders subfolder.
**Example:** `ImmutableBuffer`, `UniformBuffer`, and `WriteableBuffer`were moved out of resources and into the buffers subfolder. The resources subfolder was a catch-all that was so general it served no purpose.

There are no functional changes in the PR.


source/engine/graphics/shaders/EnvironmentDataShaderResource.h
```diff
@@ -1,16 +1,16 @@
...
-#include "ShaderDescriptorSets.h"
-#include "UniformBuffer.h"
+#include "graphics/buffers/UniformBuffer.h"
+#include "graphics/shaders/ShaderResourceService.h"
+#include "graphics/shaders/ShaderDescriptorSets.h"
 
 namespace nc::graphics
 {
-    class EnvironmentDataManager : public IShaderResourceService<EnvironmentData>
+    class EnvironmentDataShaderResource : public IShaderResourceService<EnvironmentData>
```
McCallisterRomer: To clarify my mental model of these guys: is a shader resource conceptually the whole buffer, rather than the individual items in the buffer? If so, I think this rename makes things much clearer.

What do you think about renaming `IShaderResourceService` to `IShaderResource`? There is currently a using alias in the mix named `ShaderResourceService`, whose purpose is to handle the service locator side of things. The interface doesn't directly have anything to do with services.

Throwing the rest of my comments here as I can't comment on files that were just moved:
- The types in `shaders/Environment.h` are not in the graphics namespace.
- `graphics/SceneNavigationCamera.cpp` should also go in `components`.

McCallisterRomer (approved):

source/engine/graphics/shaders/EnvironmentDataShaderResource.h
McCallisterRomer: Building off of the first and second items: There is also a concept named `ShaderResource` which maps to the individual types, not buffers. Assuming the answer to my first question is yes, then `ShaderResource` is an awkward name for the concept. I don't know what to call it. Maybe you have ideas, or we can discuss.

PR #334, opened 2023-02-15
Jare shadows fixes

jaremieromer:
### Goal

Add functionality for multiple light sources to cast shadows. Do this by creating a shadow mapping render pass per light source. 

### Changes
Primary Changes:

 - Added a new class (`Lighting.h/cpp`) to `Graphics`. This class is responsible for creating and destroying shadow mapping render passes. It will do so for all light sources (right now we only have point lights).
 - In modifying `Renderer` and `RenderPassManager` to support this new 'n render passes' flow, I realized the classes had no clear dividing line and scrapped both, combining minimal features into new class called `RenderGraph`.
 - There was a LOT going on in `ShadowMapShaderResource.cpp` that did not need to be in there.
 - Made some more progress in simplifying `Graphics::Resize()`

Cleanup Changes:
-  Modified the folder structure of the shaders in the sample and the shaders in the engine so they are now the same structure.
- Started cleanup of Graphics' use of Imgui, will continue in the Imgui bug fix PR. This one isolates much of the Imgui graphics requirements to a single class (`Imgui.h/cpp`).
- Changed the Vulkan resources in `Swapchain` to `vk::Unique_` RAII types. Various other cleanup in that class as well.
- There was a nasty bit of work in `ShaderResoures` where it owned `ShaderDescriptorSets` and then passed it out to callers with a pointer, it was really tangled up. `ShaderResources` and `ShaderDescriptorSets` are now siblings in `Graphics`. `ShaderResources` is now a struct.

Comments:
Most of the changed files are just the shader directory restructure and can be ignored.
Ignore the commits - they are not helpful. Most were when I was switching back and forth on two PCs.


source/engine/graphics/renderpasses/RenderPass.cpp
```diff
@@ -202,13 +203,97 @@ Subpass::Subpass(const AttachmentSlot& depthAttachment)
...
+                       std::string uid_,
+                       std::vector<AttachmentSlot> attachmentSlots,
+                       std::vector<Subpass> subpasses,
+                       std::vector<std::unique_ptr<Attachment>> attachments_,
+                       const AttachmentSize& size,
+                       ClearValueFlags_t clearFlags_)
+    : priority{priority_},
       uid{std::move(uid_)},
-      clearFlags{clearFlags_}
+      renderPass{std::move(CreateRenderPass(attachmentSlots, subpasses, device))},
```
McCallisterRomer: The returned object is already an r value, so you shouldn't have to move here.

source/engine/graphics/renderpasses/RenderPass.cpp
```diff
@@ -202,13 +203,97 @@ Subpass::Subpass(const AttachmentSlot& depthAttachment)
...
+      attachments{std::move(attachments_)}{}
+
+auto GetFrameBuffer(nc::graphics::RenderPass *renderPass, uint32_t index) -> nc::graphics::FrameBuffer *
+{
+    const auto frameBufferPos = std::ranges::find_if(renderPass->frameBuffers, [index](const auto &frameBuffer)
+                                                     { return (frameBuffer->index == index); });
+
+    if (frameBufferPos == renderPass->frameBuffers.end())
+    {
+        return renderPass->frameBuffers[0].get();
```
McCallisterRomer: This is unsafe if frameBuffers is empty. If its reasonable to expect it could be empty (I doubt it?), do a check first. Otherwise, use `at(0)` instead, so at least we'll blow up if logic is bad elsewhere.

Also, it isn't obvious to me why we can return the first one if we don't find the one we're looking for. Maybe a comment?

source/engine/graphics/renderpasses/RenderPass.cpp
```diff
@@ -202,13 +203,97 @@ Subpass::Subpass(const AttachmentSlot& depthAttachment)
...
+    }
+
+    return frameBufferPos->get();
+}
+
+void RegisterAttachments(vk::Device device, nc::graphics::RenderPass *renderPass, std::span<const vk::ImageView> attachmentHandles, const nc::Vector2 &dimensions, uint32_t index)
 {
+    if (const auto frameBufferPos = std::ranges::find_if(renderPass->frameBuffers, [index](const auto &frameBuffer)
```
McCallisterRomer: This is sort of hard to read with everything in the if. Maybe pull the search out.

source/engine/graphics/renderpasses/RenderPass.cpp
```diff
@@ -202,13 +203,97 @@ Subpass::Subpass(const AttachmentSlot& depthAttachment)
...
+                                                         { return (frameBuffer->index == index); });
+        frameBufferPos != renderPass->frameBuffers.end())
+    {
+        *frameBufferPos = std::move(renderPass->frameBuffers.back());
+        renderPass->frameBuffers.pop_back();
+    }
+    
+    auto frameBuffer = std::make_unique<FrameBuffer>();
+    frameBuffer->attachmentHandles = std::move(attachmentHandles);
```
McCallisterRomer: `attachmentHandles` and `frameBuffer->attachmentHandles` are both spans, so there's no need to move. But also nobody here owns anything. What guarantees are there about the lifetime of these ImageViews? It looks like something that could become a bug pretty easily.

source/engine/graphics/shaders/ShaderResources.h
```diff
@@ -11,32 +12,25 @@
...
-        ShaderResources(vk::Device device, GpuAllocator* allocator, const config::MemorySettings& memorySettings, Vector2 dimensions)
-            : m_shaderDescriptorSets{device},
-              m_objectDataShaderResource{0, allocator, &m_shaderDescriptorSets, memorySettings.maxRenderers},
-              m_pointLightShaderResource{1, allocator, &m_shaderDescriptorSets, memorySettings.maxPointLights},
-              m_textureShaderResource{2, &m_shaderDescriptorSets, memorySettings.maxTextures},
-              m_shadowMapShaderResource{device, 3, allocator, &m_shaderDescriptorSets, dimensions },
-              m_cubeMapShaderResource{device, 4, &m_shaderDescriptorSets, memorySettings.maxTextures}, // @todo make separate entry for cubeMaps
-              m_environmentDataShaderResource{5, allocator, &m_shaderDescriptorSets}
+
+        ShaderResources(vk::Device device, ShaderDescriptorSets* shaderDescriptorSets, Registry* registry, GpuAllocator* allocator, const config::MemorySettings& memorySettings, Vector2 dimensions)
```
McCallisterRomer: I don't think `Registry` is used here.

source/engine/graphics/shaders/ShadowMapShaderResource.cpp
```diff
@@ -3,97 +3,77 @@
...
-        m_dimensions = data.at(0).dimensions;
+        m_imageInfos.clear();
+        m_imageInfos.reserve(data.size());
 
-        m_depthStencil.reset();
-        m_depthStencil = std::make_unique<RenderTarget>(m_device, m_allocator, m_dimensions, true, vk::SampleCountFlagBits::e1, vk::Format::eD16Unorm);
-
-        auto descriptorImageInfo = CreateDescriptorImageInfo(m_sampler.get(), m_depthStencil->GetImageView(), vk::ImageLayout::eDepthAttachmentStencilReadOnlyOptimal);
-        m_imageInfos = std::vector<vk::DescriptorImageInfo>(1, descriptorImageInfo);
+        for (uint32_t i = 0; i < static_cast<uint32_t>(data.size()); ++i)
```
McCallisterRomer: Range-based for or maybe transform?

source/engine/graphics/shaders/ShadowMapShaderResource.cpp
```diff
@@ -3,97 +3,77 @@
...
+            m_imageInfos.push_back(CreateDescriptorImageInfo(m_sampler.get(), data[i].imageView, vk::ImageLayout::eDepthAttachmentStencilReadOnlyOptimal));
+        }
 
         m_descriptors->UpdateImage
-        (
-            BindFrequency::per_frame,
+        (   BindFrequency::per_frame,
             m_imageInfos,
-            1,
+            static_cast<uint32_t>(data.size()),
```
McCallisterRomer: While they're the same, I think `m_imageInfos.size()` would be better here.

source/engine/graphics/techniques/ShadowMappingTechnique.cpp
```diff
@@ -121,18 +122,15 @@ namespace nc::graphics
         auto pushConstants = ShadowMappingPushConstants{};
 
         // We are rendering the position of each mesh renderer's vertex in respect to each point light's view space.
-        for (const auto& pointLightVP : frameData.pointLightVPs)
-        {
-            pushConstants.lightViewProjection = pointLightVP;
+        pushConstants.lightViewProjection = frameData.pointLightVPs[m_shadowCasterIndex];
```
McCallisterRomer: There's nothing in the immediate scope to guarantee we're not going out-of-bounds here, so I prefer `at(m_shadowCasterIndex)`.

source/engine/graphics/Graphics.h
```diff
@@ -2,56 +2,62 @@
 
 #include "assets/AssetManagers.h"
 #include "camera/MainCamera.h"
-#include "GpuAllocator.h"
+#include "ecs/Registry.h"
```
McCallisterRomer: Can remove

source/engine/graphics/Graphics.h
```diff
@@ -2,56 +2,62 @@
...
         public:
-            Graphics(camera::MainCamera* mainCamera, const nc::GpuAccessorSignals& gpuAccessorSignals, HWND hwnd, HINSTANCE hinstance, Vector2 dimensions);
+            Graphics(camera::MainCamera* mainCamera, Registry* registry, const nc::GpuAccessorSignals& gpuAccessorSignals, HWND hwnd, HINSTANCE hinstance, Vector2 dimensions);
             ~Graphics() noexcept;
 
-            void OnResize(float width, float height, float nearZ, float farZ, WPARAM windowArg);
-            void Clear();
-            void InitializeUI();
+            void OnResize(float width, float height, float nearZ, float farZ, const WPARAM windowArg);
+            void Clear() const;
```
McCallisterRomer: I don't think these two functions should be made const, even if they can be. They compile because they only modify things that are pointed to by members, but if someone had a `const& Graphics`, I wouldn't want them to be able to call things like `Clear` and `Initialize`.

source/engine/graphics/renderpasses/RenderPass.h
```diff
@@ -30,43 +32,62 @@ enum class AttachmentType : uint8_t
...
+               std::vector<std::unique_ptr<Attachment>> attachments_,
+               const AttachmentSize& size,
+               ClearValueFlags_t clearFlags_);
+    uint8_t priority;
     std::string uid;
+    vk::UniqueRenderPass renderPass;
+    AttachmentSize attachmentSize;
     ClearValueFlags_t clearFlags;
+    std::vector<std::unique_ptr<ITechnique>> techniques{};
+    std::vector<std::unique_ptr<Attachment>> attachments{};
```
McCallisterRomer: attachments and frameBuffers probably don't need to be unique_ptrs, although techniques makes sense.

source/engine/graphics/Graphics.h
```diff
@@ -2,56 +2,62 @@
...
+            auto FrameBegin() -> bool;
             void Draw(const PerFrameRenderState& state);
             void FrameEnd();
 
         private:
-            void RecreateSwapchain(Vector2 dimensions);
+            void Resize(const Vector2& dimensions);
 
             camera::MainCamera* m_mainCamera;
+            Registry* m_registry;
```
McCallisterRomer, line 47: Registry doesn't need to be kept as a member, just passed in the constructor.

source/engine/graphics/RenderGraph.h
```diff
@@ -0,0 +1,75 @@
...
+
+
+    private:
+        vk::Device m_device;
+        Swapchain* m_swapchain;
+        GpuOptions* m_gpuOptions;
+        GpuAllocator* m_gpuAllocator;
+        ShaderDescriptorSets* m_descriptorSets;
+        std::vector<std::unique_ptr<RenderPass>> m_renderPasses{};
+        Vector2& m_dimensions;
```
McCallisterRomer: Reference here is ya bug.

source/engine/graphics/Graphics.cpp
```diff
@@ -59,47 +67,34 @@ namespace nc::graphics
...
-        m_swapchain.reset();
-
-        // Recreate swapchain and resources
-        auto shadowMap = ShadowMap { .dimensions = m_dimensions };
-        m_shaderResources.get()->GetShadowMapShaderResource().Update(std::vector<ShadowMap>{shadowMap});
-        m_swapchain = std::make_unique<Swapchain>(m_core->logicalDevice.get(), m_core->physicalDevice, m_core->surface.get(), m_dimensions);
-        m_commands = std::make_unique<Commands>(m_core->logicalDevice.get(), m_core->physicalDevice, m_core->surface.get(), m_swapchain.get());
-        m_renderer = std::make_unique<Renderer>(m_core->logicalDevice.get(), m_swapchain.get(), m_gpuOptions.get(), m_allocator.get(), m_shaderResources.get(), dimensions);
+        m_swapchain->Resize(dimensions);
+        m_renderGraph.reset();
```
McCallisterRomer: With the reference gone, can remove reset and make_unique, and just call Resize() here.

source/engine/graphics/Imgui.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+#include "ecs/Registry.h"
```
McCallisterRomer, line 3: can remove

source/engine/graphics/Imgui.h
```diff
@@ -0,0 +1,21 @@
...
+#include "graphics/GpuAssetsStorage.h"
```
McCallisterRomer, line 4: can remove this too

source/engine/graphics/Imgui.h
```diff
@@ -0,0 +1,21 @@
...
+
+namespace nc::graphics
+{
+class Imgui
+{
+    public:
+        explicit Imgui(vk::Device device);
+
+        /* @todo: Move contents into constructor and remove this method during fix of imgui bug.*/
+        void InitializeImgui(vk::Instance instance, vk::PhysicalDevice physicalDevice, vk::Device logicalDevice, vk::RenderPass renderPass, Commands *commands, uint32_t maxSamplesCount);
```
McCallisterRomer, line 16: Should forward declare `Commands` here.

source/engine/graphics/Imgui.h
```diff
@@ -0,0 +1,21 @@
...
+#include "graphics/GpuAssetsStorage.h"
+
+#include "vulkan/vk_mem_alloc.hpp"
+
+namespace nc::graphics
+{
+class Imgui
+{
+    public:
+        explicit Imgui(vk::Device device);
```
McCallisterRomer, line 13: nice explicit!

source/engine/graphics/Imgui.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include "graphics/PerFrameRenderState.h"
+
+#include "imgui/imgui_impl_vulkan.h"
+
+namespace
+{
+auto CreateImguiDescriptorPool(vk::Device device) -> vk::UniqueDescriptorPool
+{
+    // Create descriptor pool for IMGUI
+    const std::array<vk::DescriptorPoolSize, 11> imguiPoolSizes =
```
McCallisterRomer, line 11: This maybe con be constexpr.

source/engine/graphics/Lighting.h
```diff
@@ -0,0 +1,53 @@
+#pragma once
+
+#include "ecs/Registry.h"
+#include "renderpasses/RenderPass.h"
+
+#include "vulkan/vk_mem_alloc.hpp"
+
+namespace nc::graphics
```
McCallisterRomer, line 8: Unindent file for style, yo.

source/engine/graphics/Lighting.h
```diff
@@ -0,0 +1,53 @@
...
+{
+    class GpuAllocator;
+    class GpuOptions;
+    class PointLight;
+    class RenderGraph;
+    class ShaderDescriptorSets;
+    struct ShaderResources;
+    class Swapchain;
+
+    inline static const std::string ShadowMappingPassId = "Shadow Mapping Pass";
```
McCallisterRomer: Can be moved to cpp in anon namespace.

source/engine/graphics/Lighting.cpp
```diff
@@ -0,0 +1,111 @@
...
+        AttachmentSlot{0, AttachmentType::ShadowDepth, vk::Format::eD16Unorm, vk::AttachmentLoadOp::eClear, vk::AttachmentStoreOp::eStore, vk::SampleCountFlagBits::e1}};
+
+    std::vector<Subpass> shadowSubpasses{
+        Subpass{shadowAttachmentSlots[0]}};
+
+    std::vector<std::unique_ptr<Attachment>> attachments;
+    attachments.push_back(std::make_unique<Attachment>(device, allocator, dimensions, true, vk::SampleCountFlagBits::e1, vk::Format::eD16Unorm));
+
+    const auto size = AttachmentSize{dimensions, swapchain->GetExtent()};
+    auto renderPass = std::make_unique<RenderPass>(device, 0u, id, shadowAttachmentSlots, shadowSubpasses, std::move(attachments), size, ClearValueFlags::Depth);
```
McCallisterRomer: can move the other two vectors as well

source/engine/graphics/Lighting.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+    auto handles = std::vector<vk::ImageView>{renderPass->attachments[0]->view.get()};
+    RegisterAttachments(device, renderPass.get(), std::move(handles), dimensions, 0);
```
McCallisterRomer: Based on previous conversation, I think FrameBuffer should definitely have a vector instead of a span, otherwise its just looking at this local var.

source/engine/graphics/Lighting.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+    return std::move(renderPass);
```
McCallisterRomer: No need to return by move

source/engine/graphics/renderpasses/RenderPass.h
```diff
@@ -30,43 +32,62 @@ enum class AttachmentType : uint8_t
...
 
 struct RenderPass
 {
-    RenderPass(std::span<const AttachmentSlot> attachmentSlots, std::span<const Subpass> subpasses,
-               vk::Device device, const RenderTargetSize& size, std::string uid, ClearValueFlags_t clearFlags);
-
-    vk::UniqueRenderPass renderpass;
-    std::vector<std::unique_ptr<ITechnique>> techniques;
-    RenderTargetSize renderTargetSize;
+    RenderPass(vk::Device device,
```
McCallisterRomer: I vote class here and add free functions below as member functions.

source/engine/graphics/Lighting.cpp
```diff
@@ -0,0 +1,111 @@
...
+    m_numShadowCasters = 0u;
+    for (auto i = 0u; i < countBeforeClearing; ++i)
+    {
+        OnAddPointLightConnection();
+    }
+}
+
+void Lighting::OnAddPointLightConnection()
+{
+    const auto id = ShadowMappingPassId + std::to_string(m_numShadowCasters);
```
McCallisterRomer: Per our conversation, store ids in a vector and update Clear, Resize, OnAdd, and OnRemove.

source/engine/graphics/RenderGraph.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+    std::vector<Subpass> litSubpasses{
+        Subpass{litAttachmentSlots[0], litAttachmentSlots[1], litAttachmentSlots[2]}};
+
+    std::vector<std::unique_ptr<Attachment>> attachments;
+    attachments.push_back(std::make_unique<Attachment>(device, allocator, dimensions, true, numSamples, gpuOptions->GetDepthFormat())); // Depth Stencil
+    attachments.push_back(std::make_unique<Attachment>(device, allocator, dimensions, false, numSamples, swapchain->GetFormat())); // Color Buffer
+
+    const auto size = AttachmentSize{dimensions, swapchain->GetExtent()};
+    auto renderPass = std::make_unique<RenderPass>(device, 1u, LitPassId, litAttachmentSlots, litSubpasses, std::move(attachments), size, ClearValueFlags::Depth | ClearValueFlags::Color);
```
McCallisterRomer: can move other two vectors

source/engine/graphics/RenderGraph.cpp
```diff
@@ -0,0 +1,211 @@
...
+    {
+        std::vector<vk::ImageView> imageViews
+        {
+            colorResolveView.get(), // Color Resolve View
+            depthImageView.get(), // Depth View
+            imageView.get()
+        };
+        RegisterAttachments(device, renderPass.get(), imageViews, dimensions, index++);
+    }
+    return std::move(renderPass);
```
McCallisterRomer: no need to return by move

source/engine/graphics/RenderGraph.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+    RegisterTechnique<EnvironmentTechnique>(LitPassId);
+    RegisterTechnique<PbrTechnique>(LitPassId);
+    RegisterTechnique<ParticleTechnique>(LitPassId);
+    RegisterTechnique<UiTechnique>(LitPassId);
+}
+
+RenderGraph::~RenderGraph() noexcept
+{
+    m_renderPasses.clear();
```
McCallisterRomer: default destructor will do this

source/engine/graphics/RenderGraph.cpp
```diff
@@ -0,0 +1,211 @@
...
+    auto* renderPass = Acquire(uid);
+    renderPass->techniques.push_back(std::make_unique<ShadowMappingTechnique>(m_device, m_gpuOptions, m_descriptorSets, renderPass->renderPass.get(), shadowCasterIndex));
+}
+
+void RenderGraph::Resize(Swapchain* swapchain, const Vector2& dimensions)
+{
+    m_renderPasses.clear();
+    m_dimensions = dimensions;
+    m_swapchain = swapchain;
+    Add(CreateLitPass(m_device, m_gpuAllocator, m_gpuOptions, m_swapchain, m_dimensions));
```
McCallisterRomer: Since the stuff here and below is also done in the c'tor, I wonder if it should be factored out.

McCallisterRomer (approved):

PR #342, opened 2023-02-26
First pass graphics cleanup

McCallisterRomer:
Initial work for graphics cleanup. I've added a [Graphics Refactoring](https://github.com/NcStudios/NcEngine/milestone/10) milestone to track future work.

- Split up `Core` into `Instance`, `Device`, and a helper for physical device selection.
- `QueueFamilyIndices` is now stored in `Device` to reduce how many times it gets created.
  - Also added/refactored some member functions here.
- Removed Commands.h/cpp. Execute and queues are now part of `Device`.
- Moved `MainCamera`-related code from `Graphics` to `NcGraphicsImpl`.
- Added some parameters to the `Graphics` constructor for vulkan api version and such. Its constructor is pretty ugly at the moment, but this will continue to get cleaned up in the next few PRs.


source/engine/engine/NcEngineImpl.cpp
```diff
@@ -15,14 +15,24 @@ namespace
...
                          nc::EngineInitFlags flags) -> nc::ModuleRegistry
 {
     NC_LOG_INFO("Building module registry");
     const bool enableGraphics = nc::EngineInitFlags::None == (flags & nc::EngineInitFlags::NoGraphics);
     const bool enablePhysics = nc::EngineInitFlags::None == (flags & nc::EngineInitFlags::NoPhysics);
     const bool enableAudio = nc::EngineInitFlags::None == (flags & nc::EngineInitFlags::NoAudio);
     auto moduleRegistry = nc::ModuleRegistry{};
-    moduleRegistry.Register(nc::graphics::BuildGraphicsModule(enableGraphics, reg, gpuAccessorSignals, window));
+
+    auto graphicsInitInfo = nc::graphics::GraphicsInitInfo{
```
jaremieromer: Curly on the next line, IMO

source/engine/graphics/CMakeLists.txt
```diff
@@ -22,6 +20,7 @@ target_sources(${NC_ENGINE_LIB}
 
 add_subdirectory(buffers)
 add_subdirectory(components)
+add_subdirectory(core)
```
jaremieromer, line 23: I like the move to this being a folder

source/engine/graphics/GpuAllocator.cpp
```diff
@@ -16,10 +18,10 @@ namespace
 
 namespace nc::graphics
 {
-    GpuAllocator::GpuAllocator(vk::Device logicalDevice, vk::PhysicalDevice physicalDevice, vk::Instance instance, Commands* commands)
-        : m_commands{commands},
-          m_allocator{CreateAllocator(logicalDevice, physicalDevice, instance)},
-          m_deviceProperties{physicalDevice.getProperties()}
+    GpuAllocator::GpuAllocator(const Device* device, const Instance& instance)
+        : m_device{device},
+          m_allocator{::CreateAllocator(device->VkDevice(), device->VkPhysicalDevice(), instance.VkInstance())},
```
jaremieromer, line 23: I've started naming my vulkan types VkXX as well, I like this

source/engine/graphics/GpuAllocator.cpp
```diff
@@ -62,7 +64,7 @@ namespace nc::graphics
 
     void GpuAllocator::CopyBuffer(const vk::Buffer& sourceBuffer, const vk::Buffer& destinationBuffer, const vk::DeviceSize size)
     {
-        m_commands->ExecuteCommand([sourceBuffer, destinationBuffer, size](vk::CommandBuffer cmd)
+        m_device->ExecuteCommand([sourceBuffer, destinationBuffer, size](vk::CommandBuffer cmd)
```
jaremieromer, line 67: NICE! I love that we're moving more towards the DX11 simplicity of having the 'device' be the center of these calls

source/engine/graphics/GpuAllocator.h
```diff
@@ -48,7 +49,7 @@ namespace graphics
     class GpuAllocator
     {
         public:
-            GpuAllocator( vk::Device logicalDevice, vk::PhysicalDevice physicalDevice, vk::Instance instance, Commands* commands);
+            GpuAllocator(const Device* device, const Instance& instance);
```
jaremieromer, line 52: Why const ptr here?

source/engine/graphics/Graphics.cpp
```diff
@@ -1,42 +1,48 @@
...
+#include "Lighting.h"
+#include "RenderGraph.h"
+#include "Swapchain.h"
+#include "core/Device.h"
+#include "core/Instance.h"
+#include "shaders/ShaderDescriptorSets.h"
+#include "shaders/ShaderResources.h"
 #include "config/Config.h"
 #include "ecs/Registry.h"
 #include "graphics/Camera.h"
```
jaremieromer, line 16: missed this one (graphics/Camera)

source/engine/graphics/core/Device.cpp
```diff
@@ -0,0 +1,108 @@
...
+}
+} // anonymous namespace
+
+namespace nc::graphics
+{
+auto Device::Create(const Instance& instance, vk::SurfaceKHR surface,
+                    std::span<const char* const> requiredExtensions) -> std::unique_ptr<Device>
+{
+    const auto [physicalDevice, queueIndices] = detail::SelectPhysicalDevice(
+        instance.GetPhysicalDevices(), surface, requiredExtensions
```
jaremieromer, line 60: This is SO readable!

source/engine/graphics/NcGraphicsImpl.cpp
```diff
@@ -37,28 +37,43 @@ namespace
...
     }
 
-    NcGraphicsImpl::NcGraphicsImpl(Registry* registry, const nc::GpuAccessorSignals& gpuAccessorSignals, window::WindowImpl* window)
+    NcGraphicsImpl::NcGraphicsImpl(Registry* registry, GraphicsInitInfo info, window::WindowImpl* window)
         : m_registry{ registry },
-          m_camera{},
-          m_graphics{ &m_camera,
-                      registry,
-                      gpuAccessorSignals,
+          // TODO #341: Instead of constructing here, pass in from BuildGraphicsModule
```
jaremieromer, line 70: I like referencing the issue #, but would prefer we link it (either by URL or otherwise)

source/engine/graphics/core/Instance.cpp
```diff
@@ -0,0 +1,110 @@
...
+        return false;
+    }
+
+    auto availableLayers = std::vector<vk::LayerProperties>(layerCount);
+    if (vk::enumerateInstanceLayerProperties(&layerCount, availableLayers.data()) != vk::Result::eSuccess)
+    {
+        return false;
+    }
+
+    return std::ranges::all_of(g_validationLayers, [&availableLayers](const auto& requiredLayer)
```
jaremieromer, line 27: nice

source/engine/graphics/core/Instance.h
```diff
@@ -0,0 +1,36 @@
...
+        Instance& operator=(const Instance&) = delete;
+        Instance& operator=(Instance&&) = delete;
+
+        auto VkInstance() const noexcept -> vk::Instance
+        {
+            return m_instance.get();
+        }
+
+        auto CreateSurface(HWND hwnd, HINSTANCE hinstance) const -> vk::UniqueSurfaceKHR;
+
```
jaremieromer: Is there intentionally an extra newline here?

jaremieromer (approved):
It's getting better and better!

source/engine/graphics/Graphics.cpp
McCallisterRomer: That one is actually not a sibling file. Its include/graphics/Camera.h.

source/engine/graphics/GpuAllocator.h
McCallisterRomer: As opposed to non-const pointer, or const reference? Its const because it can be; its a pointer because `GpuAllocator` keeps it as a member. Our current convention is that we don't store references as members. Mostly that's because it gets in the way of move operations, which doesn't really apply to this non-moveable type, but I figured why break the convention?

source/engine/graphics/core/Instance.h
McCallisterRomer: Nope, fixed.

source/engine/engine/NcEngineImpl.cpp
McCallisterRomer: Fixed

Issue #341, opened 2023-02-20
Make NcGraphics API agnostic


McCallisterRomer:
`NcGraphicsImpl.h` still depends on Vulkan. To factor this out:
- Create an interface for graphics apis (`IGraphicsApi` or something) based on current `Graphics` functions. This might need to be tweaked in the future, but this should be a fine starting point.
- Change `Graphics` to `VulkanGraphicsApi` which implements `IGraphicsApi`.
- `BuildGraphicsModule()` constructs a graphics implementation based on the api specified by its inputs and passes it to `NcGraphicsImpl`.
- The `m_graphics` member of `NcGraphicsImpl` should then be a `std::unique_ptr<IGraphicsApi>`.
- Also, since most things under graphics/ are Vulkan specific, we'll have to start moving stuff to implementation folders. This is probably best done as separate work, as it'll be a ton of moved files.

McCallisterRomer:
There is still Vulkan required in NcGraphics.cpp for `PerFrameRenderState`. `ObjectData` will need to get factored out of its current location.

PR #351, opened 2023-02-27
Make NcGraphics API-agnostic

McCallisterRomer:
Part of #341

- Removing (some) Vulkan from `NcGraphics`:
  - Adding `IGraphics` interface.
  - Updated `NcGraphics` to take a pointer to `IGraphics`.
  - Changed `Graphics` to `VulkanGraphics`.
  - The header should be clear of vulkan stuff now (but not Windows). The cpp still requires it for `PerFrameRenderState`. (maybe elsewhere too).  The `ObjectData` struct will need to get pulled out into its own file.
- Updated each module to take its in corresponding config struct.
  - Removed `EngineInitFlags` - moved bools into config.
  - Added settings for audio that just has a bool for now.


source/engine/graphics/IGraphics.h
```diff
@@ -0,0 +1,46 @@
...
+namespace window
+{
+class WindowImpl;
+}
+
+namespace graphics
+{
+struct PerFrameRenderState;
+
+class IGraphics
```
jaremieromer, line 27: This is SENSATIONAL!

source/engine/graphics/IGraphics.h
```diff
@@ -0,0 +1,46 @@
...
+
+        virtual void InitializeUI() = 0;
+        virtual auto FrameBegin() -> bool = 0;
+        virtual void Draw(const PerFrameRenderState& state) = 0;
+        virtual void FrameEnd() = 0;
+        virtual void OnResize(float width, float height, const WPARAM windowArg) = 0;
+        virtual void Clear() noexcept = 0;
+};
+
+auto GraphicsFactory(const config::ProjectSettings& projectSettings,
```
jaremieromer, line 40: Wow - pared down like this, its actually easy to follow what this interface requires. 

I LOVE IT.

It makes me want to rename `GpuAccessorSignals` to something that makes as much sense as graphicsSettings, projectSettings, window and registry. Those four items make sense to anybody without needing context for our implementation - but the signals do not. 

Maybe we rename the `GpuAccessorSignals` class to something like `AssetLoaders` or something similar.  Because assetLoaders makes sense as a requirement to a graphics class (much more than signals/`GpuAccessorSignals`)

source/engine/graphics/NcGraphicsImpl.cpp
```diff
@@ -1,7 +1,9 @@
 #include "NcGraphicsImpl.h"
+#include "PerFrameRenderState.h"
```
jaremieromer, line 2: What's going on with this alphabetizing scheme? shouldn't 'PerFrameRenderState.h' be after 'ecs/View.h'? It's not clear to readers that it is in the graphics namespace imo

source/engine/graphics/NcGraphicsImpl.h
```diff
@@ -1,45 +1,43 @@
 #pragma once
 
+#include "IGraphics.h"
 #include "camera/MainCamera.h"
 #include "ecs/PointLightSystem.h"
 #include "ecs/ParticleEmitterSystem.h"
-#include "Graphics.h"
 #include "graphics/NcGraphics.h"
```
jaremieromer, line 7: if we're going with not including the folder if it's our current folder (ie if 'graphics', then no need for "graphics/" in include statements, then we should delete "graphics" in "graphics/NcGraphics.h"

source/engine/graphics/api/vulkan/VulkanGraphics.cpp
```diff
@@ -0,0 +1,154 @@
+#include "VulkanGraphics.h"
+#include "graphics/FrameManager.h"
```
jaremieromer, line 2: We should align on whether or not we are including this in include statements.
"This" being the base folder name -- so "graphics" in this case.

My 2¢ - I like to include them. So, regardless of which folder the file is in, we fully specify from "source". AKA: We keep "graphics/"

I bring this up because I know you've deleted them elsewhere. Let's discuss/pick one

source/engine/graphics/api/vulkan/VulkanGraphics.cpp
```diff
@@ -0,0 +1,154 @@
...
+#include "optick/optick.h"
+
+namespace
+{
+constexpr auto g_requiredDeviceExtensions = std::array<const char*, 1>{VK_KHR_SWAPCHAIN_EXTENSION_NAME};
+} // anonymous namespace
+
+namespace nc::graphics::vulkan
+{
+VulkanGraphics::VulkanGraphics(const config::ProjectSettings& projectSettings,
```
jaremieromer, line 30: So beautiful

source/engine/audio/NcAudioImpl.cpp
```diff
@@ -38,18 +39,16 @@ struct NcAudioStub : public nc::audio::NcAudio
 
 namespace nc::audio
 {
-auto BuildAudioModule(bool enableModule, Registry* reg) -> std::unique_ptr<NcAudio>
+auto BuildAudioModule(const config::AudioSettings& settings, Registry* reg) -> std::unique_ptr<NcAudio>
 {
-    if(enableModule)
+    if(settings.enabled)
     {
         NC_LOG_TRACE("Creating NcAudio module");
```
jaremieromer: This is a very minor nit but - will you change "Creating" to "Building" in the trace so that it aligns with the method name? Here and for the other NcXImpls

jaremieromer (commented):
This is so nice that I literally want to cry.

Awesome job - it's really inspiring me to go further with the cleanup

jaremieromer (approved):

source/engine/graphics/IGraphics.h
McCallisterRomer: I made issues for adding an asset module to replace these, as we discussed.

source/engine/graphics/NcGraphicsImpl.cpp
McCallisterRomer: Grouped by: 
1. The thing you're implementing, if applicable
2. Siblings
3. Non-siblings, still within the repo
4. Whitespace
5. Stuff from other repos
6. Whitespace
7. System includes

Each group is alphabetized.

source/engine/graphics/NcGraphicsImpl.h
McCallisterRomer: NcGraphics.h is not a sibling to this file (its at include/graphics/NcGraphics.h), so this is necessary.

source/engine/graphics/api/vulkan/VulkanGraphics.cpp
McCallisterRomer: Kind of the same as the previous one. We aren't directly under /graphics here, so we have to either ../../ up to it or use /graphics.

source/engine/audio/NcAudioImpl.cpp
McCallisterRomer: changed

PR #359, opened 2023-03-05
Make NcGraphics API-agnostic pt 2

McCallisterRomer:
Remaining vulkan references in NcGraphics were due to `ObjectData` being in `PerFrameRenderState`. Some of these shader data types are not vulkan-specific, and need to be known outside of vulkan code. `ObjectData` and `EnvironmentData` have been pulled out of the vulkan-specific resource managers, and all vulkan code was moved into graphics/api/vulkan.

Most changes are moved files and related include fixes. Other changes:
- `GpuOptions` was moved from `VulkanGraphics` into `Device`. No more pointers to it are passed around. I also removed the unused physical device and device properties members from it.
-  Some cases where we thought data needed to be persisted after a vulkan call (when making a RenderPass or FrameBuffer) have been updated. I bet there is more of this to find. For now `FrameBuffer` and `RenderPass` have to been updated to be a slightly more efficient.


jaremieromer (approved):
Woohoo!

Issue #357, opened 2023-03-04
Split PointLightInfo into API/Shader types


McCallisterRomer:
The same `PointLightInfo` struct is currently used for both `PointLight` initialization and for shader binding. There should be separate types for this, so that implementation details (like alignment, padding, projection matrix, position, etc.), are not publicly exposed.

PR #362, opened 2023-03-11
PointLightInfo cleanup

McCallisterRomer:
#357 

Removing implementation details from `PointLight.h`

- Renamed `PointLightInfo` to `PointLightData` and moved it to the source directory. Changed `PointLight`'s constructor accordingly.
- Removed all of the dirty state tracking from `PointLight` and `PointLightSystem` to simplify things. We don't expect the buffer to be large anyway.
- Moved the remaining logic from `PointLight` and related stuff in `PerFrameRenderState` into `PointLightSystem`. I have a plan to carry on with this, eventually removing all logic from `PerFrameRenderState`.


source/engine/graphics/PointLightSystem.cpp
jaremieromer, line 1: This is great

source/engine/graphics/api/vulkan/techniques/ShadowMappingTechnique.cpp
```diff
@@ -124,7 +125,8 @@ namespace nc::graphics
         auto pushConstants = ShadowMappingPushConstants{};
 
         // We are rendering the position of each mesh renderer's vertex in respect to each point light's view space.
-        pushConstants.lightViewProjection = frameData.pointLightVPs.at(m_shadowCasterIndex);
+        NC_ASSERT(m_shadowCasterIndex < frameData.pointLightVPs.size(), "Shadow caster index is out of bounds.");
```
jaremieromer, line 128: nice

jaremieromer (approved):
Woo!

Issue #361, opened 2023-03-11
Restructure PerFrameRenderState and Gfx Data Transmission


McCallisterRomer:
Goals:
- Split the logic from `PerFrameRenderState`'s constructor into separate classes and remove `MapPerFrameRenderState()`.
- Remove usage of `ShaderResourceService::Get()` for to `ObjectData`, `PointLightData`, and `EnvironmentData`. We'd like to either use signals or be able to get some interface types from `IGraphics`. The texture/cubemap versions will require additional planning, so leave them be for now.

Rough thoughts:
- `NcGraphicsImpl::Run()` calls something like `GraphicsDataChannel::Broadcast()`
  - Sends data (through signals?) to be picked up by API-specific handlers
    - `ObjectDataBroadcaster` emits `ObjectData[]`
    - `PointLightDataBroadcaster` emits `PointLightData[]`
    - `EnvironmentDataBroadcaster` emits `EnvironmentData`
  - returns `PerFrameRenderState`? This seems reasonable, but maybe don't use the name 'Broadcast' then.

The `Collider` stuff should probably be moved somewhere totally different.

PR #364, opened 2023-03-15
Split up PerFrameRenderState

McCallisterRomer:
#361 

Moving logic from `PerFrameRenderState` into distinct systems for obejcts, widgets, lights, camera... (action?). Some of them are new systems, some are reworks of existing things.

Each system produces its own state object, which may get passed into the next system, but they all ultimately end up as an aggregate `PerFrameRenderState`.

Also using this opportunity to remove some usage of `ShaderResourceService::Get()`. Any of these systems that need to supply data down to the depths of `VulkanGraphics` do so through a signal. A 'bus' object gets passed to the factory for `IGraphics` so that vulkan stuff can subscribe to the required events. Each signal in the bus gets sunk into the appropriate system so it can publish the shader resource data during its `Execute()` operation.


source/engine/graphics/PerFrameRenderState.h
```diff
@@ -1,56 +1,23 @@
...
-}
-#endif
-
-namespace graphics
+namespace nc::graphics
 {
-class Camera;
-class Environment;
-
 struct PerFrameRenderState
```
jaremieromer, line 14: This literally makes my heart SING

source/engine/graphics/api/vulkan/shaders/ShaderResources.cpp
```diff
@@ -0,0 +1,47 @@
+#include "ShaderResources.h"
+#include "graphics/shader_resource/EnvironmentData.h"
+#include "graphics/shader_resource/ShaderResourceBus.h"
+
+namespace
```
jaremieromer, line 5: I like how uniform this is becoming

source/engine/graphics/api/vulkan/techniques/EnvironmentTechnique.cpp
```diff
@@ -92,7 +92,7 @@ EnvironmentTechnique::~EnvironmentTechnique() noexcept
 
 bool EnvironmentTechnique::CanBind(const PerFrameRenderState& frameData)
 {
-    return frameData.useSkybox;
+    return frameData.environmentState.useSkybox;
```
jaremieromer, line 95: I love how everything in frameData is an xState - it makes usage a lot more clear

source/engine/graphics/NcGraphicsImpl.cpp
```diff
@@ -129,43 +137,45 @@ namespace nc::graphics
...
-            return;
-
         m_ui.FrameBegin();
 
         #ifdef NC_EDITOR_ENABLED
         /** @todo I think the old editor can start to go away. I'm
          *  hacking dt factor in here. */
         float dtFactor = 1.0f;
         m_ui.Draw(&dtFactor, m_registry);
+        auto widgetState = m_widgetSystem.Execute(View<physics::Collider>{m_registry});
```
jaremieromer, line 153: This just made everything so much clearer

jaremieromer (approved):
This is the best thing ever!

Issue #354, opened 2023-03-03
Pass asset module into NcGraphics


McCallisterRomer:
The goal here is to make graphics depend only on the asset interface (only include stuff from /include), rather than digging through its implementation. Update `NcGraphics` to take in a pointer to the asset module interface in its c'tor, instead of `GpuAccessorSignals`. 

Obliterate `GpuAccessorSignals`.

Issue #352, opened 2023-03-03
Create basic asset module


McCallisterRomer:
The interface only needs to include functions to get event signals. Moving the existing load/unload functions should be handled in a separate task.

The implementation should replace `AssetManagers` (the individual managers will stick around, but will be housed within the module impl).

Types in AssetData.h will need to be moved to /include.

PR #366, opened 2023-03-16
Add Asset Module

McCallisterRomer:
#352 #354

Creating the beginning of the asset module. For now, this only includes event signals; loading functions are unchanged. This allows us to remove `GpuAccessorSignals` and instead pass a pointer to `NcAsset` when building `NcGraphics` -  `NcGraphics` only knows about the public-facing `NcAsset` and supporting types, rather than relying on internal stuff.

The majority of the changes here are just name/namespace updates. I had to move AssetData.h to /include to make this work, and figured I should move those types to the `asset` namespace and try to make their names a bit more descriptive.


include/ncengine/asset/AssetData.h
```diff
@@ -0,0 +1,61 @@
...
+
+/** @brief Event data for mesh load and unload operations. */
+struct MeshUpdateEventData
+{
+    std::span<const asset::MeshVertex> vertices;
+    std::span<const uint32_t> indices;
+};
+
+/** @brief A data/id pair for a loaded texture. */
+struct TaggedTexture
```
jaremieromer: For this and `TaggedCubeMap`, I don't like that the name includes "Tag" but the member is actually `id`. Can we rename to TextureWithId?

source/engine/graphics/IGraphics.cpp
```diff
@@ -9,7 +9,7 @@ namespace nc::graphics
 {
 auto GraphicsFactory(const config::ProjectSettings& projectSettings,
                      const config::GraphicsSettings& graphicsSettings,
-                     const GpuAccessorSignals& signals,
+                     asset::NcAsset* assetModule,
```
jaremieromer, line 12: Sweet

jaremieromer (approved):

PR #379, opened 2023-03-25
Replace Win32 window and input with GLFW

jaremieromer:
### Goal
Further the cross platform initiative by cutting over window and input from Win32 to GLFW.

### Changes
Added GLFW library.
Replaced Win32 input and window with GLFW input and window.


source/engine/graphics/api/vulkan/VulkanGraphics.h
```diff
@@ -72,7 +74,7 @@ class VulkanGraphics : public IGraphics
         Vector2 m_dimensions;
         bool m_isMinimized;
 };
-} // namespace vulkan
+} // namespace nc::graphics::vulkan
```
McCallisterRomer: This and the one below it are wrong. Should be vulkan, graphics, nc each on their own line.

source/engine/graphics/api/vulkan/VulkanGraphics.cpp
```diff
@@ -80,10 +82,10 @@ void VulkanGraphics::Resize(const Vector2& dimensions)
     m_lighting->Resize(dimensions);
 }
 
-void VulkanGraphics::OnResize(float width, float height, const WPARAM windowArg)
+void VulkanGraphics::OnResize(float width, float height, bool isMinimized)
```
McCallisterRomer, line 85: It probably should have always been like this.

source/engine/graphics/api/vulkan/core/Instance.cpp
```diff
@@ -4,10 +4,10 @@
 
 #include <algorithm>
 #include <array>
+#include "GLFW/glfw3.h"
```
McCallisterRomer: This should go above, grouped with "ncutility/NcError.h"

source/engine/graphics/api/vulkan/core/Instance.cpp
```diff
@@ -85,10 +88,11 @@ Instance::Instance(std::string_view appName, uint32_t appVersion,
 {
 }
 
-auto Instance::CreateSurface(HWND hwnd, HINSTANCE hinstance) const -> vk::UniqueSurfaceKHR
+auto Instance::CreateSurface(GLFWwindow* window) const -> vk::UniqueSurfaceKHR
 {
-    auto createInfo = vk::Win32SurfaceCreateInfoKHR{{}, hinstance, hwnd};
-    return m_instance->createWin32SurfaceKHRUnique(createInfo);
+    VkSurfaceKHR surface;
+    glfwCreateWindowSurface(m_instance.get(), window, nullptr, &surface);
```
McCallisterRomer: Does this return an error code we need to check?

source/engine/input/Input.cpp
```diff
@@ -4,16 +4,43 @@
 #include <algorithm>
 #include <vector>
 #include <windowsx.h>
+#include "GLFW/glfw3.h"
```
McCallisterRomer, line 4: Should be before sys headers. Also, do we need <windowsx.h> now?

source/engine/input/Input.cpp
```diff
@@ -4,16 +4,43 @@
...
+
     struct
     {
-        std::vector<InputItem> downKeys = {};
-        std::vector<InputItem> upKeys = {};
+        std::unordered_map<KeyCode, KeyState> keyStates = {};
         uint32_t mouseX = 0u;
         uint32_t mouseY = 0u;
         int32_t mouseWheel = 0;
+        GLFWwindow* window = nullptr;
```
McCallisterRomer: Is this needed for something? I only see it used in SetWindow() below.

source/engine/input/Input.cpp
```diff
@@ -4,16 +4,43 @@
...
+    Released
+};
+
+auto ToKeyState(int action)
+{
+    switch(action)
+    {
+        case GLFW_PRESS: return KeyState::Pressed;
+        case GLFW_RELEASE: return KeyState::Released;
+        case GLFW_REPEAT: return KeyState::Held; // Note: 'Repeat' is not strictly the same as 'held', and we handle checking for and setting the 'held' state in Input::Flush. However, we know that if 'repeat' is true, 'held' is also true.
```
McCallisterRomer, line 26: I wonder if we want to ignore this type.

source/engine/input/Input.cpp
```diff
@@ -44,75 +71,60 @@ namespace nc::input
...
     {
-        g_state.downKeys.emplace_back(keyCode, lParam);
+        g_state.keyStates[static_cast<KeyCode>(keyCode)] = ::ToKeyState(action);
     }
 
-    void AddMouseButtonUpToQueue(KeyCode_t keyCode, LPARAM lParam)
+    void Flush()
     {
-        g_state.upKeys.emplace_back(keyCode, lParam);
+        for (auto cur = g_state.keyStates.begin(); cur != g_state.keyStates.end();)
```
McCallisterRomer, line 102: Is this still behaving strangely? I don't understand how this works as is. How is a released event ever removed?

source/engine/window/WindowImpl.cpp
```diff
@@ -38,32 +36,21 @@ namespace nc::window
...
     {
+        glfwInit();
+        glfwWindowHint(GLFW_CLIENT_API, GLFW_NO_API);
+
         g_instance = this;
-        GetModuleHandleExA(0, NULL, &m_hInstance);
 
         const auto& projectSettings = config::GetProjectSettings();
         const auto& graphicsSettings = config::GetGraphicsSettings();
+        const auto* videoMode = glfwGetVideoMode(glfwGetPrimaryMonitor());
```
McCallisterRomer, line 50: Does this need NULL check on error?

source/engine/window/WindowImpl.cpp
```diff
@@ -74,76 +61,49 @@ namespace nc::window
...
-            (LONG)(top + m_dimensions.y)
-        };
+        auto width = Clamp((int)m_dimensions.x, 0, nativeWidth);
+        auto height = Clamp((int)m_dimensions.y, 0, nativeHeight);
 
-        if(!AdjustWindowRect(&clientRect, WndStyleFlags, FALSE))
-        {
-            throw NcError("Failed to adjust client rect to window rect");
-        }
+        m_window = glfwCreateWindow(width, height, projectSettings.projectName.c_str(), nullptr, nullptr);
```
McCallisterRomer, line 71: NULL check?

McCallisterRomer: Oh there is on below SetWindow()

source/engine/window/WindowImpl.cpp
```diff
@@ -161,123 +121,78 @@ namespace nc::window
...
-        return 0;
+
+        glfwPollEvents();
     }
 
-    void WindowImpl::ProcessSystemMessages()
+    void WindowImpl::ProcessKeyEvent(GLFWwindow*, int key, int, int action, int mods)
+    {
+        nc::input::KeyCode_t keyCode = static_cast<nc::input::KeyCode_t>(key);
+        if (mods == GLFW_MOD_SHIFT) keyCode = static_cast<nc::input::KeyCode_t>(nc::input::KeyCode::Shift);
```
McCallisterRomer: Does checking the modifiers like this lose information about other keys? Like shift + B only tells us about shift?

McCallisterRomer: Also would we never know about shift + ctrl at the same time?

source/engine/window/WindowImpl.cpp
```diff
@@ -161,123 +121,78 @@ namespace nc::window
...
+        nc::input::UpdateMousePosition(static_cast<int>(xPos), static_cast<int>(yPos));
+    }
+
+    void WindowImpl::ProcessMouseButtonEvent(GLFWwindow*, int button, int action, int)
     {
         using namespace nc::input;
 
-        MSG message;
-        while(PeekMessage(&message, 0, 0, 0, PM_REMOVE))
+        // TODO: could have more than 3 mouse buttons
```
McCallisterRomer: How many mouse buttons do you have? I have a million macro buttons, not sure how glfw would receive them. It might be worth creating an issue to add the remaining mouse buttons, considering we just throw on line 184 if we get a higher value.

source/engine/window/WindowImpl.h
```diff
@@ -23,30 +22,29 @@ namespace nc::window
             WindowImpl& operator=(const WindowImpl& other) = delete;
             WindowImpl& operator=(WindowImpl&& other) = delete;
 
-            auto GetHWND() const noexcept -> HWND;
-            auto GetHINSTANCE() const noexcept -> HINSTANCE;
+            auto GetWindow() -> GLFWwindow*;
             auto GetDimensions() const noexcept -> Vector2;
+            void SetDimensions(int width, int height) noexcept;
```
McCallisterRomer: Since there are invariants involved with setting dimensions that not handled by this function, I would argue this either shouldn't be a public or it should handle the entire resize process.

source/external/imgui/src/CMakeLists.txt
```diff
@@ -14,13 +14,27 @@ target_compile_definitions(imgui
...
 
+set(NC_INCLUDE_DIR "${PROJECT_SOURCE_DIR}/include")
+set(NC_SOURCE_DIR "${PROJECT_SOURCE_DIR}/source/engine/")
+set(NC_EXTERNAL_DIR "${PROJECT_SOURCE_DIR}/source/external")
+
+# Include Dirs
 target_include_directories(imgui
+    PUBLIC
+        $<INSTALL_INTERFACE:include/>
+        $<INSTALL_INTERFACE:external/>
```
McCallisterRomer: Don't believe INSTALL_INTERFACE has any concept of an external directory.

source/external/imgui/src/CMakeLists.txt
```diff
@@ -14,13 +14,27 @@ target_compile_definitions(imgui
...
+        $<BUILD_INTERFACE:${NC_INCLUDE_DIR}>
```
McCallisterRomer: Is this one needed? I feel like the imgui lib should not need to know about the engine includes, although it may have before for the filtered win32 header.

source/external/imgui/src/CMakeLists.txt
```diff
@@ -14,13 +14,27 @@ target_compile_definitions(imgui
...
+        $<BUILD_INTERFACE:${NC_EXTERNAL_DIR}>
     PRIVATE
         ${PROJECT_SOURCE_DIR}/source/external/
         ${PROJECT_SOURCE_DIR}/include
+        $<BUILD_INTERFACE:${NC_INCLUDE_DIR}/ncengine>
```
McCallisterRomer: Where are we pathing from these two?

source/engine/window/WindowImpl.cpp
jaremieromer: Ha - we straight up crash if we press another mouse button. I will definitely fix!

McCallisterRomer (approved):

PR #381, opened 2023-03-28
NcAudio Updates

### LINKED ISSUE #355: Improve audio device selection
`NcAudio` should have functions for enumerating available audio devices (id and friendly string name) and for setting the active device. ~~The initial device should be selected based on a value in the config.~~ (move to separate issue).

### LINKED ISSUE #356: Make audio device handling more robust
The following cases must be tolerated:
- No available audio devices. (how do we test this?)
- Invalid device id passed when selecting audio device (fallback to default device).
- ~~Audio device becomes unavailable after a stream has been opened (select a different device).~~ (moved to separate issue)

### LINKED ISSUE #367: NcAudio time/ticks property
Add a mechanism for keeping things synced with the audio system. A possibility is to track number of samples written, and expose this through `NcAudio`, perhaps converted to seconds.

McCallisterRomer:
resolves #355
resolves #356
resolves #367

- Updated RtAudio
- Added functions for stream time and device selection to `NcAudio`.
- Added new `DeviceStream` class to manage the RtAudio stream and devices.
- Started to work out error handling. With these changes, failures don't crash us anymore, but we aren't able to recover and create a new stream.


CMakeLists.txt
```diff
@@ -36,6 +36,9 @@ endif()
 # TODO: Taskflow has deprecation warnings for std::aligned_XXX with c++23. We eventually want to remove this.
 add_definitions(-D_SILENCE_ALL_CXX23_DEPRECATION_WARNINGS)
 
+# TODO: RtAudio uses deprecated wchar_t conversion
```
jaremieromer: Will you add an issue for this #TODO?

include/ncengine/audio/NcAudio.h
```diff
@@ -15,14 +31,59 @@ struct NcAudio : public Module
...
      * should not be destroyed while it is registered, except on a scene change.
      * A new Entity or Entity::Null() may be passed to unregister the current
      * listener.
-     * 
-     * @param entity An Entity to use as the listener.
      */
     virtual void RegisterListener(Entity entity) noexcept = 0;
+
+    /**
+     * @brief Get the current stream time.
```
jaremieromer, line 44: At least for me, it isn't readily apparent what stream time means. (ie: could it mean the total length of the clip?) 
Can you update the @brief or @return to say:
Get the number of elapsed seconds since the stream was started?

jaremieromer (approved):

include/ncengine/audio/NcAudio.h
McCallisterRomer: Oh yeah, we should have more detail here.

PR #386, opened 2023-05-24
Add camera properties

McCallisterRomer:
Our handling of clip planes doesn't really work if you want multiple camera's configured differently. Reworking this so that the caller has more control.

- Adding `CameraProperties` struct.
- Updating matrix calculation to use fov.
- Removing near/far z values from resize functions. They aren't needed if the camera manages this.

### DIFF
--- a/include/ncengine/graphics/SceneNavigationCamera.h
+++ b/include/ncengine/graphics/SceneNavigationCamera.h
@@ -60,5 +60,5 @@ class SceneNavigationCamera : public Camera
         auto PanAndTilt(float dt, float speedMult, const Vector3& tiltAxis) -> Quaternion;
         auto Dolly(float dt, float speedMult) -> Vector3;
 };
-} // end namespace graphics
-} // end namespacenc
+} // namespace graphics
+} // namespace nc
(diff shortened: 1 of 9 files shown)


include/ncengine/graphics/SceneNavigationCamera.h
```diff
@@ -59,4 +61,4 @@ class SceneNavigationCamera : public Camera
         auto Dolly(float dt, float speedMult) -> Vector3;
 };
 } // end namespace graphics
-} // end namespacenc
\ No newline at end of file
+} // end namespacenc
```
jaremieromer: this was not introduced in this PR, but: add a space after 'namespace'

source/engine/graphics/NcGraphicsImpl.cpp
```diff
@@ -172,9 +172,9 @@ namespace nc::graphics
         m_graphics->FrameEnd();
     }
 
-    void NcGraphicsImpl::OnResize(float width, float height, float nearZ, float farZ, bool isMinimized)
+    void NcGraphicsImpl::OnResize(float width, float height, bool isMinimized)
```
jaremieromer, line 175: This method makes more sense now

jaremieromer (approved):

PR #390, opened 2023-05-28
Fixed normal map bug in PBR shader

jaremieromer:
### Goal
Fix the bug centered around the TBN matrix in the PBR shader that was causing lighting to misbehave at UV seams. The issue was caused by two things:

 1. Normal maps in Vulkan need to have their image and image views created in `vk::Format::eR8G8B8A8Unorm` format, and previously **all** textures' images and views were created in `vk::Format::eR8G8B8A8Srgb` format.
 2. The normal map sampled image values in the fragment shader needed to be mapped from [-1, 1] instead of from [0,1].

### Changes

 1. To be able to differentiate image textures and normal map textures at asset load time, and without duplicating ALL the `TextureAssetLoader` code, I added an `AssetFlags`struct that contains `uint64_t` flags in `Assets.h` as a parameter to the loader functions. There are three predefined flags in the struct.
 2. Modified `Prefabs.cpp` to use these new flags when loading all the built in image textures and normal maps.
 3. Modified the `Pbr.frag` shader to map the sampled normal map texture from [-1,1] rather than from [0,1]
 4. Deleted a few unused textures and their containing folders.
 
 Before and After
image


include/ncengine/asset/AssetData.h
```diff
@@ -30,9 +30,10 @@ struct TextureWithId
 {
     asset::Texture texture;
     std::string id;
+    bool isNormalMap;
```
McCallisterRomer: Is passing the flags here better for potential changes in the future? I haven't looked at all the code yet, but on first look, it seems easier to pass the whole set of flags to whatever is actually acting on them.

include/ncengine/asset/Assets.h
```diff
@@ -12,53 +12,62 @@
 
 namespace nc
 {
+    /** Asset flags are used to pass parameters to the asset loading/unloading functions. */
+    using asset_flags_type = uint64_t;
+    struct AssetFlags
```
McCallisterRomer, line 17: Just a though for the future - the isExternal bool could become a flag too.

McCallisterRomer (approved):

PR #392, opened 2023-06-02
Fixes for Linux + macOS 

McCallisterRomer:
Fixing some roadblocks found when building for ubuntu & macOS:
- Warnings
- CMake + definition updates
- Updating to later ImGui version
- Removing enet from the build (issue to fork repo and pull out source also covers checking cross-plat support)

Also adding an ubuntu dockerfile (still requires a few manual tweaks to get going, but it can be built).


source/CMakeLists.txt
jaremieromer, line 1: Since we shouldn't have unused code in our codebase, we should remove enet source files as well.

A few justifications:

1. We don't know how long it will be before we address the issue to fork/update a version of enet that will work for us.
2. Once we finally do fork/update, we don't know how close the version we end up with will be to the version we have in source anyway.

source/engine/graphics/components/ToonRenderer.cpp
```diff
@@ -37,13 +37,13 @@ namespace nc::graphics
...
+        ImGui::Text("%s", baseColor.c_str());
         ImGui::Text("Overlay:");
-        ImGui::Text(overlay.c_str());
+        ImGui::Text("%s", overlay.c_str());
         ImGui::Text("Hatching Texture:");
-        ImGui::Text(hatching.c_str());
+        ImGui::Text("%s", hatching.c_str());
         ImGui::Text("Hatching Tiling:");
-        ImGui::Text(std::to_string(hatchingTiling).c_str());
+        ImGui::Text("%u", hatchingTiling);
```
jaremieromer, line 46: nice

jaremieromer (approved):
This is really awesome. 
Do you have a plan to update the docs to describe nuances between platforms? (Currently no audio on linux, etc)

source/CMakeLists.txt
McCallisterRomer: obliterated

McCallisterRomer:
> This is really awesome. Do you have a plan to update the docs to describe nuances between platforms? (Currently no audio on linux, etc)

There are more changes needed for both platforms. I'd rather not mention anything about them in the docs until we can actually test/run on them.

PR #399, opened 2023-06-05
Updated VMA headers to support newer VK versions

### LINKED ISSUE #380: Support for later versions of the Vulkan SDK
Our current VMA HPP bindings are incompatible with the later versions of the Vulkan SDK. We need to update the VMA HPP bindings.

The new repo that maintains the VMA HPP code is: https://github.com/YaaZ/VulkanMemoryAllocator-Hpp

jaremieromer:
### Goal
Updated the VMA headers to support versions of Vulkan newer than 1.3.204.

### Changes

 1. Updated the VMA headers
 2. Corresponding bugfixes in `GpuAllocator.cpp` and `ShaderDescriptorSets.h`
 3. Changed our build.yml file to point to newest Vulkan version in humbletim\setup-vulkan-sdk
 
Closes #380


McCallisterRomer (approved):

McCallisterRomer:
BTW - The Vulkan API we pass in when creating the instance is hardcoded in IGraphics.cpp

PR #402, opened 2023-06-28
Starting to add back the editor

McCallisterRomer:
Adding back a few basic pieces for the editor. It doesn't do anything interesting yet, but this seemed like a good stopping point.

### DIFF
--- a/source/editor/ui/ImGuiUtility.h
+++ b/source/editor/ui/ImGuiUtility.h
@@ -1,0 +1,160 @@
+#pragma once
+
+#include "ncengine/type/EngineTypes.h"
+
+#include "imgui/imgui.h"
+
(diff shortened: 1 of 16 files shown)


source/editor/ui/ImGuiUtility.h
```diff
@@ -0,0 +1,160 @@
+#pragma once
```
McCallisterRomer, line 1: all copy paste from legacy editor

source/editor/ui/UIStyle.cpp
```diff
@@ -0,0 +1,60 @@
+#include "UIStyle.h"
```
McCallisterRomer, line 1: same style/colors as legacy editor

jaremieromer (approved):

PR #407, opened 2023-09-10
MeshRenderer Layout

### LINKED ISSUE #403: Rework MeshRenderer layout
MeshRenderer is pretty large at 248 bytes. By moving the strings/paths behind a 'cold' pointer cuts that down to 56. Make the members looks something like:
```cpp
MeshView mesh;
PbrMaterialView materialView;
TechniqueType techniqueType;
std::unique_ptr<ColdState> cold;
```

McCallisterRomer:
resolves #403

Moving path members in `MeshRenderer` behind a pointer to shrink its memory footprint. The common access pattern just looks at the views, while paths are primarily used by the editor. I also removed some of the #ifdef stuff for simplicity.

Fixed some spacing/separator issues with the built in editor.


include/ncengine/graphics/MeshRenderer.h
```diff
@@ -3,68 +3,87 @@
...
+{
+    NC_ENABLE_IN_EDITOR(MeshRenderer)
+
+    public:
+        MeshRenderer(Entity entity,
+                     std::string meshUid,
+                     PbrMaterial material,
+                     TechniqueType techniqueType = TechniqueType::PhongAndUi);
+
+        auto GetMeshView() const noexcept -> const MeshView&
```
jaremieromer, line 42: Why expand these blocks?

jaremieromer (commented):
Can you make these same changes to ToonRenderer as well?

McCallisterRomer:
> Can you make these same changes to ToonRenderer as well?

Definitely, #410.

include/ncengine/graphics/MeshRenderer.h
McCallisterRomer: Can't say I thought too much about it - I added `noexcept` to them and they felt a little long. I could be talked into changing them back.

jaremieromer (approved):

PR #409, opened 2023-09-12
Shutdown bug

McCallisterRomer:
The sample was exiting abnormally for me due to an imgui shutdown function being called at the wrong time. While looking into this, I realized these init/shutdown functions are split between `Imgui` and `UISystemImpl`. Refactoring all of this to fix the bug and (hopefully) make things a little clearer:
1. Moved all imgui impl init/shutdown funcs to one place.
2. Removed the delayed initialization in `Imgui`.
3. Moved the (now renamed) `UISystem` class inside `NcGraphics` and made it look more like the other systems.


source/engine/graphics/api/vulkan/VulkanGraphics.cpp
```diff
@@ -119,6 +111,7 @@ bool VulkanGraphics::FrameBegin()
     }
 
     m_frameManager->Begin();
+    m_imgui->FrameBegin();
```
jaremieromer, line 114: This is SO much nicer

source/engine/graphics/api/vulkan/VulkanGraphics.cpp
```diff
@@ -130,6 +123,7 @@ void VulkanGraphics::Draw(const PerFrameRenderState& state)
     OPTICK_CATEGORY("VulkanGraphics::Draw", Optick::Category::Rendering);
     auto* currentFrame = m_frameManager->CurrentFrameContext();
     if (m_isMinimized) return;
+    m_imgui->Frame();
```
jaremieromer, line 126: awesome

jaremieromer (approved):
such an improvement!

PR #413, opened 2023-09-22
Config refactor and tests

McCallisterRomer:
Making following changes related to the config:
- Log config options on engine initialization
- Adding tests
- Adding documentation
- Better whitespace handling
- Updating section names to better match their corresponding struct names ([asset_settings])
  - I'm wondering if we should just make the sections/keys exactly match the struct/member names?

For tracing + tests, it was a little bit annoying dealing with the existing Load/Save implementations so I did some refactoring. I may have gotten a bit carried away and rewrote the whole thing... :)


source/engine/engine/NcEngineImpl.cpp
```diff
@@ -11,8 +11,30 @@
...
+    for (auto c : stream.str())
+    {
+        configStr.push_back(c);
+        if (c == '\n')
+        {
+            configStr.push_back('\t');
+        }
+    }
+
+    NC_LOG_INFO_FMT("Initializing NcEngine with Config:\n{}", configStr);
```
jaremieromer, line 35: This will be very useful!

jaremieromer (approved):

PR #415, opened 2023-09-27
TaskGraph bug + refactor

### LINKED ISSUE #414: Task exceptions abort program
At some point in past refactoring, we stopped handling task exceptions well. They now (at least in some cases) are causing aborts. Investigate options for better handling of this. Also add unit test coverage.

McCallisterRomer:
resolves #414

Found several task-related issues:
- A previous refactor left us unable to check/rethrow task exceptions.
- Some potentially throwing tasks are not guarded.
- Empty phases are not handled correctly and can potentially leave dependencies unset in the primary graph. 
- `Executor` can rebuild the graph while it's running.
- `Executor` can try to run again while already running (theoretically; very unlikely to actually happen).
- There are referenced `tf::Taskflow`s whose lifetimes are managed separately from `TaskGraph`/`Executor`, making it difficult to guarantee references remain valid.
- `TaskGraph` is mostly dead code.

Addressing these with a refactor of `Module`/`TaskGraph`/`Executor`:

`TaskGraph` is now a builder interface over a context object - execution-related code is removed. There is only one graph/context at a time, and each `Module` is supplied a reference to it during the graph building process. For common cases, tasks are automatically guarded when added to the graph, while explicit guarding is still possible. Referenced data (i.e. `tf::Taskflow`) is sunk into the context to alleviate lifetime concerns.

When connecting task dependencies, empty phases get a placeholder task so there's always something to connect phases together. Once finalized, the context is extracted to the `Executor`. `Executor::Run()` is now a blocking call (not necessary, that's just how we're using it) and checks for exceptions before returning.

Also added documentation + tests.


include/ncengine/NcEngine.h
```diff
@@ -50,8 +50,12 @@ class NcEngine
         virtual auto GetModuleRegistry() noexcept -> ModuleRegistry* = 0;
 
         /**
-         * @brief Compose a new task graph from all registered modules. This is required
-         *        only if module registration is changed after calling Start().
+         * @brief Compose a new task graph from all registered modules.
+         * 
+         * A graph is built on engine initialization. It only needs rebuilding
+         * if modules are added afterwards. Only call this from the main thread
```
jaremieromer, line 56: Change to "Only call this from the main thread **and** when the graph is not currently..." to make it clear these are two separate requirements

include/ncengine/module/ModuleRegistry.h
```diff
@@ -74,7 +74,7 @@ class ModuleRegistry
         /**
          * @brief Get the collection of all registered modules.
          */
-        auto GetAllModules() noexcept -> std::vector<std::unique_ptr<Module>>&
+        auto GetAllModules() noexcept -> const std::vector<std::unique_ptr<Module>>&
```
jaremieromer, line 77: You love to see a good const add

include/ncengine/type/StableAddress.h
```diff
@@ -0,0 +1,22 @@
+#pragma once
+
+namespace nc
+{
+/**
+ * @brief Base class for non-copyable non-movable types. Additionally, all types
```
jaremieromer: Per our discussion - reword or remove "guaranteed to exist"

include/ncengine/task/ExceptionContext.h
```diff
@@ -0,0 +1,80 @@
...
+#include <mutex>
+#include <utility>
+
+namespace nc::task
+{
+/** @brief State shared between all tasks in a graph for storing an exception. */
+class ExceptionContext : StableAddress
+{
+    public:
+        /** @brief Attempt to store the current exception. Only */
```
jaremieromer: Looks like this comment was never completed

include/ncengine/task/ExceptionContext.h
```diff
@@ -0,0 +1,80 @@
...
+            {
+                return;
+            }
+
+            m_exception = e;
+        }
+
+        /**
+         * @brief Rethrow an exception caught during task execution, if one was set.
+         * @note Only call this from the main thread after the graph has finished.
```
jaremieromer: Again here - unless I'm misunderstanding; will you update the note to say

" Only call this from the main thread, and only after the graph has finished."

source/engine/task/Executor.cpp
```diff
@@ -1,126 +1,120 @@
...
+    builder.Connect(ExecutionPhase::Free, ExecutionPhase::PreRenderSync);
+    builder.Connect(ExecutionPhase::PreRenderSync, ExecutionPhase::Render);
+    builder.Connect(ExecutionPhase::Render, ExecutionPhase::PostFrameSync);
+    return builder.ReleaseContext();
+}
 
-        return out;
-    }
+Executor::Executor(std::unique_ptr<TaskGraphContext> ctx)
+    : m_executor{8},
```
jaremieromer: Will you add an issue to make this configurable - and, for now - make it a named constant?

include/ncengine/task/TaskGraph.h
```diff
@@ -0,0 +1,114 @@
...
+#include <array>
+#include <vector>
+
+namespace nc::task
+{
+/** @brief The number of execution phases in the primary task graph. */
+constexpr size_t ExecutionPhaseCount = 7ull;
+
+/** @brief Identifies an execution phase in the engine's primary task graph. */
+enum class ExecutionPhase : uint8_t
```
jaremieromer, line 24: Add that order matters in the comment - 
I don't have a great suggestion for this, but it would be great to have some indication of the phase's order when Intellisense offers the enum values.

eg:

```task::ExecutionPhase::01Logic```

I'm not a fan of numbering because of the obvious problem of inserting a new phase between existing numbers, so I don't really have a good solution, and it's not a required change by any means. Something to think about though - as it's written there will be a good amount of jumping back to this header to ensure your phase succeeds/precedes some other phase

File: include/ncengine/task/TaskGraph.h [hunk 8]
```diff
@@ -0,0 +1,114 @@
...
+        template<std::invocable<> F>
+        auto Add(ExecutionPhase phase, std::string_view name, F&& func) -> tf::Task
+        {
+            return Schedule(phase, Emplace(name, std::forward<F>(func)));
+        }
+
+        /**
+         * @brief Schedule a tf::Taskflow to run during a phase.
+         * @param phase The target phase.
+         * @param name A user-friendly name for the task.
```
jaremieromer: Recommend changing this comment to:
```* @param name A user-friendly name for the task graph.```
To differentiate it from above

File: source/engine/ecs/EcsModule.cpp [hunk 9]
```diff
@@ -0,0 +1,51 @@
...
+#include "time/Time.h"
+#include "utility/Log.h"
+
+#include "optick/optick.h"
+
+namespace nc::ecs
+{
+auto BuildEcsModule(Registry* registry) -> std::unique_ptr<EcsModule>
+{
+    NC_LOG_TRACE("Creating Ecs Module");
```
jaremieromer: Extremely minor, but logging ECS as Ecs here gives me the willies lol

jaremieromer (approved):
Really, really good work!

source/engine/task/Executor.cpp
McCallisterRomer: Already made one for it. Constant added.

PR #429, opened 2023-10-02
Added ability to dump Vulkan api calls for 0th frame in debug mode

jaremieromer:
Added debug functionality to dump the vulkan api calls for the 0th frame to a text file in the sample's install directory.


CMakeLists.txt
```diff
@@ -73,3 +73,8 @@ install(DIRECTORY   ${PROJECT_SOURCE_DIR}/source/external/imgui
         DESTINATION include
         PATTERN     "src" EXCLUDE
 )
+
+# Install Vulkan Validation Layer Settings
+install(FILES   ${PROJECT_SOURCE_DIR}/vk_layer_settings.txt
```
McCallisterRomer: Since its installing to the sample dir, I think this should be in the sample's cmake.

CMakeLists.txt
```diff
@@ -73,3 +73,8 @@ install(DIRECTORY   ${PROJECT_SOURCE_DIR}/source/external/imgui
...
+        DESTINATION sample
+)
```
McCallisterRomer: newline

source/engine/graphics/api/vulkan/core/Instance.cpp
```diff
@@ -8,7 +8,7 @@
 
 namespace
 {
-constexpr auto g_validationLayers = std::array<const char*, 1>{ "VK_LAYER_KHRONOS_validation" };
+constexpr auto g_validationLayers = std::array<const char*, 2>{ "VK_LAYER_KHRONOS_validation", "VK_LAYER_LUNARG_api_dump" };
```
McCallisterRomer, line 11: Neat, does this just automatically look for the file in the current directory? Is it well behaved if we don't have the file, but still want validation layers?

vk_layer_settings.txt
```diff
@@ -0,0 +1,136 @@
+# The settings in this file can be used to configure the behavior of
```
McCallisterRomer: Should this go inside some directory? sample or script?

McCallisterRomer (approved):

vk_layer_settings.txt
jaremieromer: Wasn't sure where the best place was. I like sample

source/engine/graphics/api/vulkan/core/Instance.cpp
jaremieromer: Yes it does look in the current dir. It more or less is well behaved (it defaults to stdout, so it's a lot of text on the console, though only when you have validation layers enabled which is already signing up for text in stdout)

PR #430, opened 2023-10-02
Adding more CI job variants

McCallisterRomer:
Adding CI jobs to build two variants on each platform. Dev is the default with the editor/debug rendering/asserts and additionally has profiling on. Prod excludes all of this and the sample.


source/engine/ui/editor/Editor.h
```diff
@@ -1,21 +1,31 @@
...
 #include "ecs/Registry.h"
 
 namespace nc::ui::editor
 {
-    class Editor
-    {
-        public:
-            Editor();
-            void Frame(float* dt, Registry* registry);
+class Editor
```
jaremieromer, line 7: This file looks like an accidental ride along. Unless I'm mistaken, remove

jaremieromer (approved):
Remove Editor.h stuff

Issue #420, opened 2023-09-28
Determine path for serialization + widgets


McCallisterRomer:
Within NcEditor, each component type needs a way to expose its properties through the UI and be saved/loaded. We need to determine how game-side components will provide this functionality to the editor.

Q: How will the editor loop through each component type?
We'd like to be able to say something like:
```
for (auto& c : registry->ForEachComponent(entity))
{
    Serialize(stream, c);
}
```
This doesn't work for pooled components, since there's no common base class. We don't have any way to loop through any sort of 'type list' of all registered types - we just have a list of `PerComponentStorageBase`.
  - Could we make some common interface wrapper that exposes funcs like `Serialize()` and `UIWidget()`?
  - Could we manage a type list somehow and loop over that? The use would need to 'set' this somehow.

Q: What does (de)serialization interface look like?
Component constructors may need access to all kinds of things during deserialization + they need to be constructed through the registry.

Q: How do we handle logic components + other user logic? If we load a scene by deserialization, maybe there is some game function that should run after that process to cover something not possible during deserialization.

ClickerMonkey:
>  If we load a scene by deserialization, maybe there is some game function that should run after that process to cover something not possible during deserialization.

This crossed my mind as well - personally I broke up creating a scene into two parts. First step was defining entity/component/storage stuff as well as something like the "logic types" where you map a function definition to some type. The second setup function was to initialize the data when the scene was not being restored from files. Then there is an amount of validation that needs to be done (if the data in the scene depends on something that was removed that's not good. like if you had EdgePanCamera logic the last game run and then you removed it but there's a serialized entity that depended on it.

Just some unsolicited thoughts!

McCallisterRomer:
In a previous POC I had made it possible to mix source code + deserialization during scene initialization, but never got far enough to handle logic serialization. So you could do something like, but `PostLoad()` would always need to know how to setup logic components:
```
void MyScene::Load()
{
    Init();
    HasSaveData() ? Deserialize(GetSaveData()) : FirstTimeInit();
    PostLoad();
}
```

I've started with the first question, but I need to figure out what it would look like to add some sort of function registry to track logic types for serialization (and possibly selecting/assigning functions through the editor).

PR #436, opened 2023-10-05
Add AnyComponent

### LINKED ISSUE #423: Move component registration
Engine components are registered in the constructor of `Registry`. There's no reason to make all testing that depends on the registry either dependent on every component or mock Registry.cpp. Move those calls to some other function and have the engine run it on initialization. It may be necessary to have the registry take some init callback in its constructor if the engine needs types registered right away (perhaps for initialization of other members?).

McCallisterRomer:
part of #420
resolves #423

Proposal for how to solve the problem of 'how does one visit all component types attached to an entity?'. This is of primary concern for drawing component UI and (de)serialization.

I'd like to avoid making `ComponentBase` polymorphic for a few reasons:
1. We can, and I argue should, get rid of the `ComponentBase` class all together. Its just giving us a shortcut for looking up a component's parent entity. That information is not critical to all component types, and its already implicitly available from the registry.
2. Giving every component a v ptr imposes limitations on the characteristics a type may have. It would be nice to allow components to be trivially destructible, for example.
3. Having a `virtual DrawUI()` member pretty much means every component source file is going to depend on UI sources. Same thing may be true for other operations. (sources now also depend on imgui, but this proposal changes that)

If we're ok with the requirement that dynamic functions need explicit registration, we could move the polymorphism outside of the class.
- When registering a component type, a `ComponentHandler<T>` may optionally be passed in, which contains handlers for various runtime component operations.
  - Right now that's just `DrawUI()`, but would eventually contain others like `Serialize()`.
  - Default operations are no-ops. Custom types must provide a handler for every desired behavior.
  - As seen before, the function mapping can also be done at compile time through template specialization. This limits us to only using function pointers, but is faster. This doesn't have to be fast, so I'm fine either way.
- `AnyComponent` acts as a polymorphic interface for components, wrapping around a component instance and its handler. Virtual functions just dispatch to the corresponding handler.
- An `AnyComponent` can be constructed component storage, from the base or concrete types. This allows somebody, like the editor, to dispatch runtime operations for everything on an entity without knowing what types are registered.

This PR implements suggested changes only for `Transform`. There are a number of additional issues/todos to complete the work.


include/ncengine/ecs/Registry.h
```diff
@@ -329,4 +339,24 @@ inline void Registry::CommitStagedChanges()
 
     m_entities.CommitStagedChanges();
 }
+
+/** @todo Temporary include for POC - Should add view/iterator support for this. */
```
McCallisterRomer: Need issue

include/ncengine/ecs/AnyComponent.h
```diff
@@ -0,0 +1,45 @@
...
+
+        /** @brief Invoke ComponentHandler::drawUI with the component instance */
+        void DrawUI()
+        {
+            NC_ASSERT(m_impl, "Cannot dispatch through a null AnyComponent.");
+            m_impl->DrawUI();
+        }
+
+    private:
+        //TODO: can store in-place and remove allocation
```
McCallisterRomer: Need issue

include/ncengine/ecs/Component.h
```diff
@@ -51,11 +52,17 @@ class FreeComponent : public ComponentBase
         #endif
 };
 
+/** @brief Type representation of a null component. */
+struct NullComponent : public ComponentBase
```
jaremieromer: Are we using this anywhere but the tests? What is the use case for this other than testing?

include/ncengine/ecs/Component.h
```diff
@@ -76,6 +83,31 @@ struct DefaultStoragePolicy
...
 
+/** @brief Optional callbacks for generic component operations. */
+template<PooledComponent T>
+struct ComponentHandler
+{
+    /** @brief Function type for the DrawUI handler. */
+    using DrawUI_t = std::function<void(T&)>;
+
+    /** @brief Signal type for on add events. */
+    using OnAdd_t = Signal<T&>;
```
jaremieromer: Should this and OnRemove be optional?

source/engine/engine/NcEngineImpl.cpp
```diff
@@ -50,6 +51,14 @@ auto BuildModuleRegistry(nc::Registry* reg,
     moduleRegistry.Register(std::make_unique<nc::Random>());
     return moduleRegistry;
 }
+
+auto BuildRegistry(size_t maxTransforms) -> nc::Registry
+{
+    // Register components immediately so they're available to modules
+    auto registry = nc::Registry{maxTransforms};
+    nc::RegisterEngineComponents(registry);
```
jaremieromer: Not sure if I'm a big fan of this builder method for the following reasons:

Is the purpose of ```nc::RegisterEngineComponents``` being extracted out like this to allow users to create their own instances of ```nc::Registry``` that don't contain engine components via calling the constructor directly?

If so, wouldn't that flow be broken because we call:
```
RegisterComponentType<Tag>();
RegisterComponentType<Transform>();
RegisterComponentType<ecs::detail::FreeComponentGroup>();
``` 
directly in the constructor body?

If that isn't the purpose, and ```nc::registry``` is only supposed to be for engine components, I don't like extracting out like this because it creates a hidden coupling at creation time for the registry

source/engine/ui/editor/EditorControls.h
```diff
@@ -131,7 +131,12 @@ namespace nc::ui::editor::controls
         ImGui::Text("Layer   %d", entity.Layer());
         ImGui::Text("Static  %s", entity.IsStatic() ? "True" : "False");
 
-        ComponentElement<Transform>(registry, entity);
+        /** @todo #425 Only Transform has an AnyComponent interface. Update remaining components. */
+        for (auto& any : registry->GetAllComponentsOn(entity))
```
jaremieromer, line 135: This is nice

source/engine/ui/editor/Widgets.h
```diff
@@ -1,5 +1,4 @@
 #pragma once
-#ifdef NC_EDITOR_ENABLED
 #include "imgui/imgui.h"
```
jaremieromer, line 2: I love to see fewer #ifdefs!

jaremieromer (approved):

source/engine/engine/NcEngineImpl.cpp
McCallisterRomer: The types that are left in the constructor of registry are the ones required for the registry to work. The ones extracted here are the types used in the modules that get built right after this in the NcEngine constructor. Since the factory for the modules is right above this, it made sense to put the registry setup for them next to it. What if I moved the Registry and ModuleRegistry factories to the same file? Like rename ComponentRegistration.h to something else, and put them both there?

include/ncengine/ecs/Component.h
McCallisterRomer: Since the decision to dispatch to onAdd/onRemove is at compile time, I had initially (not this PR, like early registry stuff) tried to make this a `std::conditional` (sort of a compile time variant), so that we could null the type if it wasn't needed, but never got it working completely.

The `std::optional` approach is a little spooky - we'd have compile time in/exclusion of code that accesses it, but runtime in/exclusion of the actual signal. We'd always have to be checking the optional state, somewhat defeating the purpose of compile time stuff.

After looking at this again, I'm thinking maybe this shouldn't have moved here at all, and it should stay in `PerComponentStorage`
1. The purpose of `ComponentHandler` was to drive `AnyComponent` behavior. The signals are a concept totally separate from `AnyComponent`.
2. As far as configuration of the signals, the user should only be be concerned with setting the `StoragePolicy` correctly, and connecting systems to the signals. There is no added value in having the user construct one and pass it in.

include/ncengine/ecs/Component.h
McCallisterRomer: I thought I had checked in another use, but I guess not. I was going to need it at least for #438, and I believe it came up for something else too. It was intended for template contexts where you need to sub in some component type as a template argument, but don't know the concrete component type being processed, like with `AnyComponent`, so like:

You can do stuff like this this:
```cpp
// Where we don't know the component type
// Get size/alignment info for an arbitrary `AnyImplConcrete<>` and declare storage
using Derived = AnyImplConcrete<NullComponent>;
alignas(alignof(Derived)) std::byte buffer[sizeof(Derived)];

// Where we do know the component type
// Construct real impl on the buffer (template substitution must not alter characteristics)
new (buffer) AnyImplConcrete<ActualType>(instance, handler);

// Back where we don't know the type
// Cast to base ptr and happily call any virtual funcs
reinterpret_cast<AnyImplBase*>(buffer)->DrawUI();
```

McCallisterRomer: Ok, I think this can actually be removed now that #442 is merged in. This was only necessary to satisfy the `std::derived_from<ComponentBase>` part of `PooledComponent` - any type can serve as a stand-in now.

McCallisterRomer: Ok, removed

include/ncengine/ecs/Component.h
McCallisterRomer: Moved them back to where they were

PR #442, opened 2023-10-06
Remove component base class requirement

### LINKED ISSUE #439: Remove ComponentBase class requirement for PooledComponent
If we can remove the requirement that components derive from ComponentBase, then a component can be almost any type. Its highly unlikely that ComponentBase is necessary, given that it only stores an entity. Any component can choose to store its parent entity, if the system processing it requires that information. Since an entity list is also maintained in each component storage instance, it should be possible to add a registry function to get an entity given a component instance.

McCallisterRomer:
resolves #439

Removing the requirement that components derive from `ComponentBase`. Whether or not a component should cache its parent entity should be up to the system processing it, the engine doesn't care. This allows us to say that anything moveable can be a component.

Added a helper to the `Registry` for finding an entity given a pointer to a component.


include/ncengine/ecs/Registry.h
```diff
@@ -141,19 +159,19 @@ auto Registry::Add(EntityInfo info) -> Entity
     return handle;
 }
 
-template<std::derived_from<ComponentBase> T, class... Args>
+template<PooledComponent T, class... Args>
```
McCallisterRomer, line 162: All the functions here are the same logic from before, they just go from if constexprs to two separate functions.

include/ncengine/ecs/Registry.h
```diff
@@ -190,47 +208,51 @@ bool Registry::Contains(Entity entity) const
...
+}
+
+template<std::derived_from<FreeComponent> T>
+auto Registry::Get(Entity entity) const -> const T*
+{
+    auto* group = Get<ecs::detail::FreeComponentGroup>(entity);
+    return group ? group->Get<T>() : nullptr;
+}
+
+template<PooledComponent T>
```
McCallisterRomer, line 252: Ok, this guy was new

jaremieromer (approved):

PR #447, opened 2023-10-10
In-place AnyComponent construction

### LINKED ISSUE #438: Construct AnyComponentImpl in-place
We're using a `unique_ptr` in `AnyComponent` just for erasing the concrete impl type. We need to hold only the base object. Since we know the size of the derived type, don't need the allocation to accomplish this, it was just the easiest route. Add some in-situ wrapper around the impl that can be constructed in-place in `AnyComponent`.

```cpp
struct InSitu
{
    using Derived = AnyImplConcrete<NullComponent>;
    alignas(alignof(Derived)) std::byte buffer[sizeof(Derived)];

    // placement new and casting ops...
};
```

McCallisterRomer:
resolves #438

Constructing the `AnyComponent` implementation in-place, rather than behind a `unique_ptr`. This enables copy operations for `AnyComponent` and saves an allocation on each construction.

Because we know every instance of `AnyImplConcrete<T>` will be the same size/alignment (its always 3 ptrs, just of varying types), we don't need the `unique_ptr` to hide any size information of the derived type. We can instead reserve sufficient space and placement new on top of it.

Also adding equality operators for `AnyComponent`.


include/ncengine/ecs/AnyComponent.h
```diff
@@ -11,49 +11,70 @@ class AnyComponent
...
+        /**
+         * @brief Invoke ComponentHandler::drawUI with the component instance.
+         * @throw NcError if invoked on a null AnyComponent or if HasDrawUI() is false.
+         */
         void DrawUI()
         {
-            NC_ASSERT(m_impl, "Invalid use of null AnyComponent.");
-            m_impl->DrawUI();
+            NC_ASSERT(m_storage.HasValue(), "Invalid use of null AnyComponent.");
+            NC_ASSERT(HasDrawUI(), "No drawUI handler is set for this component type.");
```
McCallisterRomer: Would we rather allow calling this if `HasDrawUI() == false` and just do nothing? That's the default behavior, unless a user has set a function, so I think all calling code always has to check `HasDrawUI()` on every item.

include/ncengine/ecs/AnyComponent.h
```diff
@@ -11,49 +11,70 @@ class AnyComponent
...
-        explicit AnyComponent() noexcept = default;
+        constexpr explicit AnyComponent() noexcept
+            : m_storage{} {}
 
         /**
          * @brief Construct an AnyComponent wrapper around a component.
          * @tparam T The underlying component type.
-         * @param obj A pointer to the component instance.
-         * @param handlers A pointer to the handler for T.
+         * @param obj A valid pointer to the component instance.
```
jaremieromer: the parameter names are incorrect here

include/ncengine/ecs/AnyComponent.h
jaremieromer: Per our discussion - be permissive and allow to no-op to get rid of the checks.

jaremieromer (approved):

PR #450, opened 2023-10-11
Add func to get all loaded assets

### LINKED ISSUE #448: Add IAssetService::GetLoaded() function
Asset operations in the editor require some way to ask for all currently loaded assets of a particular type. Add a method and implement for each manager. Add API funcs as well. Ideally, it will return a pure 'view' type:

`auto IAssetService::GetLoaded() const noexcept -> std::span<std::string_view>>`

McCallisterRomer:
resolves #448

Adding functions to all asset manager to get back a list of all loaded asset paths. Needed for populating asset combo controls in the editor.


jaremieromer (approved):
This will be nice to have!

PR #452, opened 2023-10-15
Improve component widgets

### LINKED ISSUE #449: Add asset combos to component widgets
Assets are read only properties in all component widget functions. Update them to use asset combo controls that enumerate all currently loaded assets of their type. Selecting an entry from the combo should update the asset used by the component.

McCallisterRomer:
resolves #449

Adding write capabilities to all existing component UI functions. For the most part this is done by adding simple properties to wrap getters/setters and routing them to the appropriate widget funtion.

Added a few functions + documentation to ImGuiUtility.h and moved it to /include. Game-side component UI likely wants what's in there.

Updated the ui styling so that the default style can be applied by a function in ImGuiUtility.h. The default style colors  in the sample are now those from the external editor.


include/ncengine/physics/Collider.h
jaremieromer, line 1: Adding a file comment here because the file that is throwing was not modified in the PR.

I am getting an exception when I try to set any of the properties in a combo box in any scene.
The exception stems from a failed assert in `imgui.cpp`, line 5343:

```cpp
void ImGui::EndChild()
{
    ...
    IM_ASSERT(window->Flags & ImGuiWindowFlags_ChildWindow);   // Mismatched BeginChild()/EndChild() calls
```
And following the stack trace it is always in `EditorControls.h` line 81, in the `Imgui::EndChild()` line

```cpp
if(ImGui::BeginChild("EntityPanel", {0,0}, true))
{
     if(SelectedEntity.Valid())
           controls::EntityPanel(registry, SelectedEntity);

} ImGui::EndChild();
```
I've poked around and can't find the offending code.

It's possible it is due to my smaller monitor - my web searching seems to suggest this can occur if some elements are offscreen, though I think everything was onscreen for me.

source/engine/ui/editor/ComponentWidgets.cpp
```diff
@@ -10,12 +11,162 @@
...
 namespace
 {
-constexpr auto g_maxAngle = std::numbers::pi_v<float> * 2.0f;
-constexpr auto g_minAngle = -g_maxAngle;
+namespace audio_source_ext
+{
+using T = nc::audio::AudioSource;
+
+constexpr auto audioClipProperty = nc::ui::Property{ &T::GetClip, &T::SetClip, "audioClip" };
+} // namespace audio_ext
```
jaremieromer: // namespace audio_source_ext

source/engine/ui/editor/ComponentWidgets.cpp
jaremieromer, line 1: This is SUCH an improvement!

jaremieromer (changes requested):
I will re-review when you can get that bug figured out. Everything else looks great!

include/ncengine/physics/Collider.h
McCallisterRomer: Hmm, looking at new combo box code, there is an early return which would miss the closing call for the combo box, but that isn't EndChild(), its like EndCombobox() or something?... unless it is the same under the hood.

McCallisterRomer: Yeah, maybe a combo box is a 'child' internally, and we early return without closing it. Then when we go to call EndChild() for our window, we close the combo box -> the mismatch shows up on the child, which isn't closed.

But it doesn't happen on my machine? Will fix either way, but I'm still confused on that one.

McCallisterRomer: Got it to reproduce. Maybe I just never tried this in a debug build? Fixed.

jaremieromer (approved):

PR #463, opened 2023-10-19
Consolidate editors

### LINKED ISSUE #457: Isolate io events to either game or editor
ImGui and game code are receiving the same input events. This is most notable in the sample right now while using the scroll wheel with the editor open - ui scrolls and game zooms. We should be checking WantCaptureXXX and hiding certain events from the game. Should be possible to check this right in the event processing callbacks.

### LINKED ISSUE #459: Consolidate editors
Refactor 'embedded' editor functions to new windows/layout. Add in remaining ui/layout changes from poc (besides serialization/scene selection)
- Remove invisible top-level window (always captures mouse)
- Context menus (maybe don't do component adding/removing)
- Drag + drop/entity hierarchy

McCallisterRomer:
resolves #459
resolves #457

Bringing some more changes from source/editor (and poc) to embedded editor:
- Adding ability to parent entities by drag + dropping in the scene graph.
- Adding easy scene graph context menu options. Adding/removing components has a separate issue.
- Polishing and splitting up scene graph/inspector code.
- Getting rid of the invisible parent window since its always trying to capture mouse input.
  - This enabled easily fixing the bug where mouse wheel will scroll ui and zoom the camera at the same time.


include/ncengine/ui/ImGuiUtility.h
```diff
@@ -29,6 +29,17 @@ constexpr auto g_minAngle = -g_maxAngle;
 constexpr auto g_minScale = 0.0001f;
 constexpr auto g_maxScale = 1000.0f;
 
+/** @brief Create a top-level window. */
+template<class F>
+void Window(const char* label, F&& drawContents);
+
+/** @brief Create a child window. */
+template<class F>
+void ChildWindow(const char* label, F&& drawContents);
```
jaremieromer, line 49: Are there any preconditions in these methods? i.e. Does a ChildWindow require a Window to be created? If so, will you list them in the comments?

include/ncengine/ui/ImGuiUtility.h
jaremieromer, line 1: This file is getting pretty long. I recommend splitting out all the styling into a class like ```ImGuiStyle.h```

source/engine/ui/editor/Editor.cpp
```diff
@@ -1,67 +1,33 @@
...
-    namespace hotkey
-    {
-        constexpr auto Editor = nc::input::KeyCode::Tilde;
-        constexpr auto Utilities = nc::input::KeyCode::F1;
-    }
+namespace hotkey
+{
+constexpr auto Editor = nc::input::KeyCode::Tilde;
 }
+} // anonymouse namespace
```
jaremieromer: A mousey typo :D

source/engine/ui/editor/EditorUI.cpp
```diff
@@ -0,0 +1,47 @@
+#include "EditorUI.h"
+#include "ncengine/ecs/Registry.h"
+#include "ncengine/ui/ImGuiUtility.h"
+#include "ncengine/window/Window.h"
+
+#define RUN_ONCE(f) { [[maybe_unused]] static auto _ = [](){f; return true;}(); }
```
jaremieromer, line 9: A comment here explaining what this does for those that haven't seen Daisy's talk would be very useful

jaremieromer (approved):
So good!

include/ncengine/ui/ImGuiUtility.h
McCallisterRomer: You can make a child window without a parent. I think there is some ephemeral debug window always lurking under the surface, and you get put inside that. You can also make a normal window from inside a child window without issues, although it doesn't become a child. Its fairly forgiving.

But now that you bring it up, all of these functions need to be called only from within a ui callback, so like an `IUI` implementation or a DrawUI handler. I wonder if there's a way to make some kind of file/scoped comment that doxygen would understand.

McCallisterRomer: With a `@file` comment, I can add a `@note` comment to the the whole file.

PR #464, opened 2023-10-21
Editor interface

### LINKED ISSUE #427: Make interface for editor
The editor should have an abstract interface (maybe reuse the same one game UI?) and factory function that builds either the real editor or a stubbed, depending on NC_EDITOR_ENABLED. Editor code can still be conditionally included/excluded, but that can be gated at the cmake level. See source/engine/ui/Editor.h for some #define cleanup opportunities.

### LINKED ISSUE #462: Delete duplicate editor files
Remove files from editor/source once all changes there are integrated into the 'embedded' editor.

McCallisterRomer:
resolves #462
resolves #427

Adding an interface for the editor and trying to include/exclude editor code at the CMake level, rather than with #ifdefs. The editor source dir has impl and stub subdirectories. Currently, which one you get depends on NC_PROD_BUILD, but I've added an issue to make a CMake option for NC_EDITOR_ENABLED so we can use that more directly in the future.

Removing the now redundant code at /source/editor.


source/engine/ui/editor/impl/EditorImpl.cpp
```diff
@@ -0,0 +1,39 @@
+#include "ui/editor/Editor.h"
```
McCallisterRomer, line 1: Should have git mv'd. This file is exactly Editor.cpp, but now has a factory function.

source/engine/ui/editor/impl/EditorImpl.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include "EditorUI.h"
+#include "input/Input.h"
+#include "window/Window.h"
+
+namespace
+{
+namespace hotkey
+{
+constexpr auto Editor = nc::input::KeyCode::Tilde;
+}
```
jaremieromer: Comment the closing of the hotkey namespace too

jaremieromer (approved):

PR #470, opened 2023-10-21
Adding header with forward declarations

### LINKED ISSUE #466: Add fwd decl header
Add a file like NcFwd.h to the API that forward declares common types. We have one or two places that forward declare every engine component, and we'll need to a few more of those in component factory and serialization files. It'd be nice to have that that all in one place.

McCallisterRomer:
resolves #466

Adding a header just to forward declare common general/component types. Not sure how much to include, but Registry + component types come up a lot. I had copy-pasted these from ComponentWidgets to another file I was working on and realized I'd probably do the same again for serialization. Better to keep them in one place.


include/ncengine/NcFwd.h
```diff
@@ -0,0 +1,67 @@
+/**
+ * @file NcFwd.h
+ * @copyright Jaremie and McCallister Romer 2023
```
jaremieromer, line 3: nice

jaremieromer (approved):

Issue #467, opened 2023-10-21
Default asset improvements for editor


McCallisterRomer:
- All asset types need a default. We should guarantee that these are all loaded at init time.
  - Creating a default constructed `ConcaveCollider` (or whatever else) through the editor requires us to guarantee something always exists.
- Default asset names should be documented in code in a consistent way
- There should be an explicit way to acquire the asset view for a default asset. I.e. rather than asking for it by name, you can say like `AcquireDefaultMesh()`. It might be ok for this functionality to be isolated to the editor.

PR #472, opened 2023-10-23
Add Module::OnSceneLoad()

McCallisterRomer:
needed for #467

Adding a hook for `Module`s to run code right before scene load. Will be used by `NcAsset` to ensure all default assets are loaded before launching into the next scene.

### DIFF
--- a/include/ncengine/module/Module.h
+++ b/include/ncengine/module/Module.h
@@ -1,3 +1,7 @@
+/**
+ * @file Module.h
+ * @copyright Jaremie and McCallister Romer 2023
+ */
 #pragma once
 
 #include <vector>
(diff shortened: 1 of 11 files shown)


include/ncengine/module/Module.h
```diff
@@ -19,8 +23,11 @@ struct Module
     /** @brief Called on registered modules when the task graph is constructed. */
     virtual void OnBuildTaskGraph(task::TaskGraph&) {}
 
-    /** @brief Called prior to clearing the module's associated data registry. This includes
-     *         scene changes and NcEngine::Shutdown(). */
+    /** @brief Called on registered modules prior to loading a new scene. */
+    virtual void OnSceneLoad() {}
```
jaremieromer: Should this be: 
`virtual void OnBeforeSceneLoad()`

The pattern already in place here with `OnBuildTaskGraph` is that "OnX" would run **after** X, unless I'm misreading the task graph comment.

jaremieromer (approved):

PR #474, opened 2023-10-24
Default assets

### LINKED ISSUE #467: Default asset improvements for editor

McCallisterRomer:
resolves #467 

Many files changed due to asset renaming/rebuilding.

- Adding string constants for all default assets and updated hardcoded paths to reference these.
- The `NcAsset` factory now takes an `AssetMap` of all default engine assets. This is used in `OnBeforeSceneLoad()` to ensure everything is loaded before jumping into scene/game code.
- Removing duplicated default assets from the sample. CMake install now merges resources/nca with sample/assets/nca so we only have to maintain in one place.
- Made naming/casing of default assets consistent. Everything is PascalCase and prefixed with 'Default'.

No changes were made for shaders since they don't yet behave like other assets.


include/ncengine/asset/DefaultAssets.h
jaremieromer, line 1: We also should have the default skybox texture

resources/manifest.json
```diff
@@ -2,44 +2,50 @@
...
         {
-            "sourcePath": "raw\\model\\cube.fbx",
-            "assetName": "meshes\\skybox"
+            "sourcePath": "raw\\model\\DefaultSphere.fbx",
+            "assetName": "mesh\\DefaultSphere"
         },
         {
-            "sourcePath": "raw\\model\\sphere.fbx",
-            "assetName": "meshes\\sphere"
+            "sourcePath": "raw\\model\\DefaultSkybox.fbx",
```
jaremieromer, line 47: We should rename to DefaultCubeMap

jaremieromer (approved):

resources/manifest.json
McCallisterRomer: I wasn't sure how to handle the names of these two since we have one for the mesh and one for the texture. To me a cubemap is purely a texture. I'm slightly opposed to referring to a mesh as a cubemap. Not sold on skybox though.

include/ncengine/asset/DefaultAssets.h
McCallisterRomer: Along with the other comment, we have `DefaultCubeMap`. I'm fine with changing that to default skybox, but if I leave the mesh named `DefaultSkybox`, things get weird.

McCallisterRomer: Thoughs on `DefaultSkyboxMesh` and `DefaultSkyboxTexture`?

resources/manifest.json
jaremieromer: That makes total sense - I think skybox is a fine name for it in that case. We need some way to differentiate it because the normals face inwards, IIRC.

include/ncengine/asset/DefaultAssets.h
jaremieromer: I like `DefaultSkyboxMesh` and `DefaultSkyboxTexture`

PR #480, opened 2023-12-14
ECS refactor

### LINKED ISSUE #440: Investigate breaking up PerComponentStorage
Don't know what's possible given how things are used, but the interfaces for `PerComponentStorageBase` and `PerComponentStorage<T>` are messy and manage a few things. The sparse set concept is split between the two. Investigate ways to restructure this. Ideally, we would extract an entire new `SparseSet<T>` class and add tests.

McCallisterRomer:
resolves #440

Posting everything at once as requested - if it gives you megalaphobia, I can split it up pretty easily.

This refactor is came out of some of the editor POC work. The various ECS types were already a bit gnarly and large, and it felt wrong to keep piling on top.

- `SparseSet` - Manages a component array and the entity-component mapping. This responsibility was previously split between `PerComponentStorageBase` and `PerComponentStorage`.
- `ComponentPoolBase` + `ComponentPool` - Basically `PerComponentStorageBase` and `PerComponentStorage` but with some cruft removed. Added stl-like container members so we can use these in algorithms/ranges/etc.
- `EntityPool` - Replaces `EntityStorage`. Mostly just made the name consistent and pulled it out of the detail namespace. Has the stl members as well.
- `ComponentRegistry` - Replaces the more lower-level parts of `Registry`. Handles type registration, pool lookup, commit/clear, but it doesn't have all the other bells and whistles.
- `EcsInterface` - Replaces the more game-facing parts of the registry - add, remove, get, contains...
  - New feature: This can be filtered to constrain type access. Most places where we're using `Registry` will eventually be changed to this.
- `AccessPolicy` - Handles the filtering for `EcsInterface`.
- `Registry` - This is now redundant, but keeping it around so we can progressively update our usage. The interface here is almost identical, but the implementation is now entirely in terms of `EcsInterface`.

With regards to our current `View` + `MultiView` - These will be fundamentally changed in a future PR. For now I've only made necessary changes. With the addition of a couple of range adaptors, c++20 ranges give us way more power and flexibility with how we view/iterate game state.


include/ncengine/ecs/AccessPolicy.h
```diff
@@ -0,0 +1,105 @@
...
+#include "ncengine/ecs/Transform.h"
+#include "ncengine/ecs/detail/FreeComponentGroup.h"
+
+#include <concepts>
+#include <type_traits>
+
+namespace nc::ecs
+{
+// TODO: make issue to update this
+using Property = nc::FreeComponent;
```
McCallisterRomer: I still need to change this - I don't want to call it `Property`.

include/ncengine/ecs/AccessPolicy.h
```diff
@@ -0,0 +1,105 @@
...
+using PropertyBag = nc::ecs::detail::FreeComponentGroup;
+
+/** @brief Filter option for AccessPolicy providing course control over type access. */
```
McCallisterRomer: typo

include/ncengine/ecs/ComponentPool.h
```diff
@@ -0,0 +1,301 @@
...
+
+        /** @brief Get the T's onRemove event Signal. */
+        auto OnRemove() noexcept -> Signal<Entity>& { return m_onRemove; }
+
+        /** @brief Sort the components according to a predicate. */
+        template<std::predicate<const T&, const T&> Predicate>
+        void Sort(Predicate&& compare);
+
+        /** @brief Get an iterator the the first component in the pool. */
+        auto begin() noexcept { return std::ranges::begin(m_storage.GetPackedArray()); }
```
jaremieromer, line 120: So neat

include/ncengine/ecs/ComponentPool.h
```diff
@@ -0,0 +1,301 @@
...
+        auto end() noexcept { return std::ranges::end(m_storage.GetPackedArray()); }
+
+        /** @brief Get a const_iterator one past the last component in the pool. */
+        auto end() const noexcept { return std::ranges::end(m_storage.GetPackedArray()); }
+
+        /** @brief Get the number of components committed to the pool. */
+        auto size() const noexcept { return Size(); }
+
+        /** @brief Check if there are no components in the pool. */
+        [[nodiscard]] auto empty() const noexcept { return GetComponents().empty(); }
```
jaremieromer, line 135: Why is only `empty()` `[[nodiscard]]`?

include/ncengine/ecs/ComponentPool.h
```diff
@@ -0,0 +1,301 @@
...
+{
+    m_storage.Sort(std::forward<Pred>(compare));
+}
+
+template<PooledComponent T>
+void ComponentPool<T>::Reserve(size_t capacity)
+{
+    m_storage.Reserve(capacity);
+    const auto existing = TotalSize();
+    if (capacity > existing)
```
jaremieromer, line 265: shouldn't this be `>=` ?

include/ncengine/ecs/ComponentRegistry.h
```diff
@@ -0,0 +1,139 @@
...
+#include "ncengine/ecs/ComponentPool.h"
+#include "ncengine/ecs/EntityPool.h"
+
+namespace nc::ecs
+{
+/**
+ * @brief Core collection of data pools and registration information game types.
+ * 
+ * The ComponentRegistry houses all entity and component data and provides a low-level interface over game state. An
+ * instance is created an bootstrapped with engine types during NcEngine initialization. This instance remains valid
```
jaremieromer: typo on `An instance is created an`

include/ncengine/ecs/Ecs.h
```diff
@@ -0,0 +1,226 @@
...
+                return false;
+
+            if constexpr (std::derived_from<T, Property>)
+            {
+                return m_policy.template OnPool<PropertyBag>([entity](auto& pool)
+                {
+                    return pool.Get(entity)->template Remove<T>();
+                });
+            }
+            else
```
jaremieromer, line 93: You don't need this `else` here, and remove it would reduce some of the arrow-like qualities of this file

include/ncengine/ecs/Ecs.h
```diff
@@ -0,0 +1,226 @@
...
+        template<Component T>
+            requires HasAccess<PolicyType, T>
+        auto Get(Entity entity) -> T*
+        {
+            if constexpr (std::derived_from<T, Property>)
+            {
+                auto bag = Get<PropertyBag>(entity);
+                return bag ? bag->template Get<T>() : nullptr;
+            }
+            else
```
jaremieromer, line 112: You also don't need this `else` here

include/ncengine/ecs/Ecs.h
```diff
@@ -0,0 +1,226 @@
...
+        template<Component T>
+            requires HasAccess<PolicyType, T>
+        auto Get(Entity entity) const -> const T*
+        {
+            if constexpr (std::derived_from<T, Property>)
+            {
+                auto bag = Get<PropertyBag>(entity);
+                return bag ? bag->template Get<T>() : nullptr;
+            }
+            else
```
jaremieromer, line 131: I really prefer remove `else` blocks if they aren't needed, I think it keeps the code cleaner

include/ncengine/ecs/Ecs.h
```diff
@@ -0,0 +1,226 @@
...
+        auto Contains(Entity entity) const -> bool
+        {
+            if constexpr (std::derived_from<T, Property>)
+            {
+                auto bag = Get<PropertyBag>(entity);
+                return bag ? bag->template Contains<T>() : false;
+            }
+            else
+            {
+                return m_policy.template OnPool<T>([entity](auto&& pool)
```
jaremieromer, line 191: Remove unnecessary `else`

test/ecs/ComponentPool_unit_tests.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+TEST(ComponentPoolTests, GetAsAnyComponent_doesNotExist_returnsNullObject)
+{
+    auto uut = nc::ecs::ComponentPool<S1>{10u, nc::ComponentHandler<S1>{}};
+    const auto entity = nc::Entity{1, 0, 0};
+    const auto actual = uut.GetAsAnyComponent(entity);
+    EXPECT_FALSE(actual);
+}
+
+TEST(ComponentPoolTests, Size_allVariations_retusnExpectedValues)
```
jaremieromer: minor, but typo in test name

include/ncengine/ecs/AccessPolicy.h
```diff
@@ -0,0 +1,101 @@
...
+#include "ncengine/ecs/Transform.h"
+#include "ncengine/ecs/detail/FreeComponentGroup.h"
+
+#include <concepts>
+#include <type_traits>
+
+namespace nc::ecs
+{
+/** @brief Filter option for AccessPolicy providing coarse control over type access. */
+enum class FilterBase
```
jaremieromer, line 21: Add a comment here that this enum is ordered by permissiveness. Even though I can see that the permissiveness increases, it wasn't apparent until parsing the constexpr below that it is required to be ordered.

include/ncengine/ecs/ComponentPool.h
```diff
@@ -0,0 +1,301 @@
...
+        virtual ~ComponentPoolBase() = default;
+
+        /** @brief Check if the pool has a component attached to an entity. */
+        virtual auto Contains(Entity entity) const -> bool = 0;
+
+        /** @brief Get the component attached to an entity as an AnyComponent.
+         *  @note Returns a null object if no component exists. */
+        virtual auto GetAsAnyComponent(Entity entity) -> AnyComponent = 0;
+
+        /** @brief Remove the component attached to an entity */
```
jaremieromer, line 35: Debatable if it's necessary here as well as the docs, also because it behaves like any other iterator now but it'd be nice to drop a comment about invalidating the iterator on remove/reserve/etc for each method

test/ecs/Registry_unit_tests.cpp
```diff
@@ -92,7 +97,8 @@ class Registry_unit_tests : public ::testing::Test
         ~Registry_unit_tests()
         {
             registry.CommitStagedChanges();
-            registry.Clear();
+            registry.Reset();
+            // registry.Clear();
```
jaremieromer: Remove or uncomment

jaremieromer (approved):

include/ncengine/ecs/ComponentPool.h
McCallisterRomer: I made the signatures for these match exactly what is in the stl. `empty()` now has the `[[nodiscard]]` attribute simply because its a frequently misinterpreted name, and the attribute guarantees to catch the misunderstanding. (clear the container vs. check if its empty).

include/ncengine/ecs/Ecs.h
McCallisterRomer: These are constexpr if statements; the else is very important.

include/ncengine/ecs/AccessPolicy.h
McCallisterRomer: Excellent point

include/ncengine/ecs/ComponentPool.h
McCallisterRomer: Looking at the actual reserve call below, if capacity == existing we would just needlessly reserve 0.

PR #484, opened 2023-12-20
Add/remove components through editor 

### LINKED ISSUE #460: Add/remove components in context menu
Add option to add and remove components from the entity node context menu. To support user-defined components, they'll have to be default constructible or there will need to be a way to provide a factory.

McCallisterRomer:
resolves #460

Adding ability to add and remove components through the scene graph and inspector context menus. Discovery of user types is done through through `ComponentHandler` - everything with a factory shows up in the list.

Example of the menu:
Screenshot 2023-12-20 070647

Additionally, I've started the transition from using `Registry` to `EcsInterface` within editor code.


include/ncengine/ecs/Ecs.h
```diff
@@ -50,7 +50,7 @@ class EcsInterface
 
         /** @brief Emplace a component. */
         template<Component T, class... Args>
-            requires HasAccess<PolicyType, T>
+            requires PolicyType::template HasAccess<PolicyType, T>
```
McCallisterRomer: PolicyType should not be a t-arg. Not caught by test?

source/engine/engine/ComponentFactories.h
```diff
@@ -0,0 +1,17 @@
+#pragma once
+
+#include "ncengine/NcFwd.h"
+
+namespace nc
+{
+auto CreateAudioSource(Entity entity, void*) -> audio::AudioSource;
```
McCallisterRomer, line 7: missed unit test for these

source/engine/engine/RegistryFactories.cpp
```diff
@@ -53,6 +54,23 @@ void Register(nc::Registry& registry, const char* name, void(*drawUI)(T&))
         .drawUI = drawUI
     });
 }
+
+template<class T>
+void Register(nc::Registry& registry,
```
McCallisterRomer, line 49: It doesn't look like the `Register` function above this is needed.

include/ncengine/ecs/Component.h
```diff
@@ -87,12 +87,22 @@ struct StoragePolicy : DefaultStoragePolicy {};
...
 
     /** @brief A name for the component with no uniqueness constraints. */
     std::string name = "User Component";
 
+    /** @brief Optional pointer to user data, which is passed to the factory handler. */
+    void* userData = nullptr;
+
+    /** @brief Handler for creating an instance of T.
+     *  @note Factory handlers enable adding/removing types through the editor. */
+    Factory_t factory = nullptr;
```
McCallisterRomer, line 104: We might be able to conditionally use nullptr or a default lambda factory, depending on whether the type is default constructible.

source/engine/engine/ComponentFactories.cpp
```diff
@@ -0,0 +1,71 @@
...
+#include "ncengine/graphics/ParticleEmitter.h"
+#include "ncengine/graphics/MeshRenderer.h"
+#include "ncengine/graphics/ToonRenderer.h"
+#include "ncengine/network/NetworkDispatcher.h"
+#include "ncengine/physics/Collider.h"
+#include "ncengine/physics/PhysicsBody.h"
+
+namespace nc
+{
+auto CreateAudioSource(Entity entity, void*) -> audio::AudioSource
```
jaremieromer, line 15: I am not a huge fan of having the majority of these methods accept a parameter they don't use. If it's an acceptable performance hit, I'd much rather they accept the param, and propagate if not null. Then a user would not have to make changes to the engine to pass their own userdata in

source/engine/ui/editor/impl/windows/Inspector.cpp
```diff
@@ -16,23 +22,30 @@ void ElementHeader(std::string_view name)
...
+        }
+
         ElementHeader("Entity");
+        DragAndDropSource<Entity>(&entity);
         ImGui::Text("Index   %d", entity.Index());
         ImGui::Text("Layer   %d", entity.Layer());
         ImGui::Text("Static  %s", entity.IsStatic() ? "True" : "False");
-
-        for (auto& any : registry->GetAllComponentsOn(entity))
+        std::ranges::for_each(world.GetComponentPools(), [entity](auto&& pool)
```
jaremieromer, line 40: so neat that it works this way

jaremieromer (approved):

include/ncengine/ecs/Component.h
McCallisterRomer: Leaving this out for now. Can adjust later, if needed.

source/engine/engine/ComponentFactories.cpp
McCallisterRomer: discussed

source/engine/engine/ComponentFactories.h
McCallisterRomer: As discussed, these are a bit more involved than what we want in our unit tests. Time is better spent removing the problematic bits. Issue added.

PR #486, opened 2023-12-22
Updated NcEngine to consume latest NcTools converter and library

### LINKED ISSUE #422: Target exact version of dependencies
Update everything we build from source (across all 3 repos) to always target an exact version. Can be tag or commit hash. Without this it is very difficult to build from old commits or branches because they might just reference 'master'.

jaremieromer:
Updated NcEngine to consume v3.0.0 of NcTools.
Regenerated all assets in sample and resources using new nc-convert.exe.

Combining additional changes from #485 with this to satisfy build:
resolves #422 #481

- Updating the following:
  - NcCommon/NcTools
  - glfw
  - taskflow
  - gtest
- Changing all dependencies to target an actual tag, rather than branch/commit hash
  - Commit hashes with shallow cloning isn't allowed, at least for the last several CMake versions. Older versions don't complain, but it has been a Cadillac-sized thorn.
  - Using a branch name is the most convenient, but we've hit build issues in the past when needing to build from an older commit.


McCallisterRomer (approved):
Thanks!

McCallisterRomer:
~~Shoot, I think the tests are failing because collateral needs to be rebuilt as well, but I don't see the raw files anywhere?!?~~ fixed

McCallisterRomer:
@jaremieromer I can't request a review from you since its your PR, but will you take a quick look at my changes before merging?

PR #489, opened 2023-12-23
Fix TransformUIWidget rotation

McCallisterRomer:
A bug as old as time. I guess we never made an issue for this?


jaremieromer (approved):
Wahoo!!!!!! All my dreams have come true

PR #491, opened 2023-12-26
Skeletal Animation 01 - Asset Managing

jaremieromer:
### Goal
Add support for managing skeletal animation assets into NcEngine. 

### Background
`SkeletalAnimation` assets are generated from FBX files that contain animation clips. NcTools converts the FBX files into `.nca` files. This PR allows import of those `.nca` files.

### Changes

 1. Added `SkeletalAnimationAssetManager` to load/unload `SkeletalAnimation` asset files.
 2. Modified `MeshAssetManager` to parse armature data from mesh `.nca` files.
 3. Added unit test for `SkeletalAnimationAssetManager`
 4. Fixed namespace spacing in a few Asset files


include/ncengine/config/Config.h
```diff
@@ -31,6 +31,7 @@ struct AssetSettings
     std::string hullCollidersPath = "resources/assets/hull_colliders/";
     std::string meshesPath = "resources/assets/meshes/";
     std::string shadersPath = "resources/assets/shaders/";
+    std::string skeletalAnimationsPath = "resources/assets/skeletal_animations/";
```
McCallisterRomer, line 34: Config.cpp will also need to be updated.

include/ncengine/asset/Assets.h
```diff
@@ -13,148 +12,162 @@
...
+template<class T>
+concept AssetView = std::same_as<T, AudioClipView>       ||
+                    std::same_as<T, ConvexHullView>      ||
+                    std::same_as<T, ConcaveColliderView> ||
+                    std::same_as<T, MeshView>            ||
+                    std::same_as<T, TextureView>         ||
+                    std::same_as<T, CubeMapView>         ||
+                    std::same_as<T, ShaderView>          ||
+                    std::same_as<T, SkeletalAnimationView>;
+}
```
McCallisterRomer: Several files have had their EOF newline eaten.

source/engine/assets/manager/SkeletalAnimationAssetManager.cpp
```diff
@@ -0,0 +1,130 @@
...
+bool SkeletalAnimationAssetManager::Load(std::span<const std::string> paths, bool isExternal, asset_flags_type)
+{
+    if (m_assetIds.size() + paths.size() >= m_maxSkeletalAnimationCount)
+    {
+        throw NcError("Cannot exceed max skeletal animations count.");
+    }
+
+    bool anyLoaded = false;
+
+    for(const auto& path : paths)
```
McCallisterRomer, line 53: consider `any_of`

source/engine/assets/manager/SkeletalAnimationAssetManager.cpp
```diff
@@ -0,0 +1,130 @@
...
+            std::span<const asset::SkeletalAnimation>{m_skeletalAnimations},
+            asset::UpdateAction::Load
+        });
+    }
+    return anyLoaded;
+}
+
+bool SkeletalAnimationAssetManager::Unload(const std::string& path, asset_flags_type)
+{
+    auto pos = std::find(m_assetIds.begin(), m_assetIds.end(), path);
```
McCallisterRomer: Prefer ranges here so you don't have to name two iterators. Its slightly less error prone (not that this is wrong).

source/engine/graphics/api/vulkan/Initializers.cpp
```diff
@@ -40,7 +40,7 @@ auto CreatePipelineShaderStageCreateInfo(ShaderStage stage, const vk::ShaderModu
     };
 }
 
-vk::PipelineVertexInputStateCreateInfo CreateVertexInputCreateInfo(const vk::VertexInputBindingDescription& vertexInputDesc, const std::array<vk::VertexInputAttributeDescription, 5>& vertexAttributeDesc)
+vk::PipelineVertexInputStateCreateInfo CreateVertexInputCreateInfo(const vk::VertexInputBindingDescription& vertexInputDesc, const std::array<vk::VertexInputAttributeDescription, 7>& vertexAttributeDesc)
```
McCallisterRomer: Not major, but this could take a span instead so it doesn't need to be touched if the size changes again.

source/engine/assets/NcAssetImpl.h
```diff
@@ -32,17 +33,20 @@ class NcAssetImpl : public NcAsset
 
         void OnBeforeSceneLoad() override;
 
+        auto OnBoneUpdate() noexcept -> Signal<const BoneUpdateEventData&>& override;
```
McCallisterRomer, line 36: My initial thoughts are that bones/meshes are not separate assets, therefore they (maybe?) shouldn't need different signals. I'm looking for the receiving code to better understand, but I'm only seeing the emit calls. Is there anything subscribing to this event in this PR, or is that coming down the pike?

McCallisterRomer (changes requested):
Requesting changes for additional config updates.

source/engine/assets/NcAssetImpl.h
jaremieromer: My personal preference is to leave it as two signals - we could combine it but I think it makes SkeletalAnimationSystem less clear, and we'd have to do "onSignalReceived() { if mesh.bonesData.has_value() ... } in the system as well, which I'd rather avoid

McCallisterRomer (approved):

PR #493, opened 2024-01-01
Skeletal Animation Part II: the system and the calculations.

jaremieromer:
### Goal
Add the system to process `SkeletalAnimator.h` components into a buffer of vertex-to-animated-position matrices and send them to the GPU. Note: For ease of reviewability - this PR ends before updating shader binaries to handle the data, so no animations are rendered. This PR also excludes the state machine.

### Background
`SkeletalAnimator` components integrate with the `SkeletalAnimationSystem.h` via registering a `Signal<const anim::PlayState&>`. 
The system owns polling the `SkeletalAnimator` components who will emit the signal to the system.
When the signal emits, the system will add the data to its working vector of playing animations (`std::vector<anim::UnitOfWork>`). 
For each unit, the system will blend between the past and current animation (signals are only emitted on transitions between animation states), then it will perform all the matrix combinations required and output the vector of transformations to a buffer.
The buffer is then sent to the GPU.

### Notes
Notes on Optick.h for unit tests: It seems that we have never supported Optick in unit tests. The `Executor` unit test and CPP did include Optick, but `Executor.cpp` did not actually invoke it at all, which seems to be why that test worked. I've removed Optick both from that test, its calling code (`Executor.cpp`) and from the `SkeletalAnimationCalculations_test` and `SkeletalAnimationCalculations.cpp`. Profiling is still used in the larger calling code for skeletal animation processing.


include/ncengine/graphics/SkeletalAnimationTypes.h
```diff
@@ -0,0 +1,80 @@
...
+    std::string meshUid;
+    std::string prevAnimUid;
+    std::string curAnimUid;
+    Action action;
+    Entity entity;
+};
+
+struct Initial
+{
+    Initial(std::string animUid_);
```
McCallisterRomer: Inline all of these constructors that are 'simple' - aka it doesn't require additional includes or utility functions.

include/ncengine/graphics/SkeletalAnimationTypes.h
```diff
@@ -0,0 +1,80 @@
...
+    std::vector<DirectX::XMMATRIX> boneToParent;
+    DirectX::XMMATRIX globalInverseTransform;
+    std::vector<std::string> boneNames;
+    std::vector<std::tuple<uint32_t, uint32_t>> boneToParentIndices;
+    std::vector<uint32_t> offsetsMap;
+};
+
+struct DecomposedMatrix
+{
+    nc::Vector3 pos;
```
McCallisterRomer: Unnecessary ns qualification

source/engine/graphics/system/SkeletalAnimationTypes.cpp
```diff
@@ -0,0 +1,68 @@
+#include "ncengine/graphics/SkeletalAnimationTypes.h"
+#include "ncasset/Assets.h"
+
+namespace
+{
+auto GetGlobalInverseTransform(const std::vector<nc::asset::BoneSpaceToParentSpace>& offsets) -> DirectX::XMMATRIX
+{
+    const auto pos = std::ranges::find_if(offsets, [](const auto& offset)
+    {
+        return offset.boneName.find(std::string{"Root"}) != std::string::npos;
```
McCallisterRomer: No need to allocate a string each time, or at all:
```cpp
static constexpr auto root = std::string_view{"Root"};
return offset.boneName.find(root) != std::string::npos;
```

source/engine/graphics/system/SkeletalAnimationTypes.cpp
```diff
@@ -0,0 +1,68 @@
...
+{
+Initial::Initial(std::string animUid_)
+    : animUid{std::move(animUid_)}{}
+
+State::State(const Initial& initialProperties)
+    : id{},
+      action{Action::Loop},
+      animUid{std::move(initialProperties.animUid)}{}
+
+PackedRig::PackedRig(nc::asset::BonesData bonesData)
```
McCallisterRomer: `BonesData` is heavy and is not a sink param here - take by const ref.

source/engine/graphics/system/SkeletalAnimationTypes.cpp
```diff
@@ -0,0 +1,68 @@
...
+    auto& bspsVec = bonesData.boneSpaceToParentSpace;
+    auto& vsbsVec = bonesData.vertexSpaceToBoneSpace;
+
+    vertexToBone.reserve(vsbsVec.size());
+    boneToParent.reserve(bspsVec.size());
+    boneNames.reserve(bspsVec.size());
+    boneToParentIndices.reserve(bspsVec.size());
+    offsetsMap.reserve(vsbsVec.size());
+
+    std::transform(vsbsVec.begin(), vsbsVec.end(), std::back_inserter(vertexToBone),
```
McCallisterRomer: Prefer ranges to guarantee to mismatched begin/end iterators.

source/engine/graphics/system/SkeletalAnimationTypes.cpp
```diff
@@ -0,0 +1,68 @@
...
+                [](const nc::asset::VertexSpaceToBoneSpace& vsbs) -> DirectX::XMMATRIX { return vsbs.transformationMatrix; });
+
+    std::transform(bspsVec.begin(), bspsVec.end(), std::back_inserter(boneToParent),
+                [](const nc::asset::BoneSpaceToParentSpace& bsps) -> DirectX::XMMATRIX { return bsps.transformationMatrix; });
+
+    std::transform(bspsVec.begin(), bspsVec.end(), std::back_inserter(boneNames),
+                   [](const nc::asset::BoneSpaceToParentSpace& bsps) -> std::string { return bsps.boneName; });
+
+    std::transform(bspsVec.begin(), bspsVec.end(), std::back_inserter(boneToParentIndices),
+                [](const nc::asset::BoneSpaceToParentSpace& bsps) -> std::tuple<uint32_t, uint32_t> { return std::make_tuple(bsps.indexOfFirstChild, bsps.numChildren); });
```
McCallisterRomer: Why tuple over pair or a custom struct? I find the tuple interface to be very non-ergonomic.

source/engine/graphics/system/SkeletalAnimationTypes.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+    for (const auto& vsbs : bonesData.vertexSpaceToBoneSpace)
```
McCallisterRomer: `std::ranges::for_each`

source/engine/graphics/system/SkeletalAnimationTypes.cpp
```diff
@@ -0,0 +1,68 @@
...
+    {
+        auto pos = std::ranges::find_if(bonesData.boneSpaceToParentSpace, [vsbs](const auto& bsps)
```
McCallisterRomer: If leaving like this, capture as &vsbs, but can't this just do like `std::ranges::find(c, vsbs.boneName)`?

include/ncengine/graphics/SkeletalAnimator.h
```diff
@@ -0,0 +1,50 @@
+#pragma once
+
+#include "ncengine/ecs/Component.h"
+#include "ncengine/graphics/SkeletalAnimationTypes.h"
+
+#include <memory>
```
McCallisterRomer: Not used?

include/ncengine/graphics/SkeletalAnimator.h
```diff
@@ -0,0 +1,50 @@
...
+
+namespace nc::graphics
+{
+class SkeletalAnimator : public ComponentBase
+{
+    public:
+        SkeletalAnimator(Entity entity, std::string assetUid);
+        SkeletalAnimator(SkeletalAnimator&& other) noexcept;
```
McCallisterRomer: I'm not seeing a reason move/d'tor need to do the 'default in the cpp' thing. Maybe this used to be a pimpl (hence the extra <memory> header)?

include/ncengine/graphics/SkeletalAnimator.h
```diff
@@ -0,0 +1,50 @@
...
+        SkeletalAnimator(SkeletalAnimator& other) = delete;
+        auto operator=(SkeletalAnimator&& other) noexcept -> SkeletalAnimator&;
+        auto operator=(SkeletalAnimator& other) -> SkeletalAnimator& = delete;
+        ~SkeletalAnimator() noexcept;
+
+        // Main
+        void Execute();
+
+        // Connect to system
+        auto Connect() -> Signal<const anim::PlayState&>& { return m_onPlayStateChanged; };
```
McCallisterRomer: ~const~ noexcept

include/ncengine/graphics/SkeletalAnimator.h
```diff
@@ -0,0 +1,50 @@
...
+
+        // Get UID of mesh that will be animated
+        auto MeshUid() -> const std::string& { return m_meshUid; }
```
McCallisterRomer: const noexcept

include/ncengine/graphics/SkeletalAnimator.h
```diff
@@ -0,0 +1,50 @@
...
+        bool m_initialStateInitialized;
+};
+} // namespace nc::graphics
+
+namespace nc
+{
+template<>
+struct StoragePolicy<graphics::SkeletalAnimator> : DefaultStoragePolicy
+{
+        static constexpr bool EnableOnAddCallbacks = true;
```
McCallisterRomer: unindnet by 1

source/engine/graphics/components/SkeletalAnimator.cpp
```diff
@@ -0,0 +1,71 @@
+#include "graphics/SkeletalAnimator.h"
+#include "graphics/MeshRenderer.h"
+#include "graphics/ToonRenderer.h"
+#include "ecs/Registry.h"
+
+namespace
+{
+auto GetMeshPath(nc::Entity entity) -> std::string
```
McCallisterRomer: Please don't use `ActiveRegistry`, its just another usage that'll have to be fixed before we remove it. Take the mesh path in the constructor.

source/engine/graphics/api/vulkan/shaders/SkeletalAnimationShaderResource.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include "graphics/api/vulkan/Initializers.h"
+#include "graphics/shader_resource/SkeletalAnimationData.h"
+
+namespace nc::graphics
+{
+constexpr uint32_t MaxAnimatedBones = 30000u;
+
+SkeletalAnimationShaderResource::SkeletalAnimationShaderResource(uint32_t bindingSlot, GpuAllocator* allocator, ShaderDescriptorSets* descriptors)
+    : m_bindingSlot{bindingSlot},
+        m_allocator{allocator},
```
McCallisterRomer: indentation

source/engine/graphics/api/vulkan/shaders/SkeletalAnimationShaderResource.cpp
```diff
@@ -0,0 +1,58 @@
+#include "SkeletalAnimationShaderResource.h"
+#include "graphics/api/vulkan/Initializers.h"
+#include "graphics/shader_resource/SkeletalAnimationData.h"
+
+namespace nc::graphics
+{
+constexpr uint32_t MaxAnimatedBones = 30000u;
```
McCallisterRomer: Should this be more publicly visible? Or configurable?

source/engine/graphics/shader_resource/ShaderResourceBus.h
```diff
@@ -15,5 +15,6 @@ struct ShaderResourceBus
     Signal<const EnvironmentData&> environmentChannel;
     Signal<const std::vector<ObjectData>&> objectChannel;
     Signal<const std::vector<PointLightData>&> pointLightChannel;
+    Signal<const std::vector<SkeletalAnimationData>&> skeletalAnimationChannel;
```
McCallisterRomer, line 19: Add forward declaration for `SkeletalAnimationData`. Getting away with something here.

source/engine/graphics/system/SkeletalAnimationCalculations.cpp
```diff
@@ -0,0 +1,276 @@
...
+        }
+        
+        auto animatedBoneOffset = anim::DecomposedMatrix
+        {
+            .pos =   GetInterpolatedPosition(timeInTicks, iter->second.positionFrames),
+            .rot =   GetInterpolatedRotation(timeInTicks, iter->second.rotationFrames),
+            .scale = GetInterpolatedScale(timeInTicks, iter->second.scaleFrames)
+        };
+
+        animationMatrices.offsets.push_back(std::move(animatedBoneOffset));
```
McCallisterRomer: There's no meaningful c'tor to select with a move for this type. You could emplace_back.

source/engine/graphics/system/SkeletalAnimationCalculations.cpp
```diff
@@ -0,0 +1,276 @@
...
+                     const nc::asset::SkeletalAnimation* animation) -> anim::PackedAnimation
+{
+    auto decomposedAnimation = GetAnimationOffsets(timeInTicks, boneNames, animation);
+
+    auto packedAnimation = anim::PackedAnimation{};
+    auto elementsCount = decomposedAnimation.offsets.size();
+    packedAnimation.offsets.reserve(elementsCount);
+    packedAnimation.hasValues = std::vector<anim::HasValue>{decomposedAnimation.hasValues.begin(), decomposedAnimation.hasValues.end()};
+
+    for (auto i = 0u; i < elementsCount; i++)
```
McCallisterRomer: Avoid raw for loops whenever possible. From the back of a napkin:
```
std::ranges::transform(decomposedAnimation.offsets, std::back_inserter(packedAnimation.offsets), [](auto&& offset)
{
    return ToTransMatrix(offset.pos) * ToRotMatrix(offset.rot) * ToScaleMatrix(offset.scale);
});
```

source/engine/graphics/system/SkeletalAnimationCalculations.h
```diff
@@ -0,0 +1,34 @@
...
+                            float blendToTime,
+                            float blendFactor,
+                            const std::vector<std::string>& boneNames,
+                            const nc::asset::SkeletalAnimation* blendFromAnim,
+                            const nc::asset::SkeletalAnimation* blendToAnim) -> anim::PackedAnimation;
+
+auto AnimateBones(const anim::PackedRig& rig,
+                  const anim::PackedAnimation& anim) -> std::vector<nc::graphics::SkeletalAnimationData>;
+
+auto GetInterpolatedPosition(float timeInTicks, const std::vector<nc::asset::PositionFrame> positionFrames) -> nc::Vector3;
```
McCallisterRomer: Missing & on all vector params

source/engine/graphics/system/SkeletalAnimationCalculations.cpp
```diff
@@ -0,0 +1,276 @@
...
+    {
+        throw NcError("Animation has no rotation data.");
+    }
+
+    if (rotationFrames.size() == 1)
+    {
+        return Normalize(rotationFrames[0].rotation);
+    }
+
+    auto rotationIndex = UINT_LEAST32_MAX;
```
McCallisterRomer: LEAST32 an autocomplete error?

source/engine/graphics/system/SkeletalAnimationCalculations.cpp
```diff
@@ -0,0 +1,276 @@
...
+    for (auto i = 0u; i < rotationFrames.size(); i++)
+    {
+        if (timeInTicks < rotationFrames[i+1].timeInTicks)
```
McCallisterRomer: i+1 goes out of bounds if we're on the last iteration. Maybe its impossible to arrive there, but I can't guarantee that just by looking. Maybe a `find_if` using `std::distance(beg, std::prev(pos))` to get the index? Probably requires a check to make sure we don't decrement from the beginning.

McCallisterRomer: Hmm, I already typed this comment, but it disappeared. Apologies if it duplicates.

i+1 goes out of bounds on the last iteration. Maybe use a `std::find_if` and something along the lines of `std::distance(beg, std::prev(pos))` to get the index. Probably requires a check to make sure we don't decrement from begin.

source/engine/graphics/system/SkeletalAnimationCalculations.cpp
```diff
@@ -0,0 +1,276 @@
...
+auto GetInterpolatedRotation(float timeInTicks, const std::vector<nc::asset::RotationFrame> rotationFrames) -> nc::Quaternion
+{
+    auto interpolatedRotation = nc::Quaternion{};
+
+    if (rotationFrames.size() == 0)
+    {
+        throw NcError("Animation has no rotation data.");
+    }
+
+    if (rotationFrames.size() == 1)
```
McCallisterRomer: Is this actually a special case that requires its own check, or would it be covered by the loop below? If the latter, I'd remove it for s p e e d.

source/engine/graphics/system/SkeletalAnimationCalculations.cpp
McCallisterRomer: Ok, I should have look a bit closer. Based on code below, it looks like it *shouldn't* read past the end, but it still makes me feel icky to have to potential looming around. I don't think you need the index at all; just dereference the iterator(s) instead of doing a subscript.

source/engine/graphics/system/SkeletalAnimationCalculations.cpp
```diff
@@ -0,0 +1,276 @@
...
+    {
+        if (timeInTicks < positionFrames[i+1].timeInTicks)
+        {
+            positionIndex = i;
+            break;
+        }
+    }
+    if (positionIndex == UINT_LEAST32_MAX)
+    {
+        throw NcError("Animation has no position data at time: ", std::to_string(timeInTicks));
```
McCallisterRomer: Use format strings for these:
```
NcError(fmt::format("... at time {}", timeInTicks));
```

source/engine/graphics/system/SkeletalAnimationCalculations.cpp
```diff
@@ -0,0 +1,276 @@
...
+    auto vertexToBoneSize = rig.vertexToBone.size();
+    for (auto i = 0u; i < vertexToBoneSize; i++)
+    {
+        animatedBones.push_back(nc::graphics::SkeletalAnimationData{rig.vertexToBone[i] * boneToParentSandbox[rig.offsetsMap[i]] * globalInverseTransform});
+    }
+
+    return animatedBones;
+}
+
+auto GetInterpolatedPosition(float timeInTicks, const std::vector<nc::asset::PositionFrame> positionFrames) -> nc::Vector3
```
McCallisterRomer: Same comment as Rotation

source/engine/graphics/system/SkeletalAnimationCalculations.cpp
```diff
@@ -0,0 +1,276 @@
...
+    if (interpolationFactor < 0.0f || interpolationFactor > 1.0f)
+    {
+        throw NcError("Error calculating the interpolation factor: ", std::to_string(interpolationFactor));
+    }
+
+    interpolatedRotation = Slerp(rotationFrames[rotationIndex].rotation, rotationFrames[nextRotationIndex].rotation, static_cast<float>(interpolationFactor));
+    return Normalize(interpolatedRotation);
+}
+
+auto GetInterpolatedScale(float timeInTicks, const std::vector<nc::asset::ScaleFrame> scaleFrames) -> nc::Vector3
```
McCallisterRomer: Same comments as Rotation

source/engine/graphics/system/SkeletalAnimationCalculations.cpp
```diff
@@ -0,0 +1,276 @@
...
+{
+    auto interpolatedScale = nc::Vector3{};
+    if (scaleFrames.size() == 0)
```
McCallisterRomer: Now that I think about it, what do you think about making all the error handling use NC_ASSERT just to compress things a bit. Its not really an assert, as it throws. However, it is something we can turn on and off. I suppose it comes down to whether these are more like sanity checks (prefer NC_ASSERT), or if they're reasonably easy to arrive at(keep as is). Can discuss, if needed.

include/ncengine/graphics/SkeletalAnimator.h
```diff
@@ -0,0 +1,50 @@
...
+{
+    public:
+        SkeletalAnimator(Entity entity, std::string assetUid);
+        SkeletalAnimator(SkeletalAnimator&& other) noexcept;
+        SkeletalAnimator(SkeletalAnimator& other) = delete;
+        auto operator=(SkeletalAnimator&& other) noexcept -> SkeletalAnimator&;
+        auto operator=(SkeletalAnimator& other) -> SkeletalAnimator& = delete;
+        ~SkeletalAnimator() noexcept;
+
+        // Main
```
McCallisterRomer: Use doxygen-style comments for public facing stuff. I'm ok if you want to exclude them for now.

include/ncengine/graphics/SkeletalAnimationTypes.h
```diff
@@ -0,0 +1,80 @@
+#pragma once
```
McCallisterRomer, line 5: Add file + copyright info

include/ncengine/graphics/SkeletalAnimator.h
```diff
@@ -0,0 +1,50 @@
+#pragma once
```
McCallisterRomer, line 5: Add file + copyright info

include/ncengine/NcFwd.h
```diff
@@ -49,6 +49,7 @@ class MeshRenderer;
 class ParticleEmitter;
 class PointLight;
 class SceneNavigationCamera;
+class SkeletalAnimator;
```
McCallisterRomer, line 52: GitHub extension won't let me comment on file - update copyright year.

McCallisterRomer (commented):
Haven't made it all the way through yet, just pushing out current comments.

source/engine/graphics/system/SkeletalAnimationSystem.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+namespace
+{
+auto PrepareAnimation(nc::graphics::anim::UnitOfWork& unit, const nc::graphics::anim::PackedRig& rig, float dt) -> nc::graphics::anim::PackedAnimation
+{
+    auto packedAnimation = nc::graphics::anim::PackedAnimation{};
+
+    if (unit.blendFromAnim && unit.blendToAnim)
+    {
+        unit.blendToTime =   fmod((unit.blendToTime + dt),   (static_cast<float>(unit.blendToAnim->durationInTicks)   / unit.blendToAnim->ticksPerSecond));
```
McCallisterRomer, line 16: I think all of the static_casts here are unnecessary.

source/engine/graphics/system/SkeletalAnimationSystem.cpp
```diff
@@ -0,0 +1,242 @@
...
+        unit.blendFromTime = fmod((unit.blendFromTime + dt), (static_cast<float>(unit.blendFromAnim->durationInTicks) / unit.blendFromAnim->ticksPerSecond));
+
+        auto blendFromTimeTicks = unit.blendFromTime * unit.blendFromAnim->ticksPerSecond;
+        auto blendToTimeTicks =   unit.blendToTime   * unit.blendToAnim->ticksPerSecond;
+
+        return nc::graphics::ComposeBlendedMatrices(blendFromTimeTicks, blendToTimeTicks, unit.blendFactor, rig.boneNames, unit.blendFromAnim, unit.blendToAnim);
+    }
+
+    float* timeInSeconds;
```
McCallisterRomer: Initialize these with an immediately invoked lamda.

source/engine/graphics/system/SkeletalAnimationSystem.cpp
```diff
@@ -0,0 +1,242 @@
...
+        timeInSeconds = &unit.blendToTime;
+        animData = unit.blendToAnim;
+    }
+
+    *timeInSeconds = fmod((*timeInSeconds + dt), (static_cast<float>(animData->durationInTicks)/animData->ticksPerSecond));
+    auto timeInTicks = *timeInSeconds * animData->ticksPerSecond;
+    return nc::graphics::ComposeMatrices(timeInTicks, rig.boneNames, animData);
+}
+
+auto HasCompletedAnimationCycle(const nc::graphics::anim::UnitOfWork unit, float dt) -> bool
```
McCallisterRomer, line 37: missing & on param

source/engine/graphics/system/SkeletalAnimationSystem.cpp
```diff
@@ -0,0 +1,242 @@
...
+    {
+        animator.Execute();
+    }
+
+    if (m_units.empty()) return SkeletalAnimationSystemState{};
+
+    auto stateIndex = 0u;
+    auto buffer = std::vector<SkeletalAnimationData>{};
+    auto state = SkeletalAnimationSystemState{};
+    auto dt = time::DeltaTime();
```
McCallisterRomer: const

source/engine/graphics/system/SkeletalAnimationSystem.cpp
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    *timeInSeconds = fmod((*timeInSeconds + dt), (static_cast<float>(animData->durationInTicks)/animData->ticksPerSecond));
+    auto timeInTicks = *timeInSeconds * animData->ticksPerSecond;
+    return nc::graphics::ComposeMatrices(timeInTicks, rig.boneNames, animData);
+}
+
+auto HasCompletedAnimationCycle(const nc::graphics::anim::UnitOfWork unit, float dt) -> bool
+{
+    return (unit.blendToAnim && (unit.blendToTime + dt > static_cast<float>(unit.blendToAnim->durationInTicks)/unit.blendToAnim->ticksPerSecond));
```
McCallisterRomer: unnecessary outer parens and static_cast

source/engine/graphics/system/SkeletalAnimationSystem.cpp
```diff
@@ -0,0 +1,242 @@
...
+}
+}
```
McCallisterRomer, line 40: // anonymous namespace

source/engine/graphics/system/SkeletalAnimationSystem.cpp
```diff
@@ -0,0 +1,242 @@
...
+    return state;
+}
+
+void SkeletalAnimationSystem::UpdateSkeletalAnimationStorage(const asset::SkeletalAnimationUpdateEventData& eventData)
+{
+    switch (eventData.updateAction)
+    {
+        case asset::UpdateAction::Load:
+        {
+            for (auto i = 0u; i < eventData.ids.size(); i++)
```
McCallisterRomer: Hopefully `std::views::zip`

McCallisterRomer: Hopefully `std::views::zip`

source/engine/graphics/system/SkeletalAnimationSystem.cpp
```diff
@@ -0,0 +1,242 @@
...
+            {
+                m_animationAssets.emplace(eventData.ids[i], eventData.data[i]);
+            }
+            break;
+        }
+        case asset::UpdateAction::Unload:
+        {
+            for (auto i = 0u; i < eventData.ids.size(); i++)
```
McCallisterRomer: `std::ranges::for_each(eventData.ids, [this](auto&& id) { m_animationAssets.erase(id); });`

source/engine/graphics/system/SkeletalAnimationSystem.cpp
```diff
@@ -0,0 +1,242 @@
...
+    }
+}
+
+void SkeletalAnimationSystem::UpdateBonesStorage(const asset::BoneUpdateEventData& eventData)
+{
+    switch (eventData.updateAction)
+    {
+        case asset::UpdateAction::Load:
+        {
+            for (auto i = 0u; i < eventData.ids.size(); i++)
```
McCallisterRomer: same

source/engine/graphics/system/SkeletalAnimationSystem.cpp
```diff
@@ -0,0 +1,242 @@
...
+            {
+                m_rigs.emplace(eventData.ids[i], eventData.data[i]); /* @todo Need to move here and not have span in event data? To avoid copying? */
+            }
+            break;
+        }
+        case asset::UpdateAction::Unload:
+        {
+            for (auto i = 0u; i < eventData.ids.size(); i++)
```
McCallisterRomer: same

source/engine/graphics/system/SkeletalAnimationSystem.cpp
```diff
@@ -0,0 +1,242 @@
...
+void SkeletalAnimationSystem::Start(const anim::PlayState& playState)
+{
+    if (m_rigs.find(playState.meshUid) == m_rigs.end() ||
+        m_animationAssets.find(playState.curAnimUid) == m_animationAssets.end() ||
+        m_animationAssets.find(playState.prevAnimUid) == m_animationAssets.end())
+    {
+        return;
+    }
+
+    auto pos = std::find_if(m_unitEntities.begin(), m_unitEntities.end(), [playState](Entity entity){ return playState.entity.Index() == entity.Index(); });
```
McCallisterRomer: `std::ranges::find(m_unitEntities, playState.entity)`

source/engine/graphics/system/SkeletalAnimationSystem.cpp
```diff
@@ -0,0 +1,242 @@
...
+            m_unitEntities.clear();
+            m_rigs.clear();
+            break;
+        }
+    }
+}
+
+void SkeletalAnimationSystem::Start(const anim::PlayState& playState)
+{
+    if (m_rigs.find(playState.meshUid) == m_rigs.end() ||
```
McCallisterRomer: `contains()` would be a bit cleaner than `find()`

source/engine/graphics/system/SkeletalAnimationSystem.cpp
```diff
@@ -0,0 +1,242 @@
...
+    {
+        return;
+    }
+
+    auto pos = std::find_if(m_unitEntities.begin(), m_unitEntities.end(), [playState](Entity entity){ return playState.entity.Index() == entity.Index(); });
+
+    if (pos == m_unitEntities.end())
+    {
+        m_unitEntities.push_back(playState.entity);
+        m_units.push_back(anim::UnitOfWork
```
McCallisterRomer: `emplace_back()`

source/engine/graphics/system/SkeletalAnimationSystem.cpp
```diff
@@ -0,0 +1,242 @@
...
+        {
+            .index         = playState.entity.Index(),
+            .rig           = &m_rigs.at(playState.meshUid),
+            .blendFromAnim = playState.prevAnimUid == std::string{} ? nullptr : &m_animationAssets.at(playState.prevAnimUid),
```
McCallisterRomer: `x.empty()` here and below

source/engine/graphics/system/SkeletalAnimationSystem.cpp
```diff
@@ -0,0 +1,242 @@
...
+        return;
+    }
+
+    auto posIndex = std::distance(m_unitEntities.begin(), pos);
+    m_unitEntities.at(posIndex) = playState.entity;
+    m_units.at(posIndex) = anim::UnitOfWork
+    {
+        .index         = playState.entity.Index(),
+        .rig           = &m_rigs.at(playState.meshUid),
+        .blendFromAnim = playState.prevAnimUid == std::string{} ? nullptr : &m_animationAssets.at(playState.prevAnimUid),
```
McCallisterRomer: `empty()` again

source/engine/graphics/system/SkeletalAnimationSystem.h
```diff
@@ -0,0 +1,62 @@
...
+
+namespace nc::graphics
+{
+struct SkeletalAnimationSystemState
+{
+    // Index consumed by shader that signifies first animation matrix for the given animation
+    std::unordered_map<Entity::index_type, uint32_t> animationIndices;
+};
+
+class SkeletalAnimationSystem
```
McCallisterRomer: Since the this pointer is captured by signals, I'd suggest deriving from `StableAddress` to catch any future errors.

source/engine/graphics/system/SkeletalAnimationSystem.cpp
```diff
@@ -0,0 +1,242 @@
...
+{
+    auto& signal = animator.Connect();
+    auto connectionState = std::make_unique<Connection<const anim::PlayState&>>(signal.Connect(this, &SkeletalAnimationSystem::OnPlayStateChanged));
+    m_onPlayStateChangedHandlers.emplace_back(std::move(connectionState));
+    m_handlerIndices.emplace_back(animator.ParentEntity().Index());
+}
+
+void SkeletalAnimationSystem::Remove(Entity entity)
+{
+    auto pos = std::find(m_handlerIndices.begin(), m_handlerIndices.end(), entity.Index());
```
McCallisterRomer: consider ranges

source/engine/graphics/system/SkeletalAnimationSystem.cpp
```diff
@@ -0,0 +1,242 @@
...
+    if(pos == m_handlerIndices.end())
+    {
+        return;
+    }
+
+    auto index = std::distance(m_handlerIndices.begin(), pos);
```
McCallisterRomer: const

source/engine/graphics/system/SkeletalAnimationTypes.cpp
jaremieromer: Muscle memory from python days, probably

test/skeletal_animation/SkeletalAnimationCalculations_tests.cpp
```diff
@@ -0,0 +1,520 @@
...
+
+        void TearDown() override
+        {
+            testData = nullptr;
+        }
+};
+
+bool MatrixEquals(const DirectX::XMMATRIX& a, const DirectX::XMMATRIX& b)
+{
+    DirectX::XMFLOAT4X4 aView;
```
McCallisterRomer: Once again annoyed there is no matrix equal func, but can we do something like:
```
return XMVector4Equal(a.r[0], b.r[0]) &&
       XMVector4Equal(a.r[1], b.r[1]) &&
       XMVector4Equal(a.r[2], b.r[2]) &&
       XMVector4Equal(a.r[3], b.r[3]);
```

source/engine/graphics/system/SkeletalAnimationCalculations.cpp
jaremieromer: It is it's own case - we don't want to interpolate if we only have one frame

source/engine/graphics/system/SkeletalAnimationSystem.cpp
jaremieromer: MSVC doesn't mind if I omit it, but GCC get's upset with the -Werror=conversion error.

source/engine/graphics/system/SkeletalAnimationSystem.cpp
jaremieromer: I need the static cast for GCC, see above comment

source/engine/graphics/system/SkeletalAnimationCalculations.cpp
```diff
@@ -0,0 +1,176 @@
...
+    {
+        auto iter = animation->framesPerBone.find(boneName);
+        if (iter == animation->framesPerBone.end())
+        {
+           animationMatrices.offsets.push_back(anim::DecomposedMatrix{});
+           animationMatrices.hasValues.push_back(0);
+           continue;
+        }
+        
+        animationMatrices.offsets.emplace_back(anim::DecomposedMatrix
```
McCallisterRomer: with `emplace_back` I mean to remove the extra object construction:

```cpp
animationMatrices.offset.emplace_back
(
    GetInterpolatedPosition(timeInTicks, iter->second.positionFrames),
    GetInterpolatedRotation(timeInTicks, iter->second.rotationFrames),
    GetInterpolatedScale(timeInTicks, iter->second.scaleFrames)
);
```

source/engine/graphics/system/SkeletalAnimationCalculations.cpp
```diff
@@ -0,0 +1,176 @@
...
+auto ComposeMatrices(float timeInTicks,
+                     const std::vector<std::string>& boneNames,
+                     const nc::asset::SkeletalAnimation* animation) -> anim::PackedAnimation
+{
+    auto decomposedAnimation = GetAnimationOffsets(timeInTicks, boneNames, animation);
+
+    auto packedAnimation = anim::PackedAnimation{};
+    auto elementsCount = decomposedAnimation.offsets.size();
+    packedAnimation.offsets.reserve(elementsCount);
+    packedAnimation.hasValues = std::vector<anim::HasValue>{decomposedAnimation.hasValues.begin(), decomposedAnimation.hasValues.end()};
```
McCallisterRomer: A little hard to see as it is, but `decomposedAnimation` isn't used anywhere else. You should just move it here, rather than copying it. I'm always worried about 'non-obvious moved-from things' getting overlooked in the future. Maybe it would be easier to spot if it was moved right away, and `decomposedAnimation` was replaced with a structured binding? Idk.

```cpp
auto [offsets, hasValues] = GetAnimationOffsets(...);
auto packedAnimation = anim::PackedAnimation{.hasValues = std::move(hasValues)};
```

source/engine/graphics/system/SkeletalAnimationCalculations.cpp
```diff
@@ -0,0 +1,176 @@
...
+                            const std::vector<std::string>& boneNames,
+                            const nc::asset::SkeletalAnimation* blendFromAnim,
+                            const nc::asset::SkeletalAnimation* blendToAnim) -> anim::PackedAnimation
+{
+    auto blendFromDecomposed = GetAnimationOffsets(blendFromTime, boneNames, blendFromAnim);
+    auto blendToDecomposed = GetAnimationOffsets(blendToTime, boneNames, blendToAnim);
+
+    auto packedAnimation = anim::PackedAnimation{};
+    packedAnimation.hasValues = std::vector<anim::HasValue>{blendFromDecomposed.hasValues.begin(), blendFromDecomposed.hasValues.end()};
+
```
McCallisterRomer, line 74: Same thing with moving hasValues here.

source/engine/graphics/system/SkeletalAnimationCalculations.cpp
```diff
@@ -0,0 +1,176 @@
...
+auto AnimateBones(const anim::PackedRig& rig,
+                  const anim::PackedAnimation& anim) -> std::vector<nc::graphics::SkeletalAnimationData>
+{
+    // Copy the boneToParent vector to perform modifications in place.
+    auto boneToParentSandbox = std::vector<DirectX::XMMATRIX>{rig.boneToParent.begin(), rig.boneToParent.end()};
+    auto boneToParentSandboxSize = boneToParentSandbox.size();
+
+    // Replace each boneToParent offset with its animation offset, if present. Else, leave as the original offset.
+    for (auto i = 0u; i < boneToParentSandboxSize; i++)
```
McCallisterRomer: Try to continue removing the raw loops:
```cpp
auto boneToParent = std::vector<DirectX::XMMATRIX>{};
boneToParent.reserve(rig.boneToParent.size());
std::ranges::transform(
    std::views::zip(rig.boneToParent, anim.offsets, anim.hasValues),
    std::back_inserter(boneToParent),
    [](auto&& in)
    {
        auto&& [parent, offset, hasValue] = in;
        return parent * !hasValue + offset * hasValue;
    }
);
```

source/engine/graphics/system/SkeletalAnimationCalculations.cpp
```diff
@@ -0,0 +1,176 @@
...
+    {
+        boneToParentSandbox[i] = (boneToParentSandbox[i] * !anim.hasValues[i]) + (anim.offsets[i] * anim.hasValues[i]);
+    }
+
+    // Multiply each child (siblings are contiguous) with its parent.
+    for (auto i = 0u; i < boneToParentSandboxSize; i++)
```
McCallisterRomer:
```cpp
std::ranges::for_each(
    std::views::zip(rig.boneToParentIndices, boneToParent),
    [&boneToParent](auto&& in)
    {
        auto&& [indices, parent] = in;
        auto&& [first, count] = indices; // prob not necessary with custom struct
        std::ranges::for_each(
            std::views::counted(boneToParent.begin() + first, count),
            [&parent](auto& child){ parent *= child; }
        );
    }
);
```

source/engine/graphics/system/SkeletalAnimationCalculations.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+    // Create the output vector.
+    auto animatedBones = std::vector<nc::graphics::SkeletalAnimationData>{};
+    animatedBones.reserve(rig.vertexToBone.size());
+
+    // Create a final transform for each bone by multiplying the (vertex-space-to-bone-space matrix) with the (bone-space-to-animated-parent-bone-space matrix) with the (global inverse transform matrix).
+    // This outputs a matrix that can be used to transform a vertex into its final animated position.
+    auto globalInverseTransform = rig.globalInverseTransform;
+    auto vertexToBoneSize = rig.vertexToBone.size();
+    for (auto i = 0u; i < vertexToBoneSize; i++)
```
McCallisterRomer:
```cpp
std::ranges::transform(
    std::views::zip(rig.vertexToBone, rig.offsetsMap),
    std::back_inserter(animatedBones),
    [globalInverseTransform = rig.globalInverseTransform, &boneToParent](auto&& in)
    {
        auto&& [matrix, offset] = in;
        return SkeletalAnimationData{matrix * boneToParent.at(offset) * globalInverseTransform};
    }
);
```

source/engine/graphics/system/SkeletalAnimationCalculations.cpp
```diff
@@ -129,144 +123,66 @@ auto AnimateBones(const anim::PackedRig& rig,
...
 {
-    auto interpolatedPosition = nc::Vector3{};
+    NC_ASSERT(positionFrames.size() > 0, "Animation has no position data for the node.");
+    if (positionFrames.size() == 1) return positionFrames[0].position;
 
-    if (positionFrames.size() == 0)
-    {
-        throw NcError("Animation has no position data for the node.");
-    }
+    auto nextFramePos = std::find_if(positionFrames.begin(), positionFrames.end(), [timeInTicks](auto&& frame){ return timeInTicks < frame.timeInTicks;});
```
McCallisterRomer: Since we already know the container has at least 2 elements, and we know we don't want the first one, you can omit the first element from the loop. Then you can also remove the != begin assertion:
```cpp
const auto nextFrame = std::ranges::find_if(
    std::views::drop(positionFrames, 1),
    [timeInTicks](auto&& frame){ return timeInTicks < frame.timeInTicks;}
);

NC_ASSERT(nextFrame != positionFrames.end(), "...");
```

Also, prefer to put failure checks as early as possible. The way its written, there's potentially undefined behavior calling `std::prev` before checking if `nextFramePos != positionFrames.begin()`.

McCallisterRomer: Also, every variable here can be const.

source/engine/graphics/system/SkeletalAnimationCalculations.cpp
```diff
@@ -129,144 +123,66 @@ auto AnimateBones(const anim::PackedRig& rig,
...
+    NC_ASSERT(interpolationFactor >= 0.0f && interpolationFactor <= 1.0f, fmt::format("Error calculating the interpolation factor: {}", interpolationFactor));
 
-    if (interpolationFactor < 0.0f || interpolationFactor > 1.0f)
-    {
-        throw NcError("Error calculating the interpolation factor: ", std::to_string(interpolationFactor));
-    }
-
-    interpolatedPosition = Lerp(positionFrames[positionIndex].position, positionFrames[nextPositionIndex].position, static_cast<float>(interpolationFactor));
-    return interpolatedPosition;
+    return Lerp(positionFrames[positionIndex].position, positionFrames[nextPositionIndex].position, static_cast<float>(interpolationFactor));
```
McCallisterRomer: Can remove static_cast

source/engine/graphics/system/SkeletalAnimationCalculations.cpp
```diff
@@ -129,144 +123,66 @@ auto AnimateBones(const anim::PackedRig& rig,
...
 }
 
-auto GetInterpolatedRotation(float timeInTicks, const std::vector<nc::asset::RotationFrame> rotationFrames) -> nc::Quaternion
+auto GetInterpolatedRotation(float timeInTicks, const std::vector<nc::asset::RotationFrame>& rotationFrames) -> nc::Quaternion
```
McCallisterRomer, line 145: Same comments from `GetInterpolatedPosition` apply here and in `GetInterpolatedRotation`.

source/engine/graphics/system/SkeletalAnimationSystem.cpp
```diff
@@ -22,30 +22,18 @@ auto PrepareAnimation(nc::graphics::anim::UnitOfWork& unit, const nc::graphics::
...
-    }
-    else
-    {
-        timeInSeconds = &unit.blendToTime;
-        animData = unit.blendToAnim;
-    }
-
+    float* timeInSeconds = unit.blendFromAnim ? &unit.blendFromTime : &unit.blendToTime;
+    nc::asset::SkeletalAnimation* animData = unit.blendFromAnim ? unit.blendFromAnim : unit.blendToAnim;
     *timeInSeconds = fmod((*timeInSeconds + dt), (static_cast<float>(animData->durationInTicks)/animData->ticksPerSecond));
```
McCallisterRomer, line 32: Ahh, sorry about previous static_cast comments. My F12 was taking me to the wrong type definition.

source/engine/graphics/system/SkeletalAnimationSystem.cpp
```diff
@@ -22,30 +22,18 @@ auto PrepareAnimation(nc::graphics::anim::UnitOfWork& unit, const nc::graphics::
...
-        timeInSeconds = &unit.blendFromTime;
-        animData = unit.blendFromAnim;
-    }
-    else
-    {
-        timeInSeconds = &unit.blendToTime;
-        animData = unit.blendToAnim;
-    }
-
+    float* timeInSeconds = unit.blendFromAnim ? &unit.blendFromTime : &unit.blendToTime;
```
McCallisterRomer: I know its a nit, but I suggested an immediately invoked lambda expression because it has evolved from convention to [guideline](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es28-use-lambdas-for-complex-initialization-especially-of-const-variables). The benefits are greater for const things, but its a good habit to internalize:

```cpp
auto [timeInSeconds, animData] = [&unit]()
{
    return unit.blendFromAnim ?
        std::pair{&unit.blendFromTime, unit.blendFromAnim} :
        std::pair{&unit.blendToTime, unit.blendToAnim};
}();
```

source/engine/graphics/system/SkeletalAnimationSystem.cpp
```diff
@@ -167,24 +149,23 @@ void SkeletalAnimationSystem::UpdateBonesStorage(const asset::BoneUpdateEventDat
...
 
-    auto pos = std::find_if(m_unitEntities.begin(), m_unitEntities.end(), [playState](Entity entity){ return playState.entity.Index() == entity.Index(); });
-
+    auto pos = std::ranges::find(m_unitEntities, playState.entity);
     if (pos == m_unitEntities.end())
     {
-        m_unitEntities.push_back(playState.entity);
-        m_units.push_back(anim::UnitOfWork
+        m_unitEntities.emplace_back(playState.entity);
+        m_units.emplace_back(anim::UnitOfWork
```
McCallisterRomer: `emplace_back()` is used for in-place construction by forwarding arguments to the constructor. If you construct a `UnitOfWork` at the call site here, the in-place construction happens via a move constructor (aka two constructors run instead of one):
```cpp
m_units.emplace_back
(
    playState.entity.Index(),
    &m_rigs.at(playState.meshUid),
    playState.prevAnimUid.empty() ? nullptr : &m_animationAssets.at(playState.prevAnimUid),
    playState.curAnimUid.empty() ? nullptr : &m_animationAssets.at(playState.curAnimUid),
    0.0f,
    0.0f,
    0.0f
);
```

McCallisterRomer (approved):

source/engine/graphics/system/SkeletalAnimationSystem.cpp
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    if (m_units.empty()) return SkeletalAnimationSystemState{};
+
+    auto stateIndex = 0u;
+    auto buffer = std::vector<SkeletalAnimationData>{};
+    auto state = SkeletalAnimationSystemState{};
+    const auto dt = time::DeltaTime();
+
+    for (auto i = 0u; i < m_units.size(); i++)
```
McCallisterRomer: Can use `std::views::zip` here too.

PR #498, opened 2024-01-03
Component Ids

McCallisterRomer:
part of #420

Another piece needed for serialization. When deserializing component data, we need a way to map the individual blobs back to something that knows what to do with them (pool/handler). Adding an integer id that can be either automatically assigned or manually overridden.

All engine types have explicitly set ids. I added these as static members to each type, which isn't necessary, but it seems like a potentially useful pattern for templated code (for example, see RegistryFactories.cpp in this changeset).

Added a few utility functions to `ComponentRegistry` and `AnyComponent` for future use of the id.

Unrelated to the id, but I also changed the order in which components are registered. This influences the order of component widgets in the editor (would be nice to be able to change that, but that's a whole thing...).


include/ncengine/audio/AudioSource.h
```diff
@@ -28,6 +32,8 @@ class AudioSource : public ComponentBase
     NC_ENABLE_IN_EDITOR(AudioSource)
 
     public:
+        static constexpr auto ComponentId = 14ull;
```
jaremieromer: Optional, but I'd rather see these isolated to a single header, something like: 
`ComponentIds.h`
```cpp
static constexpr auto AudioSourceId = 14ull;
static constexpr auto MeshRendererId = 15ull;
...
```

include/ncengine/ecs/AnyComponent.h
```diff
@@ -43,13 +43,23 @@ class AnyComponent
             return !(*this == other);
         }
 
+        /**
+         * @brief Get the value of Componenthandler::id for the concrecte component type.
```
jaremieromer: typo in "concrecte"

include/ncengine/ecs/Component.h
```diff
@@ -93,6 +103,14 @@ struct ComponentHandler
     /** @brief Function type for the DrawUI handler. */
     using DrawUI_t = std::function<void(T&)>;
 
+    /**
+     * @brief A unique identifier for the component.
+     * @note Set to 0 to get assigned an available id. The range [EngineIdRangeBegin, EngineIdRangeEnd]
+     *       is reserved for engine components. Automatically assigned ids are assigned in reverse order
+     *       starting from std::numeric_limites<size_t>::max().
```
jaremieromer: typo in std::numeric_limites

jaremieromer (approved):

PR #499, opened 2024-01-04
Asset + Entity serialization

### LINKED ISSUE #432: Inline Transform member funcs
Some `Transform` functions are very inline friendly and don't expose anything internal/extra. Do a pass over the member funcs, moving anything small and frequently used (like `LocalPosition()`, `PositionXM()`, etc.) over to the header.

McCallisterRomer:
part of #420
resolves #432 (unrelated to serialization, but this has been bugging me nearly every time I write a unit test, and I had to hit some missing const/noexcept anyway)

Adding partial code for scene serialization - includes assets and entities, but not components or module state. Nothing is hooked up to the editor yet.

Serialization allows targeting a subset of game state, referred to as a scene fragment. Loading a fragment overlays the state on top of the current scene. This still allows saving the entirety of a scene, but with the following benefits:
- In the future, prefabs can be implemented as fragments. A prefab is really just a miniature scene; maybe its all nested under a top-level entity that gets positioned somewhere, but everything under that is just assets, entities, and components.
- Easier handling of things like one-time init logic and network spawned entities. For example, all persistent entities or those with a particular layer/component can be skipped so that they're always handled by hardcoded logic.


include/ncengine/utility/MatrixUtilities.h
```diff
@@ -0,0 +1,44 @@
+/**
```
jaremieromer, line 1: This will be nice to have

source/engine/serialize/SceneSerialization.cpp
```diff
@@ -0,0 +1,60 @@
...
+                       asset::NcAsset& assetModule)
+{
+    auto header = SceneFragmentHeader{};
+    serialize::Deserialize(stream, header);
+    if (header.magicNumber != g_sceneFragmentMagicNumber)
+        throw NcError("Unexpected SceneFragment header");
+
+    if (header.version != g_currentSceneFragmentVersion)
+        throw NcError("Unexpected SceneFragment version");
+
```
jaremieromer, line 46: This struck me as a little wonky here with the lack of braces on the if clauses and the temporary scopes here. I recommend bracing the if clauses here just for readability

test/serialize/SceneSerialization_unit_tests.cpp
```diff
@@ -0,0 +1,247 @@
...
+            .tag = "last",
+            .layer = nc::Entity::layer_type{2},
+            .flags = nc::Entity::Flags::Static
+        },
+        {.flags = nc::Entity::Flags::NoCollisionNotifications}
+    };
+
+    const auto unexpectedInfos = std::vector<nc::EntityInfo>
+    {
+        {.tag = "miss me", .flags = nc::Entity::Flags::Persistent},
```
jaremieromer, line 169: lol

jaremieromer (approved):

PR #503, opened 2024-01-07
Component Serialization

### LINKED ISSUE #501: Component serialization

McCallisterRomer:
resolves #501

Adding component serialization to scene fragment load/store. Engine components still need their (de)serialize functions written and tested, but I'll do that in a separate PR. Code paths are covered by tests here, and I had success manually testing saving and loading `MeshRenderer`s.

 Added `Insert()` to `ComponentPool` and renamed `Add()` to `Emplace()` to be consistent with `Insert()` and the naming in `EcsInterface`.

MinGW has had enough with our template shenanigans and refused to assemble `RegistryFactories.cpp`. It was exceeding a limit on the number of sections in the obj file, which there is a flag to increase, but I think there may be a file size limit as well. I'm kicking the can down the road a bit by just splitting up the file.


include/ncengine/ecs/Component.h
```diff
@@ -87,13 +88,22 @@ struct DefaultStoragePolicy
...
+
 /** @brief Optional data and callbacks for generic component operations. */
 template<PooledComponent T>
 struct ComponentHandler
 {
     /** @brief Function type for the Factory handler. */
     using Factory_t = std::function<T(Entity, void*)>;
 
+    /** @brief Function type for the serialize handler. */
+    using Serialize_t = std::function<void(std::ostream&, const T&, const SerializationContext&, void*)>;
```
jaremieromer: Nice to have param tags in the comment for consistency here.

include/ncengine/ecs/Component.h
```diff
@@ -87,13 +88,22 @@ struct DefaultStoragePolicy
...
+
+    /** @brief Function type for the deserialize handler. */
```
jaremieromer, line 107: Nice to have param tags in the comment for consistency here.

include/ncengine/ecs/ComponentPool.h
```diff
@@ -53,24 +56,30 @@ class ComponentPoolBase : public StableAddress
...
+        virtual auto HasDrawUI() const noexcept -> bool = 0;
 
         /**
          * @brief Factory-construct a component attached to an entity.
          * @note Returns a null AnyComponent if there is no registered factory handler.
          */
         virtual auto AddDefault(Entity entity) -> AnyComponent = 0;
 
+        /** @brief Serialize the component attached to an entity to a binary stream. */
+        virtual void Serialize(std::ostream& stream, Entity entity, const SerializationContext& ctx) = 0;
```
jaremieromer, line 78: Add param tags in those that were added

include/ncengine/ecs/ComponentPool.h
```diff
@@ -101,7 +110,10 @@ class ComponentPool final : public ComponentPoolBase
 
         /** @brief Emplace a component attached to an entity. */
         template<class... Args>
-        auto Add(Entity entity, Args&&... args) -> T*;
+        auto Emplace(Entity entity, Args&&... args) -> T*;
+
+        /** @brief Insert a component attached to an entity. */
```
jaremieromer, line 115: I am not a big fan of using "Insert" here. Is its difference from `Emplace` just that you pass in an already existing component rather than the args to create one? If so, why not use `PushBack`, or I'd even prefer `Emplace`.  If you can't because `Emplace`'s parameter pack causes overload resolution to fail, I would choose a different name than insert.

Insert, to me and from what I've seen in the std, is different than emplace/push_back specifically because it requires a position in which the object is inserted into the container, and this method doesn't satisfy that convention.

source/engine/serialize/SceneSerialization.cpp
```diff
@@ -8,57 +8,175 @@
...
-        serialize::Deserialize(stream, assets);
-        assetModule.LoadAssets(assets);
+        nc::RemapEntity(entityData.info.parent, ctx.entityMap);
+        ctx.entityMap.emplace(entityData.fragmentId, ctx.ecs.Emplace<nc::Entity>(entityData.info));
+    });
+}
+
+auto GetSerializablePools(nc::ecs::Ecs ecs)
+{
+    auto filter = std::views::filter(ecs.GetComponentPools(), [](auto pool){ return pool->HasSerialize(); });
```
jaremieromer, line 69: Hooray for views! This is neat

source/engine/serialize/SceneSerialization.cpp
```diff
@@ -8,57 +8,175 @@
...
     {
-        auto assets = asset::AssetMap{};
-        serialize::Deserialize(stream, assets);
-        assetModule.LoadAssets(assets);
+        nc::RemapEntity(entityData.info.parent, ctx.entityMap);
+        ctx.entityMap.emplace(entityData.fragmentId, ctx.ecs.Emplace<nc::Entity>(entityData.info));
+    });
+}
+
+auto GetSerializablePools(nc::ecs::Ecs ecs)
```
jaremieromer: We should lay down guidelines for when we want to use the automatic return type vs explicit return type. 
My preference would be: prefer explicit return type as it acts as documentation for the code and enhances readability - unless the return type is some garbled template mishmash and then we use auto.

jaremieromer: I don't think we should go fast and loose with automatic return types because IMO it does decrease readability.

jaremieromer (approved):

source/engine/serialize/SceneSerialization.cpp
McCallisterRomer: I'm still waiting for `std::ranges::to` to get implemented as the cherry on top:
```cpp
return std::views::filter(...) | std::ranges::to<std::vector>();
```

source/engine/serialize/SceneSerialization.cpp
McCallisterRomer: Good catch! I agree with your statements, and that's what we've typically been doing it. This one was a 'garbled template mishmash' (originally returned a view from somewhere) but didn't get fixed.

include/ncengine/ecs/ComponentPool.h
McCallisterRomer: Hmm, interesting. I think there is a little more nuance here. The sequence containers, like `vector`/`deque`/`list`/`string`(I'm including here because of its API, even though it isn't technically this), have `insert`/`emplace` overloads that always take a position because they have something like `push_back`/`emplace_back` (lists are weird though, but who uses 'em?). The (unordered) associative containers, like `map`/`unordered_map`/`set`, all have `insert`/`emplace` that take a `value_type` only. They may also have overloads that also take a pos that acts as a hint.

I'm a little torn here, because this container acts as both, but I'm thinking `PushBack`/`EmplaceBack` convey the wrong message due to the item being moved to a staging pool initially, not the actual back of the pool. Can discuss further.

include/ncengine/ecs/ComponentPool.h
McCallisterRomer: We should discuss doc strategy. I've been intentionally keeping docs to just a brief, unless there is some actually some meaningful information that isn't covered by the brief or code. For example, all documentation in this file is just brief with the occasional note. To me, parameters can generally be made self evident by the function name and/or brief. We should land on something consistent though.

include/ncengine/ecs/ComponentPool.h
McCallisterRomer: Just wanted to add that I really like this comment. It kept me thinking while shoveling. If we had such a thing, I'd send you some corporate recognition coins.

include/ncengine/ecs/Component.h
McCallisterRomer: Yeah, the `void*` in particular is probably a little mysterious.

include/ncengine/ecs/ComponentPool.h
jaremieromer: Sounds good - I generally agree that param docs are unnecessary

include/ncengine/ecs/ComponentPool.h
jaremieromer: I'll buy a Nightcrawler Tee with my imaginary coins!
Agreed that since it is also partially an associative container that Insert works.

PR #505, opened 2024-01-07
Skeletal Animation Final PR - State Machine

jaremieromer:
Note: Almost all of the files changed here are assets. Relevant files are `SkeletalAnimator` and `SkeletalAnimationTypes`

### Goal
Finalize the skeletal animation feature by adding and integrating the state machine into `SkeletalAnimator.h`

### Changes
 - Added all possible `Action` types to `SkeletalAnimationTypes.h`
 - Added the `State` struct and all of it's helper structs (`Stop`, `PlayOnce`, etc) to `SkeletalAnimationTypes.h`
 - Added a `StableSet` to `SkeletalAnimationTypes.h`
 - Added and implemented the state machine interface into `SkeletalAnimator`
 - Created a demo scene and added assets.


include/ncengine/graphics/SkeletalAnimationTypes.h
```diff
@@ -32,29 +34,186 @@ struct PlayState
...
+};
+
+/**
+ * @brief A type used to simplify creating SkeletalAnimator state machine states.
+ * 
+ * The Loop state is used to loop an animaton until the exitWhen condition is met.
+ */
+struct Loop
+{
+    uint32_t enterFrom;
```
McCallisterRomer, line 57: Check your padding bytes on these structs. At least `Loop` and `PlayOnce` can be made smaller by rearranging members.

include/ncengine/graphics/SkeletalAnimationTypes.h
```diff
@@ -32,29 +34,186 @@ struct PlayState
...
 struct State
 {
     static constexpr uint32_t NullId = UINT32_MAX;
-    State(const Initial& initialProperties)
+
+    /**
+     * @brief Construct a new State in the state machine representing the state the SkeletalAnimator is initialized playing.
+     * @param initialProperties The properties of the root state.
+     */
+    State(const Root& initialProperties)
```
McCallisterRomer: I think you want to be taking all of these properties by value since you're moving the members in.

include/ncengine/graphics/SkeletalAnimationTypes.h
```diff
@@ -32,29 +34,186 @@ struct PlayState
...
         : id{},
           action{Action::Loop},
-          animUid{std::move(initialProperties.animUid)}{}
+          animUid{std::move(initialProperties.animUid)},
```
McCallisterRomer: Just ideating, but if we had our own custom move functions, I think we could catch move from const at compile time.

include/ncengine/graphics/SkeletalAnimationTypes.h
```diff
@@ -32,29 +34,186 @@ struct PlayState
...
+          firstRunComplete{nullptr},
+          successors{},
+          enterFrom{State::NullId},
+          enterWhen{[](){return false;}},
```
McCallisterRomer: Ignore if you wish, but you could have a default transition predicate as a static constexpr member.

include/ncengine/graphics/SkeletalAnimationTypes.h
```diff
@@ -32,29 +34,186 @@ struct PlayState
...
+    /**
+     * @brief Construct a new State in the state machine representing adding an animation which plays once and then exits to the exit state.
+     * @param playOnceProperties Specifies from which state and from which condition you enter the play once state from, 
+     * which state you enter after the animation has completed one cycle, and which animation you play once.
+     */
+    State(const PlayOnce& playOnceProperties)
+        : id{},
+          action{Action::PlayOnce},
+          animUid{std::move(playOnceProperties.animUid)},
+          firstRunComplete{std::make_unique<bool>()},
```
McCallisterRomer: Its not strictly necessary, but consider explicitly initializing to false.

include/ncengine/graphics/SkeletalAnimationTypes.h
```diff
@@ -32,29 +34,186 @@ struct PlayState
...
+
+    /**
+     * @brief Add a state as a successor to the current state.
+     */
+    void AddSuccessor(uint32_t successor);
 
     uint32_t id;
     Action action;
     std::string animUid;
+    std::unique_ptr<bool> firstRunComplete;
```
McCallisterRomer, line 197: Didn't do the math, but this may another opportunity for better packing.

include/ncengine/graphics/SkeletalAnimationTypes.h
```diff
@@ -32,29 +34,186 @@ struct PlayState
...
+    uint32_t enterFrom;
+    std::function<bool ()> enterWhen;
+    std::function<bool ()> exitWhen;
+    uint32_t exitTo;
+};
+
+class SparseSet
+{
+    public:
+        SparseSet(uint32_t maxStates);
```
McCallisterRomer: explicit

include/ncengine/graphics/SkeletalAnimationTypes.h
```diff
@@ -32,29 +34,186 @@ struct PlayState
...
+        auto Insert(State toInsert) -> uint32_t;
+        auto Remove(uint32_t toRemove) -> bool;
+        auto Contains(uint32_t id) -> bool;
+        auto Get(uint32_t id) -> State*;
+        auto GetLast() -> State*;
+
+    private:
+        auto AssignId() -> uint32_t { return static_cast<uint32_t>(m_dense.size()); }
```
McCallisterRomer: also const noexcept

source/engine/graphics/system/SkeletalAnimationTypes.cpp
```diff
@@ -18,6 +18,61 @@ auto GetGlobalInverseTransform(const std::vector<nc::asset::BoneSpaceToParentSpa
...
+void State::AddSuccessor(uint32_t successor)
+{
+    successors.push_back(successor);
+}
+
+SparseSet::SparseSet(uint32_t maxStates)
+    : m_dense{},
+      m_sparse(maxStates, State::NullId)
+{
+    m_dense.reserve(maxStates);
```
McCallisterRomer: I haven't seen the typical value of `maxStates` yet, but why reserve? To me that indicates we expect to be at or near that capacity most of the time, which doesn't seem correct. I may be wrong.

source/engine/graphics/system/SkeletalAnimationTypes.cpp
```diff
@@ -18,6 +18,61 @@ auto GetGlobalInverseTransform(const std::vector<nc::asset::BoneSpaceToParentSpa
...
+}
+
+auto SparseSet::Insert(State toInsert) -> uint32_t
+{
+    toInsert.id = AssignId();
+    if (toInsert.id >= m_sparse.size())
```
McCallisterRomer: If the id was coming from a parameter, this check would make sense to give more sensible error message. Since the id is internally generated, and the `.at()` below already does the bounds checking, I think its superfluous.

source/engine/graphics/system/SkeletalAnimationTypes.cpp
```diff
@@ -18,6 +18,61 @@ auto GetGlobalInverseTransform(const std::vector<nc::asset::BoneSpaceToParentSpa
...
+    {
+        throw nc::NcError("Capacity reached. Cannot insert state.");
+    }
+
+    m_sparse.at(toInsert.id) = static_cast<uint32_t>(m_dense.size());
```
McCallisterRomer: I think you need tests for this. I'm a little skeptical of the id generation. `AssignId()` returns the same expression that appears on the right here. If that happens to always valid, doesn't that mean you don't need a sparse set at all? There'd be no reason to ever look up the dense index because its equivalent to the id. Maybe that's the case, and this can all be simplified - neat, but its possible there is wonkiness here.

include/ncengine/graphics/SkeletalAnimator.h
```diff
@@ -13,38 +13,96 @@ namespace nc::graphics
...
+        static constexpr uint32_t MaxStates = 100u;
+
         /**
          * @brief Construct a new SkeletalAnimator component.
          * @param entity The parent entity.
          * @param meshUid The UID of the mesh that will be animated by this component.
          * @param animationUid The UID of the asset::SkeletalAnimation that will become the initial animation for this component. The initial state will always loop.
          */
-        SkeletalAnimator(Entity entity, std::string meshUid, std::string animationUid);
+        SkeletalAnimator(Entity entity, const std::string& meshUid, const std::string& animationUid);
```
McCallisterRomer: Why change to const&? They're still sink params.

include/ncengine/graphics/SkeletalAnimator.h
```diff
@@ -13,38 +13,96 @@ namespace nc::graphics
...
+
+        /** @brief Get the first state in the state machine, initialized by the constructor. */
+        auto RootState() -> uint32_t { return m_states.Get(0)->id; }
+
         /** @brief Connect to SkeletalAnimationSystem. **/
-        auto Connect() noexcept -> Signal<const anim::PlayState&>& { return m_onPlayStateChanged; }
+        auto Connect() noexcept -> Signal<const anim::StateChange&>& { return m_onStateChanged; }
 
-        /** @brief Get UID of mesh that will be animated **/
+        /** @brief Get the UID of the mesh that will be animated **/
```
McCallisterRomer, line 81: Based on our discussion last night, this also needs a method to update the mesh.

include/ncengine/graphics/SkeletalAnimator.h
```diff
@@ -13,38 +13,96 @@ namespace nc::graphics
...
+         */
+        void LoopImmediate(const std::string& animUid, const std::function<bool ()>& exitWhen, uint32_t exitTo);
+
+        /**
+         * @brief Interrupt the state machine with an immediate animation that plays once.
+         * Transitions back into the state machine when the animation has completed.
+         * @param animUid The animation to play
+         * @param exitTo Exit back to this state.
+         */
+        void PlayOnceImmediate(const std::string& animUid, uint32_t exitTo);
```
McCallisterRomer: Not sure how much sense this makes, but I think it would be conventient if the `exitTo` params had a default. Is it sensible to just go back to the root/initial state, if nothing else is specified?

include/ncengine/graphics/SkeletalAnimator.h
```diff
@@ -13,38 +13,96 @@ namespace nc::graphics
...
+         * @brief Called by SkeletalAnimationSystem. Executes the state machine.
+         * All animations whose transition conditions are met emit a signal to be processed.
+         * */
         void Execute();
 
+        /** 
+         * @brief Add an anim::Loop state to the state machine.
+         * @param loopProperties Properties required to create an anim::Loop state.
+         */
+        auto AddState(const anim::Loop& loopProperties) -> uint32_t;
```
McCallisterRomer: Shouldn't all of these params be by value?

include/ncengine/graphics/SkeletalAnimator.h
```diff
@@ -13,38 +13,96 @@ namespace nc::graphics
...
+        /**
+         * @brief Interrupt the state machine and stop the currently playing animation.
+         * Transitions back into the state machine when exitWhen is met.
+         * @param exitWhen Exit back to the state machine when this condition is met.
+         * @param exitTo Exit back to this state.
+         */
+        void StopImmediate(const std::function<bool ()>& exitWhen, uint32_t exitTo);
+
+        /** @brief Get the first state in the state machine, initialized by the constructor. */
+        auto RootState() -> uint32_t { return m_states.Get(0)->id; }
```
McCallisterRomer: const

include/ncengine/graphics/SkeletalAnimator.h
```diff
@@ -13,38 +13,96 @@ namespace nc::graphics
...
 
     private:
-        auto ExecuteInitialState() -> bool;
+        auto ExecuteRootState() -> bool;
+        auto ExecuteImmediateState() -> bool;
+        auto ExecuteExitState() -> bool;
+        auto ExecuteChildEnterState(uint32_t childStateHandle) -> bool;
+        auto AddState(anim::StopImmediate stopImmediateProperties) -> uint32_t;
 
+        anim::SparseSet m_states;
```
McCallisterRomer: This is 408 bytes or 6 cache lines. Member order will help a bit, but the killer is the `std::optional<anim::State>`. IIRC the immediate state is sort of a one-off, rare thing. If so, we could put it behind a `std::unique_ptr`. 168 bytes is feasible with minimal changes. We could also consider a cold pointer, but we'd probably want to land below 128 bytes to make it worth it; requires some benchmarking. That might be better left for a separate issue.

source/engine/graphics/components/SkeletalAnimator.cpp
```diff
@@ -1,45 +1,204 @@
 #include "graphics/SkeletalAnimator.h"
-#include "graphics/MeshRenderer.h"
-#include "graphics/ToonRenderer.h"
-#include "ecs/Registry.h"
 
 namespace nc::graphics
 {
-SkeletalAnimator::SkeletalAnimator(Entity entity, std::string meshUid, std::string animationUid)
+SkeletalAnimator::SkeletalAnimator(Entity entity, const std::string& meshUid, const std::string& animationUid)
```
McCallisterRomer: Pretty much every parameter in this file should be moved (and taken by value).

source/engine/graphics/components/SkeletalAnimator.cpp
```diff
@@ -1,45 +1,204 @@
...
+}
 
 void SkeletalAnimator::Execute()
 {
-    if (ExecuteInitialState())
+    if (ExecuteRootState()) { return; }
+    if (ExecuteImmediateState()) { return; }
+    if (ExecuteExitState()) { return; }
+
+    for (auto childStateHandle : m_states.Get(m_activeState)->successors)
```
McCallisterRomer: `any_of`?

include/ncengine/graphics/SkeletalAnimationTypes.h
```diff
@@ -14,13 +14,15 @@ namespace nc::graphics::anim
...
+    Loop,
+    PlayOnce,
+    Stop
 };
 
 /**
  * @brief A simplified representation of a SkeletalAnimator state machine state sent to SkeletalAnimationSystem on state transition.
  */
-struct PlayState
+struct StateChange
```
McCallisterRomer, line 29: I think the three mesh paths here should be `std::string_view`s. Its a better default choice for event data where the receiver may likely only need to act upon the data before discarding it. They can still choose to make a copy if needed. Looking at the system, I think they're only being used to index maps anyway.

include/ncengine/graphics/SkeletalAnimationTypes.h
jaremieromer: Leaving per conversation

include/ncengine/graphics/SkeletalAnimator.h
jaremieromer: I want to save this for the issue - We should really have a discussion about how we want to handle inter-component dependencies and soft coupling

source/engine/graphics/components/SkeletalAnimator.cpp
jaremieromer: It actually makes the code more verbose and less understandable, in this case.

include/ncengine/graphics/SkeletalAnimationTypes.h
```diff
@@ -9,18 +9,24 @@
 
 namespace nc::graphics::anim
 {
+static constexpr uint32_t NullState = UINT32_MAX;
```
McCallisterRomer: You don't need static on these.

Would these be better named like `NullStateIndex` since they don't actually represent a state, but an index where the state is expected? Maybe they're fine. I'll defer to you.

include/ncengine/graphics/SkeletalAnimationTypes.h
```diff
@@ -31,30 +37,232 @@ struct PlayState
...
+ * The Stop state is used to stop an animation and remove it from SkeletalAnimationSystem's active
+ * animations container.
  */
-struct Initial
+struct Stop
 {
-    Initial(std::string animUid_)
-        : animUid{std::move(animUid_)}{} 
-    std::string animUid;
+        uint32_t enterFrom;
```
McCallisterRomer: Another weird editor thing? A bunch of members got an extra indent.

include/ncengine/graphics/SkeletalAnimationTypes.h
```diff
@@ -31,30 +37,232 @@ struct PlayState
...
+ */
+class StableSet
+{
+    public:
+        /**
+         * @brief Construct a new Stable Set object
+         * 
+         * @param maxStates The container capacity.
+         */
+        StableSet(uint32_t maxStates);
```
McCallisterRomer: explicit

source/engine/graphics/components/SkeletalAnimator.cpp
```diff
@@ -1,45 +1,204 @@
...
+}
+
+void SkeletalAnimator::CompleteFirstRun()
+{
+    if (m_immediateState && m_immediateState->action == anim::Action::PlayOnce)
+    {
+        *m_immediateState->firstRunComplete.get() = true;
+    }
+
+    if (m_states.Get(m_activeState) && m_states.Get(m_activeState)->action == anim::Action::PlayOnce)
```
McCallisterRomer: The compiler cannot see that `Get()` doesn't modify anything (without link-time optimization) and must leave all three calls in. I'd cache the result in a local variable.

source/engine/graphics/system/SkeletalAnimationTypes.cpp
```diff
@@ -18,6 +18,66 @@ auto GetGlobalInverseTransform(const std::vector<nc::asset::BoneSpaceToParentSpa
...
+void State::AddSuccessor(uint32_t successor)
+{
+    successors.push_back(successor);
+}
+
+StableSet::StableSet(uint32_t maxStates)
+    : m_states{},
+      m_sparse(maxStates, NullState)
+{
+    m_states.reserve(maxStates);
```
McCallisterRomer: I think we discussed - reserving 100 each time is overkill. Leaving as default initialized is probably fine, but if you want to reserve, I'd do some reasonable average of like 5 (or whatever you think is appropriate).

source/engine/graphics/system/SkeletalAnimationTypes.cpp
```diff
@@ -18,6 +18,66 @@ auto GetGlobalInverseTransform(const std::vector<nc::asset::BoneSpaceToParentSpa
 
 namespace nc::graphics::anim
 {
+void State::AddSuccessor(uint32_t successor)
+{
+    successors.push_back(successor);
+}
+
+StableSet::StableSet(uint32_t maxStates)
```
McCallisterRomer: I'm fairly certain this class can be further simplified. Lets discuss.

include/ncengine/graphics/SkeletalAnimationTypes.h
jaremieromer: Fixed static. I prefer omitting the "Index" because from the client code it's a little nicer to see: `.exitTo = anim::NullState` and I've been following a convention of referring to state indices as "state" in the client code, as that's the relevant info to them

include/ncengine/graphics/SkeletalAnimationTypes.h
jaremieromer: ugh

jaremieromer: Yep, it's the dang "Default to 8 space indent" thing I have to reset every day.

source/engine/graphics/system/SkeletalAnimationTypes.cpp
jaremieromer: Yup - clobbered when I copy-pasted my working changes with the Code Review's ones afer already making the fix

McCallisterRomer (approved):

PR #511, opened 2024-01-09
Engine Component Serialization Functions

### LINKED ISSUE #504: Add serialize funcs for engine components

McCallisterRomer:
resolves #504

Adding serialization functions for engine components. Rebase went a little sideways, so there are a few redundant commits from the last PR.


include/ncengine/physics/PhysicsBody.h
```diff
@@ -58,6 +58,10 @@ class PhysicsBody final : public ComponentBase
         void Sleep() noexcept { m_awake = false; }
 
         auto GetProperties() const noexcept -> const PhysicsProperties& { return m_properties; }
+        auto GetLinearFreedom() const noexcept -> Vector3 { return ToVector3(GetLinearFreedomXM()); }
```
jaremieromer, line 61: Not exactly sure how ImGui works internally, but are these going to be called every frame? If so, consider caching the NC types.

source/engine/serialize/ComponentSerialization.cpp
```diff
@@ -0,0 +1,219 @@
...
+    serialize::Deserialize(stream, id);
+    serialize::Deserialize(stream, colliderType);
+    serialize::Deserialize(stream, isTrigger);
+    const auto entity = ctx.entityMap.at(id);
+    switch (colliderType)
+    {
+        case physics::ColliderType::Box:
+        {
+            auto properties = physics::BoxProperties{};
+            serialize::Deserialize(stream, properties);
```
jaremieromer, line 80: As an argument for using pointers as out params rather than references - I did really like that the calling code makes it clear via capturing the address that it was an out param.

jaremieromer (approved):

source/engine/serialize/ComponentSerialization.cpp
McCallisterRomer: Yeah, I kind of like it too, but ultimately I think static analysis is right. If you're taking a pointer, you should be doing a null check, which results in a lot of tedious "this should never happen, but..." type code. It also looks (at least to me) horrendously out of place for `std` types.

include/ncengine/physics/PhysicsBody.h
McCallisterRomer: Its only called during serialization right now, so very rarely. My gut tells me caching for imgui would likely have no effect, and may be a pessimization for physics code due to increasing obj size. It would only be called once per frame if the currently selected object has a physics body.

PR #512, opened 2024-01-10
Module Ids

McCallisterRomer:
precursor to #502

Adding unique ids for modules. Like component pools, modules also need some identifier to serialize so we can determine where to route data during deserialization.


include/ncengine/audio/NcAudio.h
jaremieromer, line 1: Add our copyright to all these files

jaremieromer (approved):

Issue #513, opened 2024-01-10
Update resizing and ensure resizing works properly with aspect ratios.


jaremieromer:
Right now, we have disabled screen resizing.
Per a previous discussion, we did not implement aspect ratio adjustment for screen resizing.

We need to do the work to re enable screen sizing (See 'hint' in `WindowImpl.cpp`) and also ensure that aspect ratio adjustment works on resize.

PR #514, opened 2024-01-10
Forced aspect ratio to 16:9

jaremieromer:
### Changes
 - Implemented a forced aspect ratio of 16:9. Now, the image will maintain it's exact bounds across resolutions, but the clear color will fill whatever dimension is negative space between the bounds of the screen extent and the window.
 - Dimensions specified in the config that do not conform to 16:9 will have the screen extent adjusted (not the window) such that it conforms to 16:9 but does not exceed the original dimensions.
 - Added a new method in `Window/WindowImpl` to get the adjusted screen extent.
 - Set the default resolution for the sample to 1600*900
 - Added a helper method in `RenderGraph` to set the viewport and scissor to the adjusted screen extent, and a method to set the viewport and scissor to the full window. This second is required for shadowmapping as its viewport needs to be consistent with its framebuffer size.

### Notes
- I am not sure how to control the location of the Tilde context menus. I know they cache to wherever you placed them last. Right now they don't start in the bounds of the screen extent. Should we fix this? See images. Right now, the "Scene Selection" editor panel conforms to the screen extent but the Tilde context menus do not.
- After some experimentation, I like docking the main editor panel (the one that shows the scenes) to the bottom of the window, not the screen extent. It's horizontal limits are limited to the screen extent. What do you think?
- Notes, per our discussion, I've not targeted resizing for this PR. See #513

1600 x 900

1000 x 1000

2400 x 1200

900 x 1200


source/engine/graphics/api/vulkan/RenderGraph.cpp
```diff
@@ -92,10 +103,9 @@ namespace nc::graphics
 RenderGraph::RenderGraph(const Device& device, Swapchain* swapchain, GpuAllocator* gpuAllocator, ShaderDescriptorSets* descriptorSets, Vector2 dimensions)
     : m_swapchain{swapchain},
       m_gpuAllocator{gpuAllocator},
-      m_descriptorSets{descriptorSets},
-      m_dimensions{dimensions}
+      m_descriptorSets{descriptorSets}
 {
-    auto litPass = CreateLitPass(device, m_gpuAllocator, m_swapchain, m_dimensions);
+    auto litPass = CreateLitPass(device, m_gpuAllocator, m_swapchain, std::move(dimensions));
```
McCallisterRomer: Very minor, but move is equivalent to copy for vector types.

source/engine/graphics/api/vulkan/RenderGraph.cpp
```diff
@@ -110,15 +120,23 @@ RenderGraph::RenderGraph(const Device& device, Swapchain* swapchain, GpuAllocato
...
+    SetViewportAndScissorFullWindow(cmd, dimensions);
+
+    auto& shadowMappingPass = m_renderPasses[0];
+    shadowMappingPass.Begin(cmd, frameBufferIndex);
+    shadowMappingPass.Execute(cmd, frameData);
+    shadowMappingPass.End(cmd);
+
+    SetViewportAndScissorAspectRatio(cmd, dimensions, screenExtent);
+
+    for (auto& renderPass : m_renderPasses | std::views::drop(1))
```
McCallisterRomer, line 139: nice

test/graphics/GraphicsUtilities_tests.cpp
```diff
@@ -0,0 +1,34 @@
+#include "gtest/gtest.h"
```
McCallisterRomer, line 1: Our first graphics test! :)

McCallisterRomer (approved):
Awesome!

PR #565, opened 2024-01-31
Editor save/load scene

### LINKED ISSUE #547: Scene save/load in editor
Add options for New Scene, Save Scene, and Load Scene in the editor.

McCallisterRomer:
resolves #547

Adding New/Save/Load scene options to editor. These have major improvements over what was in dev/game-jam
- Save has options for including/excluding assets, persistent entities, and can filter entities by layer.
- Save can optionally create backups if you're overwriting an existing save file.
- Load can open in a new scene or overlay on to the current scene.
- Dialogs have a dedicated place to display errors and we can propagate some filesystem related error messages.

Building an FPS overlay into the editor like we had for the game jam. It can be toggled through a menu in the editor. It stays open if the editor is closed.

Also a bunch of refactoring so the editor operates on a `ModuleProvider` rather than just `NcAsset`. For these changes, we also needed to pass in `NcScene`, but the list will continue to grow as features are added.


jaremieromer (commented):
Summarizing a few bugs/observations found here (might be all related to the same bug)

1. Scene -> Create New Scene crashes with bad_function_call if done on any scene with an existing FPS counter
2. If you load a scene that has assets that are not currently loaded, we crash with "Asset not loaded: X"

**Steps to reproduce:**

-  Open "JareTestScene"
- Save as a scene
- Close the application
- Open the sample, stay in Worms scene.
- Load the saved scene from Step 2, see assets loaded as part of JareTestScene are not loaded

Note: Item 2 is not valid as written because I missed the "Include Loaded Assets" checkbox. However, following those same steps but saving with "Include Loaded Assets" causes a hang with vk::DeviceLost that we should investigate.

Note2: This does not crash as explained in MSVC, just GCC

McCallisterRomer:
> 1. Scene -> Create New Scene crashes with bad_function_call if done on any scene with an existing FPS counter

Fixed. Had a bug in all sample scenes where the registered scene-specific UI wasn't being unregistered in scene destructors. Its been hidden forever because all sample scene constructors overwrite the value with their own widget or nullptr.

jaremieromer (approved):
I have thoroughly tested this branch with my changes  from PR #570 merged in. Approved once mine is in and merged.

PR #572, opened 2024-02-06
Wireframe Rendering, Create Entity Dialog, EditorCamera

### LINKED ISSUE #446: Cleanup Collider editor selecting
`Collider` manages its selection state in the editor and has a member for creating a debug widget. This should be tracked by the editor/widget system and remaining `#ifdef NC_EDITOR_ENABLED` should be removed from `Collider`.

### LINKED ISSUE #548: Show selected entity in editor
Add ability to render a wireframe model for the current selected entity.

### LINKED ISSUE #549: Configure Entity props when adding through editor
We need to be able to set Entity flags and layers when adding an Entity (or child) through the editor.

### LINKED ISSUE #566: Entity node collapsing
Update Entity nodes in scene graph so they expand/collapse only when the arrow is clicked.

McCallisterRomer:
resolves #446, resolves #548, resolves #549, resolves #566

As discussed, since these individual changes trampled the same editor functions, I've combined them into one (fairly large) changeset.

Major changes:
- Adding wireframe rendering for the selected `Entity`.
  - To avoid previous special case handling of similar things, I've added a new `WireframeRenderer` component, and updated `WireframeTechnique` and `WidgetSystem` to operate on this. Much nicer, IMO.
  - Updated `Collider` wireframe rendering to use this same solution.
  - Removed `DebugRenderer` since its terrible and can be reworked to use this solution just as easily.
- Adding a dialog for creating entities that allows configuring all of the EntityInfo properties. I left a 'Create Default Entity' option in the context menu to skip the dialog and just initialize the `Entity` with `EntityInfo{}`.
- Adding an internal `EditorCamera` component so we can always have a debug/scene camera. By default, F5 toggles between this camera and whatever camera the scene registered.

Minor changes:
- Adding an object menu to the scene graph context menu for quickly making basic cubes, lights, etc.
- Tweaking the scene graph so you can have multiple root nodes open at once.
- Adding remaining flags to `Entity` in the inspector. Also putting these behind a collapsible header and making them (always disabled) checkboxes for easier reading.
- Fixed various `Entity` functions that that are supposed to return a bool (they returned uint8_t which was just implicitly converted everywhere).

Edit
Additional stuff for MinGW:
- Fixed self move bug when clearing scene data: When we change scenes, we shift all persistent data to the front of their respective pools. If an object is already in the correct position, we would move assign to itself, which is valid, but not what we want. WRT rvalue argument binding, the standard says "...the implementation is free to optimize away aliasing checks which might be needed if the argument were an lvalue..." GCC appears to take advantage of this and leaves those objects in a 'moved from' state.
- Linking `RegistryFactory.obj` is failing again for MinGW. I'm going to split type registration into multiple .cpp files address this. Since that's an additional 15 or so files changed, I'm going to do it in a separate PR. I'll get that up before merging this, but just note this branch won't link with MinGW (real deal GCC is fine).


include/ncengine/graphics/WireframeRenderer.h
```diff
@@ -0,0 +1,39 @@
...
+ * When 'source' is set to WireframeSource::Collider, the underlying geometry of the Collider attached to 'target' will
+ * be used to render a wireframe. If 'target' does not contain a Collider, 'target' will be set to Entity::Null() during
+ * the next pass of ExecutionPhase::Render.
+ * 
+ * The WireframeRenderer instance's parent Entity or another Entity may be specified for 'target'. If 'target' no longer
+ * refers to a valid Entity, it will be set to Entity::Null() during the next pass of ExecutionPhase::Render.
+ */
+struct WireframeRenderer
+{
+    Entity target = Entity::Null();
```
McCallisterRomer, line 36: In case its confusing why this takes a target rather than operating on its own data, I just didn't want the editor to have to modify client objects. This allows us to make our own entities/components and non-intrusively target the objects we're interested in.

source/engine/ui/editor/Editor.h
```diff
@@ -14,24 +15,55 @@ struct EditorHotkeys
     input::KeyCode openNewSceneDialog = input::KeyCode::F1;
     input::KeyCode openSaveSceneDialog = input::KeyCode::F2;
     input::KeyCode openLoadSceneDialog = input::KeyCode::F3;
+    input::KeyCode toggleEditorCamera = input::KeyCode::F5;
```
McCallisterRomer, line 18: Feel free to suggest changes to default hotkeys.

include/ncengine/graphics/DebugWidget.h
jaremieromer, line 1: Nice

source/engine/graphics/api/vulkan/techniques/WireframeTechnique.cpp
```diff
@@ -8,17 +8,13 @@
...
 #include "graphics/api/vulkan/shaders/ShaderUtilities.h"
-#include "graphics/debug/DebugRenderer.h"
 #include "graphics/PerFrameRenderState.h"
 
 namespace nc::graphics
 {
     WireframeTechnique::WireframeTechnique(const Device& device, ShaderDescriptorSets*, vk::RenderPass* renderPass)
         : m_pipeline{nullptr},
           m_pipelineLayout{nullptr}
-          #ifdef NC_DEBUG_RENDERING_ENABLED
```
jaremieromer, line 19: so nice to see this go away!

source/engine/graphics/api/vulkan/techniques/WireframeTechnique.cpp
```diff
@@ -103,55 +92,20 @@ namespace nc::graphics
...
-        #endif
+        return !frameData.widgetState.wireframeData.empty();
     }
 
     void WireframeTechnique::Record(vk::CommandBuffer* cmd, const PerFrameRenderState& frameData)
     {
         auto pushConstants = WireframePushConstants{};
         pushConstants.viewProjection = frameData.cameraState.view * frameData.cameraState.projection;
 
-        const auto& widget = frameData.widgetState.selectedCollider;
```
jaremieromer, line 121: So much simpler!

source/engine/ui/editor/impl/EditorCamera.cpp
```diff
@@ -0,0 +1,53 @@
+#include "EditorCamera.h"
```
jaremieromer, line 1: As it is now, the "both" cameras have their transforms tied. Can we cache the transform data of each as we toggle so that moving one camera does not also move the other?

source/engine/ui/editor/impl/EditorImpl.cpp
```diff
@@ -1,27 +1,70 @@
...
-#include "input/Input.h"
-#include "window/Window.h"
+#include "ncengine/ecs/Logic.h"
+#include "ncengine/ecs/InvokeFreeComponent.h"
+#include "ncengine/input/Input.h"
+#include "ncengine/window/Window.h"
+
+namespace
+{
+auto BuildEditorObjectx(nc::ecs::Ecs world, nc::ModuleProvider modules, nc::input::KeyCode cameraHotkey) -> std::pair<nc::Entity, nc::Entity>
```
jaremieromer: typo in method name, should be "BuildEditorObject"

jaremieromer (approved):
Approved once the camera bug is fixed

PR #577, opened 2024-02-08
Update Sample Scenes

McCallisterRomer:
Reducing/combining sample scenes:
- PhysicsTest is a combination of the previous JointsTest and SpawnTest scenes
- GraphicsTest is just a rename of JareTestScene
- Benchmarks has bulk spawners for `MeshRenderer`, `ToonRenderer`, `PointLight`, `Collider`, `PhysicsBody` and variable depth `Entity` hierarchies. Shapes can be set through the UI now.

All other scenes are deleted, along with prefabs and unused assets.

Also added `Transform::LookAt()` and `DecomposeMatrix()` functions for a camera in the physics scene.


jaremieromer (approved):
So nice! I love the simplified scenes

Issue #433, opened 2023-10-04
Remove/Refactor ActiveRegistry() usage


McCallisterRomer:
There are still cases where `ActiveRegistry()` is used to query for a static `Registry` instance. We'd eventually like to support multiple registries, and removing reliance on a static instance will help that along. The concept of an active registry should still exist, but the state should be queried from the engine somehow, rather than globally.

PR #579, opened 2024-02-09
Transform Updates

### LINKED ISSUE #490: Simplify Transform constructor
Entity parenting happens inside the constructor of `Transform`. Its one of the global state/`ActiveRegistry` usages that needs to be refactored. Code was written before concepts were available (and probably before I fully grokked constexpr if. It looks very possible to hoist that responsibility out to `EcsInterface`.

Consider inlining the constructor as well.

### LINKED ISSUE #556: Transform hierarchy serialize/load issues
"There are bugs related to transform hierarchies in probably both the editor and serialization. With a child parented to an Entity with a non-identity transform, modifying world space values in the editor is unpredictable (might be good to expose local values too). Loading the same scenario from a prefab seems to perform incorrect multiplication. Maybe we're saving the child's world space values and then double transforming on load? Idk, should check existing tests for coverage of cases like this."

McCallisterRomer:
resolves #490, resolves #556, part of #433

- Fixed UI and scene loading issues when parent entities have non-identity transforms. We were reading world space values into the UI widgets and then writing those back to local space. Yikes. Since all of our transform writes go into the local space matrix, I've just changed the UI to show local values. There's nothing stopping us from having both, but we'd have to add/test a bunch of functions.
- Pulled parent/child information out of `Transform` into a `Hierarchy` component. This was primarily done to remove calls to `ActiveRegistry()` in `Transform.cpp` (it doesn't even include `Registry.h` anymore!), but it has additional benefits: `Transform` is a bit smaller; scene traversal usually doesn't care about matrices and vice versa.
  - Also added some utility functions to `EcsInterface` to make usage easier.
- Changed our world space calculation approach from running on each write to a `Transform` to once per frame. We've never really hit a performance penalty here since we don't usually have many/deep hierarchies, but its a killer when you do have them. The update is done by the `EcsModule` during `ExecutionPhase::PreRenderSync` for now. I'm exactly 83% sure it could run in parallel to some physics things since physics only reads/writes at the beginning and end of the pipeline, but I didn't want to mess with that right now.
- Combined redundant vector/matrix functions in transform and animation code to `MatrixUtiliites.h`.

I'm very excited - this gets us down to only 3 calls to `ActiveRegistry()`!


jaremieromer (approved):
I am amazed this works so well even saving and loading scenes with hierarchies. Nice job!

PR #583, opened 2024-02-12
Add GetEntityByTag

### LINKED ISSUE #509: Add convenience function for finding Entity by Tag
Its possible to find an entity by `Tag` with something like:
```cpp
auto tags = ecs.GetAll<Tag>();
auto pos = std::ranges::find_if(tags, [&str](auto tag){ return tag->Value() == str; });
if (pos != std::ranges::end(tags)
{
    // ...
}
```

Although it would always be preferred to just store the `Entity` to avoid this, I expect there will be some cases where we want to be doing this, and its is a little too verbose for my liking.

Add something like `auto EcsInterface::Get<Entity>(std::string_view) -> Entity`. It'll have to be constrained on the `AccessPolicy` having access to `Entity` and `Tag`.

McCallisterRomer:
resolves #509


jaremieromer (approved):
We probably want a method to get all entities by tag as well. 
Use case: turn all trees with tag "SickTree" into healthy trees, etc.

McCallisterRomer:
> We probably want a method to get all entities by tag as well. Use case: turn all trees with tag "SickTree" into healthy trees, etc.

If you really had to do that, you can filter + transform the `Tag` pool like:
```cpp
auto matches = world.GetAll<Tag>()
    | std::views::filter([v](auto&& tag) { return tag.Value() == v; })
    | std::views::transform([](auto&& tag) { return tag.ParentEntity(); });
``` 

The same could be done for `GetEntityByTag()`; I'm only adding it because it came up a lot in SS & DD due to missing editor features.

I didn't add a find all here because I think its a mistake to encourage bulk operations based on tag, and we have a number of options that accomplish similar things. For your example use case, it would be more performant and less error prone to operate on Healthy/Sick component pools. More generally, if entities are added entirely through code, you can cache the entities and eliminate the need to search (we do this in the `Spawner` from the sample). If the entities are created from a scene fragment, that fragment can group them by layer (we did this in SS & DD).

PR #584, opened 2024-02-12
Configurable audio buffer size

### LINKED ISSUE #555: Investigate proper audio buffer size
During game jam, audio buffer size had to be bumped to 512. This seemed fine all people's devices, but maybe we want a config option to control this.

McCallisterRomer:
resolves #555

Allowing preferred audio buffer frames to be set through the config.


include/ncengine/config/Config.h
```diff
@@ -83,6 +83,7 @@ struct PhysicsSettings
 struct AudioSettings
 {
     bool enabled = true;
+    unsigned bufferFrames = 512u; // must be a multiple of two in the range [16, 2056]
```
jaremieromer: Change comment to "Must be a **power** of two..."

jaremieromer: Also, change to [16, 2048]

jaremieromer (approved):

Issue #574, opened 2024-02-06
Disable Transform UI for Static Entities


McCallisterRomer:
Probably requires further consideration - there may be a better option.

Static entities aren't supposed to move. We don't make a ton of decisions based on this, but if the entity happens to have a concave collider, its position will not get updated when the transform is moved through the UI. The simplest solution is to just add `ImGui::BeginDisabled(e.IsStatic())` to the transform widget so you can't ever move them. This makes level editing more annoying, but maybe we could expose some sort of recalculate statics option. AFAIK, this would only have to rebuild the bsp tree right now, but future optimizations on statics would need a way to hook into this as well.

McCallisterRomer:
not needed, auto updates now

PR #587, opened 2024-02-14
Enable Static Entity Rebaking

McCallisterRomer:
related to #574

Under normal circumstances, static entities should not move. When positioning things in the editor, however, they do. Adding a signal that can be fired by the editor to notify modules of the broken promise. I imagine someday we'll want to update this to include exactly which entities need updating, but, given the scope of what we're currently doing, its totally fine to recompute everything.

We're not really optimizing for this everywhere, but the following are changed accordingly:
- `BspTree` connects a `Rebuild()` method.
- `EcsModule` connects a `UpdateStaticWorldMatrices()` method.
  - Changed the normal update to skip static entities.
- The editor debug menu has an option to fire the event.
  - This is workable, but probably pesky. I want to experiment with it a bit before making any more involved changes, like automatically firing when you update a static transform.

### DIFF
--- a/source/engine/physics/collision/BspTree.h
+++ b/source/engine/physics/collision/BspTree.h
@@ -101,10 +101,6 @@ void BspTree::FindPairs(std::span<const ProxyType> proxies)
         if (proxy.Id().IsStatic())
             continue;
 
-        // DO NOT CHECKIN
-        if (proxy.Id().IsStatic())
-            continue;
-
         /** Find mesh colliders that are in the same region as the estimate. */
         narrowTestMeshIndices.clear();
         BroadTest(0u, proxy.estimate, narrowTestMeshIndices);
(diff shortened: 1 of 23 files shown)


source/engine/physics/collision/BspTree.h
```diff
@@ -97,6 +99,10 @@ void BspTree::FindPairs(std::span<const ProxyType> proxies)
     {
         const auto& proxy = proxies[i];
 
+        // DO NOT CHECKIN
```
jaremieromer: Should be removed

source/engine/ui/editor/impl/EditorUI.cpp
```diff
@@ -6,6 +6,10 @@
 #include "ncengine/ui/ImGuiUtility.h"
 #include "ncengine/window/Window.h"
 
+
```
jaremieromer: A couple extra newlines here

jaremieromer (approved):

PR #592, opened 2024-02-19
Simplify PointLight and Tag

McCallisterRomer:
Simplifying `PointLight` and `Tag` components by making them structs. `Tag` doesn't need to derive from `ComponentBase`, so I've also removed that. The same is true for `PointLight`, but that's a breaking change to serialization, and I don't want struggle through that with any future changes to SS&DD.

I'm also removing the editor macros from all component declarations. They're not relevant in the new editor.


include/ncengine/ecs/Tag.h
```diff
@@ -19,28 +19,8 @@ namespace nc
  * 
  * A Tag's initial value can be set in EntityInfo when adding an Entity.
  */
-class Tag final : public ComponentBase
+struct Tag
 {
```
jaremieromer, line 23: so slim!

jaremieromer (approved):

Issue #363, opened 2023-03-13
Don't include module implementations in engine


McCallisterRomer:
The BuildXXXModule() factories should be in the public header for each module, and the engine should be able to construct all modules without including the impl headers. NcGraphics is the only difficult one because it takes some 'internal' parameters.

PR #599, opened 2024-02-22
Audio/Ecs Module Cleanup

McCallisterRomer:
part of #363

- Making `EcsModule` look like other modules. We don't have any functions we need to expose right now, so the public-facing header is pretty empty.
- Moving the `NcAudio` factory out of NcAudioImpl.h.
- Replacing all usage of `Registry` in `NcAudio` and `NcEcs` with the newer types. I want to do this for all modules, but some of the usage is pretty heavy. These ones were pretty simple.


include/ncengine/audio/NcAudio.h
```diff
@@ -4,14 +4,23 @@
  */
 #pragma once
 
+#include "ncengine/audio/AudioSource.h"
 #include "ncengine/ecs/Entity.h"
+#include "ncengine/ecs/Ecs.h"
```
jaremieromer, line 8: Alphabetize here

jaremieromer (approved):

source/engine/ecs/NcEcsImpl.cpp
```diff
@@ -1,27 +1,26 @@
-#include "EcsModule.h"
+#include "NcEcsImpl.h"
 #include "ncengine/Events.h"
 #include "ncengine/ecs/Logic.h"
+#include "ncengine/ecs/ComponentRegistry.h"
```
jaremieromer, line 2: Alphabetize here

test/ecs/Transform_unit_tests.cpp
```diff
@@ -1,6 +1,6 @@
 #include "gtest/gtest.h"
 #include "ecs/Registry.h"
-#include "ecs/EcsModule.h"
+#include "ecs/NcEcsImpl.h"
```
jaremieromer: Alphabetize here

jaremieromer (approved):

Issue #601, opened 2024-02-23
Investigate SwiftShader for CI GFX Testing


McCallisterRomer:
Possible cross platform gpu/driver replacement for CI environments: https://github.com/google/swiftshader

McCallisterRomer:
Forked repo [here](https://github.com/NcStudios/swiftshader).  The build time on this is long, so I'll create a release with artifacts at some point. Need to sort out issues with macOS (build fails with clang; I don't think we can use XCode for 'reasons'; arm64 build of project is based on host architecture (and you have to pay for Apple Silicon on GitHub), so we may have to fiddle with their cmake to produce a universal binary - need to look deeper).

Use is pretty slick. You can just rename the dll to vulkan-1.dll and place it next to your executable. To get validation layers working with it, I left the filename as is, and went through the sdk loader with VK_ICD_FILENAME pointing to the icd json (created from swiftshader build), which points to the dll. Something like this:
```
project
|  app.exe
|  vk_swifshader.dll
|  vk_swiftshader_icd.json
```

vk_swiftshader_icd.json:
```json

{
    "file_format_version": "1.0.0",
    "ICD": {
      "library_path": "..\\vk_swiftshader.dll",
      "api_version": "1.0.5"
    }
}
```

Then launch with:
```powershell
$env:VK_ICD_FILENAMES="vk_swiftshader_ics.json"; .\\app.exe
```

McCallisterRomer:
Did some more testing. Some features we use are not supported. Carelessly disabling them allows things to run, but the smoke test scene crashes. Obvisouly, features need to be sorted out correctly, but I was able to get GH agents to launch the sample with this.

Issue #252, opened 2022-03-12
Automated Integration Tests


McCallisterRomer:
A lot of our testing is done through manual verification:
 1. Make a change in physics.
 2. Build the sample project.
 3. Check changes in a few of the scenes.
 
This whole flow should be handled automatically in a GitHub action and branch protection rules should exist for vnext, blocking on failure.

This depends on first being able to build and run a headless sample on a GitHub runner.

Questions:
* Do we use an existing testing framework?
  * I was able to get a prototype working with gtest, but there is a bit of legwork in writing the tests because the engine is actually running: needs a scene, a way to perform test actions, a mechanism for collecting test restults, then the results are actually tested after stopping the engine.
* If we don't use gtest, how do we collect, verify, and publish the test results?

McCallisterRomer:
We spoke a little more about this, and maybe a simpler smoke test-based thing is sufficient for now. Rather than trying to do granular testing of each step as the engine is running, we just run a test app and check the exit code. So there's no real test framework needed, just one pass/fail check. Context/error details could be logged and we can publish that log from CI.

McCallisterRomer:
SmokeTest scene added. Needs #601 before adding to CI.

PR #602, opened 2024-02-24
Add Smoke Test

McCallisterRomer:
part of #252

Adding a test scene to the sample that can be run from CI. Requires #601 before we can actually run it there, but it can be run locally with `./Sample --run-test`.

It caught a bug with `AudioSource`, so I fixed that and added some test coverage.


jaremieromer (approved):
Amazing!!

Issue #562, opened 2024-01-31
Investigate physics bugs and/or cost of using an alternative implementation


McCallisterRomer:
~"Static entities get manipulated sometimes. They're also getting included in collision checks that should be elided earlier on. useGravity and isKinematic aren't always respected?"~ fixed

Certain bias/softness values on joints occasionaly failed asserts in DirectXMath.

Also general stability issues.

I've seen the 'NAN in barycentric protection' assert a few times. Reproducible in PhysicsTest scene when spawning capsules.

McCallisterRomer:
For movement of static bodies: It appears this is due to position correction. Check the `CollisionEventType` generation for gaps.

PR #604, opened 2024-02-26
Rework PhysicsBody and Add PhysicsMaterial

McCallisterRomer:
part of #433, part of #562

Position correction could incorrectly move static bodies. The only reason we want a `PhysicsBody` on a static `Entity` is for friction/restitution values. I've pulled those properties out into a `PhysicsMaterial` component which will get used in any collision event (no `PhysicsBody` required). Any colliding `Entity` that doesn't have one uses the default properties.

I also reworked the `PhysicsBody` constructor to take the `Transform` and `Collider` directly rather than finding them through `ActiveRegistry()`. It no longer needs to derive from `ComponentBase` after this.

`ClientObjectProperties` had a weird constructor only used in tests, so I cleaned that up (I was originally trying to fix the problem here).

I had to expose a `GetPool()` function for the sample scenes so I could reserve pools to keep references valid. I ended up just replacing the `OnPool()` function with this.


jaremieromer:
It'd be nice if we could disable the Transform component (gray it out) if an entity is static. Is this easy to add?

include/ncengine/physics/PhysicsBody.h
```diff
@@ -24,23 +22,29 @@ enum class IntegrationResult
     Integrated, Ignored, PutToSleep
 };
 
-class PhysicsBody final : public ComponentBase
+class Collider;
+
+class PhysicsBody final
```
jaremieromer: final seems a little unnecessary here since it is not involved in inheritance anymore

McCallisterRomer:
> It'd be nice if we could disable the Transform component (gray it out) if an entity is static. Is this easy to add?

We have an issue for it. Easy to add disabling, but we also want a way to override that and allow editing, which is more involved.

jaremieromer (approved):

PR #610, opened 2024-02-29
Create separate update/render task graphs

### LINKED ISSUE #605: Add separate task graph for rendering

McCallisterRomer:
resolves #605

Adding an additional task graph for rendering to start the decoupling effort. No logical changes here, it just touched a lot of files, so its going up in its own PR.

Okay, one minor logical change: `WindowImpl` was calling `NcEngine::Stop()` twice on window close (once explicitly and once through callback). Fixed.


source/engine/task/Executor.h
```diff
@@ -7,28 +7,39 @@
 
 namespace nc::task
 {
+// Executor state consiting of both update and render contexts
```
jaremieromer: *consisting

jaremieromer (approved):

PR #612, opened 2024-02-29
StepTimer

### LINKED ISSUE #417: Add thread count to config
The `Executor` hardcodes the number of threads to create. This value should be added to the config and read dynamically. There should also be a default value that effectively uses `std::thread::hardware_concurrency() - numOtherEngineThreads`.

### LINKED ISSUE #606: Add step timer for update loop
By default we want to run the whole update loop on a fixed step, rather than just physics. We could also have an option to run on a variable time step, as long as there is a way to stop getting too far ahead of rendering.

McCallisterRomer:
resolves #606, resolves #417

Replacing `TimeModule` with a `StepTimer`. Rather than having time calculation be a step in the update loop, things are inverted so time calculation conditionally fires the update loop. This allows setting a fixed time step for the whole loop, rather than just physics. It can still be set to a variable time step mode.

Since I was already updating `Config` with the new `EngineSettings`, I also put a thread count option in there to knock out #417.


include/ncengine/config/Config.h
```diff
@@ -145,6 +154,9 @@ auto Validate(const Config& config) -> bool;
 /** @brief Get the ProjectSettings NcEngine was initialized with. */
 auto GetProjectSettings() -> const ProjectSettings&;
 
+/** @brief Get the EngineSettigns NcEngine was initialized with. */
```
jaremieromer: Typo in comment

sample/config.ini
```diff
@@ -2,6 +2,10 @@
 project_name=NcEngine Samples
 log_file_path=Sample.log
 log_max_file_size=1000000
+[engine_settings]
```
jaremieromer, line 5: Is this something we can easily obfuscate if we were to make, for example, a P2P networked game? We would not want users to be able to modify their time_step like this in that setting.

jaremieromer: Ignore, we discussed this

jaremieromer: We addressed my other concerns in person. Looks good!

jaremieromer (approved):

PR #621, opened 2024-03-02
Support true 'Per Frame In Flight' rendering

### LINKED ISSUE #563: Split out descriptor set into several
The monolithic descriptor set needs to be split out to provide room for more descriptors while staying within current portability hardware maximums per descriptor set.
Also, dividing the descriptors into multiple sets will improve performance as we have many pipelines that do not need to bind all the data.

## Current Descriptor Set

0: Object Data (SSBO) `readonly buffer ObjectBuffer`

```cpp
struct ObjectData
{
    // N MVP matrices
    mat4 model;
    mat4 modelView;
    mat4 viewProjection;

    // Textures
    uint baseColorIndex;
    uint normalIndex;
    uint roughnessIndex;
    uint metallicIndex;

    uint skeletalAnimationIndex;
};
```
1: Point Lights (SSBO) `readonly buffer ObjectBuffer`
```cpp
struct PointLight
{
    mat4 lightViewProj;
    vec3 lightPos;
    int castShadows;
    vec3 ambientColor;
    float attLin;
    vec3 diffuseColor;
    float attQuad;
    float specularIntensity;
    int isInitialized;
};
```

2: Textures `uniform sampler2D textures[]`

3: Shadow Maps `uniform sampler2D shadowMaps[];`

4: Cube Maps `uniform samplerCube cubeMaps[];`

5: Environment Data  
```cpp
uniform EnvironmentDataBuffer
{
    vec3 cameraWorldPosition;
    int skyboxCubemapIndex;
} environmentData;
```

6: Animated Bones `readonly buffer AnimatedBoneTransforms`
```cpp 
struct SkeletalAnimationData
{
    mat4 finalTransform;
};

readonly buffer AnimatedBoneTransforms
{
	SkeletalAnimationData animatedBones[];
} skeletalAnimationBuffer;
```

## Proposed Changes
**Descriptor Set #0, Global Data**
0: Global Data Buffer
```cpp
struct GlobalDataBuffer
{
	mat4 cameraViewProjMat;
    vec3 cameraWorldPos;
    int useCubeMap;
    vec3 ambientColor;
	int cubeMapTextureIndex;
}
```
1: Cube Maps `uniform samplerCube cubeMaps[];`

**Descriptor Set #1, Per Object Data**
0: Object Data (SSBO) `readonly buffer ObjectBuffer`

```cpp
struct ObjectData
{
    // N MVP matrices
    mat4 model;
    
    // Textures
    uint baseColorIndex;
    uint normalIndex;
    uint roughnessIndex;
    uint metallicIndex;

    uint skeletalAnimationIndex;
};
```
1: Textures `uniform sampler2D textures[]`

2: Animated Bones `readonly buffer AnimatedBoneTransforms`
```cpp 
struct SkeletalAnimationData
{
    mat4 finalTransform;
};

readonly buffer AnimatedBoneTransforms
{
	SkeletalAnimationData animatedBones[];
} skeletalAnimationBuffer;
```
**Descriptor Set #2, Lighting Data**
0: Point Lights (SSBO) `readonly buffer ObjectBuffer`
```cpp
struct PointLight
{
    mat4 lightViewProjMat;
    vec3 lightWorldPos;
    int castShadows;
    vec3 diffuseColor;
    float specularIntensity;
    int isInitialized;
};
```
1: Shadow Maps `uniform sampler2D shadowMaps[];`

### LINKED ISSUE #564: Split rendering into separate sets per frame in flight
We are currently rendering the same set of state twice, once per each frame in flight. we want to render one frame ahead

jaremieromer:
## Goal
Split out the singular descriptor set and shader resources into N sets, where N is `MaxFramesInFlight`.

## Changes

 - Created api-agnostic buffer handles in `graphics/shader_resource/`.
 - These buffers are created with a size and shader binding parameter and can then be bound with data which is sent to the GPU via `ShaderResourceBus`.
 - A new class, `ShaderStorage`, receives the events sent from the buffer handle types and manages the underlying vulkan data.
 - All the hard-coded classes surrounding shader resources have been removed.
 - `MaxFramesInFlight` and other constants have been moved to a new header called `GraphicsConstants`
 - All shader resource instances (point lights, objects, environment data, shadow maps, meshes, textures and cubemaps) are now created in `NcGraphicsImpl`via the api-agnostic handles.

## Todo

 1. Use `std::variant` to eliminate extra data being passed in the api-agnostic buffer handle `xUpdateEventData` types.
 2. Add tests

resolves #564
resolves #563


source/engine/graphics/api/vulkan/buffers/CubeMapArrayBuffer.h
```diff
@@ -0,0 +1,45 @@
...
+{
+struct CubeMapWithId;
+} // namespace asset
+
+namespace graphics::vulkan
+{
+class CubeMap
+{
+    public:
+        inline static uint32_t CubeMapUid = 0u;
```
McCallisterRomer: Assuming, since this is a 'uid', it should be const/constexpr?

source/engine/graphics/api/vulkan/buffers/StorageBuffer.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+void StorageBuffer::Bind(const void* dataToMap, uint32_t dataSize)
+{
+    auto allocation = m_buffer.Allocation();
+    void* dataContainer = m_allocator->Map(allocation);
+
+    if (m_previousDataSize > 0u)
+    {
+        auto zeroBuffer = std::vector<char>(m_previousDataSize);
+        memcpy(dataContainer, zeroBuffer.data(), m_previousDataSize);
```
McCallisterRomer: There's a c function called `memset` that is more what you're looking for.

source/engine/graphics/api/vulkan/buffers/UniformBuffer.cpp
```diff
@@ -1,50 +1,40 @@
...
 
-    void UniformBuffer::Clear()
-    {
-        m_buffer.Release();
-        m_allocator = nullptr;
-    }
+void UniformBuffer::Clear()
+{
+    auto zeroBuffer = std::vector<char>(m_alignedSize);
+    Bind(zeroBuffer.data(), m_alignedSize);
```
McCallisterRomer: I would `memset` here to, rather than trying to implement in terms of `Bind'.

source/engine/graphics/api/vulkan/ShaderBindingManager.h
```diff
@@ -0,0 +1,78 @@
...
+    std::unordered_map<SetIndex, DescriptorWrites> writesPerSet;
+    std::unordered_map<SetIndex, bool> isDirty;
+};
+
+struct DescriptorSetLayoutsChanged
+{
+    uint32_t unused;
+};
+
+class ShaderBindingManager
```
McCallisterRomer: Since you're caching pointers to this guy in a few places, I'd recommend deriving from `StableAddress`(or just deleting copy/move).

source/engine/graphics/api/vulkan/FrameManager.cpp
```diff
@@ -2,16 +2,18 @@
...
 
 namespace
 {
 auto CreatePerFrameGpuContextVector(const nc::graphics::Device& device) -> std::vector<nc::graphics::PerFrameGpuContext>
 {
     auto out = std::vector<nc::graphics::PerFrameGpuContext>{};
     out.reserve(nc::graphics::MaxFramesInFlight);
-    std::generate_n(std::back_inserter(out), nc::graphics::MaxFramesInFlight, [&device]()
+    auto index = 0u;
+    std::generate_n(std::back_inserter(out), nc::graphics::MaxFramesInFlight, [&device, &index]()
```
McCallisterRomer: Its completely fine the way it is, so more of an 'in case you didn't know', you can use init capture to create variables entirely within the closure's scope. Just helps keeping surrounding scope clean of any helpers.

`[i = 0u]() mutable {...};`

(lambda has to be mutable if you modify a non-reference capture)

source/engine/graphics/api/vulkan/RenderGraph.h
```diff
@@ -1,47 +1,90 @@
...
         const Device* m_device;
         Swapchain* m_swapchain;
         GpuAllocator* m_gpuAllocator;
-        ShaderDescriptorSets* m_descriptorSets;
-        std::vector<RenderPass> m_shadowMappingPasses;
-        RenderPass m_litPass;
+        ShaderBindingManager* m_shaderBindingManager;
+
+        // Render Passes
+        std::array<std::vector<std::unique_ptr<RenderPass>>, MaxFramesInFlight> m_shadowMappingPasses;
```
McCallisterRomer: I've been poking around, and I can't figure out why these are `unique_ptr`s now. Going out on a limb, is it because the `array` tries to initialize them, and they aren't default constructible? If so, there's a way around that with pack expansion. For lit passes, you'd make a factory like this:
```cpp
auto CreateLitPasses(const nc::graphics::Device* device,
                     nc::graphics::GpuAllocator* alloc,
                     nc::graphics::Swapchain* swapchain,
                     const nc::Vector2& dimensions)
{
    return [&] <size_t... N> (std::index_sequence<N...>)
    {
        return std::array{(N, CreateLitPass(device, alloc, swapchain, dimensions))...};
    }(std::make_index_sequence<nc::graphics::MaxFramesInFlight>());
}
```

McCallisterRomer: Its a bit weird, but the array initialization expands to something like this, with the element count based on `MaxFramesInFlight`:

`std::array{(0, pass0), (1, pass1)}`

The comma operator then discards the numbers, leaving `{pass0, pass1}`.

source/engine/graphics/api/vulkan/RenderGraph.cpp
```diff
@@ -126,103 +110,186 @@ auto CreateLitPass(const nc::graphics::Device* device, nc::graphics::GpuAllocato
...
-    for (auto i : std::views::iota(0u, m_maxLights))
+    auto* currentFrame = m_frameManager->CurrentFrameContext();
+    auto frameIndex = currentFrame->Index();
+
+    if (!m_isDescriptorSetLayoutsDirty.at(frameIndex))
     {
-        m_shadowMappingPasses.push_back(CreateShadowMappingPass(m_device, m_gpuAllocator, m_swapchain, m_dimensions, i));
+        return;
     }
+    m_shadowMappingPasses.at(frameIndex).clear();
```
McCallisterRomer: For each of these calls to `at`, I'd suggest doing it once and storing the result in a reference.

McCallisterRomer (commented):
Haven't made it all the way through yet. Looking good!

source/engine/graphics/api/vulkan/RenderGraph.cpp
```diff
@@ -126,103 +110,186 @@ auto CreateLitPass(const nc::graphics::Device* device, nc::graphics::GpuAllocato
...
+    OPTICK_CATEGORY("RenderGraph::SinkPostProcessImages", Optick::Category::Rendering);
+
+    auto* currentFrame = m_frameManager->CurrentFrameContext();
+    auto frameIndex = currentFrame->Index();
+    auto view = m_dummyShadowMap.view.get();
+    m_postProcessImageViews.at(PostProcessImageType::ShadowMap).perFrameViews.at(frameIndex) = std::vector<vk::ImageView>(m_maxLights, view);
+
+    for (auto i : std::views::iota(0u, m_activeShadowMappingPasses))
+    {
+        m_postProcessImageViews.at(PostProcessImageType::ShadowMap).perFrameViews.at(frameIndex).at(i) = m_shadowMappingPasses.at(frameIndex).at(i)->GetAttachmentView(0u);
```
McCallisterRomer: This does the map and frameIndex based lookup each iteration, but they don't change.

source/engine/graphics/api/vulkan/RenderGraph.cpp
```diff
@@ -126,103 +110,186 @@ auto CreateLitPass(const nc::graphics::Device* device, nc::graphics::GpuAllocato
...
+    if (!m_isDescriptorSetLayoutsDirty.at(frameIndex))
     {
-        m_shadowMappingPasses.push_back(CreateShadowMappingPass(m_device, m_gpuAllocator, m_swapchain, m_dimensions, i));
+        return;
     }
+    m_shadowMappingPasses.at(frameIndex).clear();
+
+    for (auto i : std::views::iota(0u, m_activeShadowMappingPasses))
+    {
+        m_shadowMappingPasses.at(frameIndex).push_back(CreateShadowMappingPass(m_device, m_gpuAllocator, m_swapchain, m_dimensions, i, frameIndex));
```
McCallisterRomer: Same thing, `m_shadowMappingPasses.at(frameIndex)` isn't dependent on the iteration.

source/engine/graphics/api/vulkan/RenderGraph.cpp
```diff
@@ -126,103 +110,186 @@ auto CreateLitPass(const nc::graphics::Device* device, nc::graphics::GpuAllocato
...
+        }
+    });
+}
+
+void RenderGraph::SinkPostProcessImages()
+{
+    OPTICK_CATEGORY("RenderGraph::SinkPostProcessImages", Optick::Category::Rendering);
+
+    auto* currentFrame = m_frameManager->CurrentFrameContext();
+    auto frameIndex = currentFrame->Index();
```
McCallisterRomer: Every `frameIndex` variable should be const

source/engine/graphics/api/vulkan/RenderGraph.cpp
```diff
@@ -126,103 +110,186 @@ auto CreateLitPass(const nc::graphics::Device* device, nc::graphics::GpuAllocato
...
-    cmd->end();
+    m_litPass.at(frameIndex)->Begin(cmd, frameBufferIndex);
+    m_litPass.at(frameIndex)->Execute(cmd, frameData, frameIndex);
+    m_litPass.at(frameIndex)->End(cmd);
 }
 
 void RenderGraph::Resize(const Vector2& dimensions)
 {
-    m_litPass = CreateLitPass(m_device, m_gpuAllocator, m_swapchain, m_descriptorSets, dimensions);
+    for (auto i : std::views::iota(0u, MaxFramesInFlight))
```
McCallisterRomer: There are many cases in this file where `iota` is used instead of what could be a range based for/`for_each`. Also, for this function, the data traversal looks very scattered. Why not do all lit pass updating, then move on to the next thing?

source/engine/graphics/api/vulkan/RenderGraph.cpp
```diff
@@ -126,103 +110,186 @@ auto CreateLitPass(const nc::graphics::Device* device, nc::graphics::GpuAllocato
...
-        m_shadowMappingPasses.push_back(std::move(CreateShadowMappingPass(m_device, m_gpuAllocator, m_swapchain, m_dimensions, i)));
+        for (auto j : std::views::iota(0u, m_activeShadowMappingPasses))
+        {
+            m_shadowMappingPasses.at(i).push_back(CreateShadowMappingPass(m_device, m_gpuAllocator, m_swapchain, m_dimensions, j, i));
+            m_postProcessImageViews.at(PostProcessImageType::ShadowMap).perFrameViews.at(i).emplace_back(m_shadowMappingPasses.at(i).back()->GetAttachmentView(0u));
+        }
     }
 
-    for (auto i : std::views::iota(0u, m_activeShadowMappingPasses))
+    for (auto i : std::views::iota(0u, MaxFramesInFlight))
```
McCallisterRomer: Use range based loops for everything below here

source/engine/graphics/api/vulkan/ShaderStorage.cpp
```diff
@@ -0,0 +1,589 @@
...
+{
+template<typename T>
+void RemoveAt(std::vector<T>& collection, uint32_t index)
+{
+    collection.at(index) = std::move(collection.back());
+    collection.pop_back();
+}
+
+template<typename T, typename U>
+void RemoveAt(std::vector<T>& indexCollection, T index, std::vector<U>& inSyncCollection, std::string notFound)
```
McCallisterRomer: Why create a string for a failure message? Just have the function return false instead of throwing and have the caller handle it. No need to allocate on every call.

source/engine/graphics/api/vulkan/ShaderStorage.cpp
```diff
@@ -0,0 +1,589 @@
...
+                GetStageFlags(eventData.stage),
+                vk::DescriptorBindingFlagBitsEXT::ePartiallyBound
+            );
+            break;
+        }
+        case CabUpdateAction::Add:
+        {
+            OPTICK_CATEGORY("CabUpdateAction::Add", Optick::Category::Rendering);
+
+            auto& sampler = storage.buffers.at(eventData.uid)->sampler.get();
```
McCallisterRomer: call `at()` once and store

source/engine/graphics/api/vulkan/RenderGraph.cpp
```diff
@@ -126,103 +110,186 @@ auto CreateLitPass(const nc::graphics::Device* device, nc::graphics::GpuAllocato
...
+    auto view = m_dummyShadowMap.view.get();
+    m_postProcessImageViews.at(PostProcessImageType::ShadowMap).perFrameViews.at(frameIndex) = std::vector<vk::ImageView>(m_maxLights, view);
+
+    for (auto i : std::views::iota(0u, m_activeShadowMappingPasses))
+    {
+        m_postProcessImageViews.at(PostProcessImageType::ShadowMap).perFrameViews.at(frameIndex).at(i) = m_shadowMappingPasses.at(frameIndex).at(i)->GetAttachmentView(0u);
+    }
+}
+
+auto RenderGraph::GetPostProcessImages(PostProcessImageType imageType) -> std::vector<vk::ImageView>
```
McCallisterRomer: Return by const&

source/engine/graphics/api/vulkan/ShaderStorage.cpp
```diff
@@ -0,0 +1,589 @@
...
+            );
+
+            auto& sampler = storage.buffers.at(eventData.imageType)->sampler.get();
+            auto& views = storage.buffers.at(eventData.imageType)->views;
+            auto& imageInfos = storage.buffers.at(eventData.imageType)->imageInfos;
+
+            views.clear();
+            imageInfos.clear();
+
+            auto postProcessViews = m_renderGraph->GetPostProcessImages(eventData.imageType);
```
McCallisterRomer: `const auto&` if `GetPostProcessImages()` is updated to return by const&

source/engine/graphics/api/vulkan/ShaderStorage.cpp
```diff
@@ -0,0 +1,589 @@
...
+            OPTICK_CATEGORY("PpiaUpdateAction::Update", Optick::Category::Rendering);
+            
+            auto& sampler = storage.buffers.at(eventData.imageType)->sampler.get();
+            auto& views = storage.buffers.at(eventData.imageType)->views;
+            auto& imageInfos = storage.buffers.at(eventData.imageType)->imageInfos;
+
+            views.clear();
+            imageInfos.clear();
+
+            auto postProcessViews = m_renderGraph->GetPostProcessImages(eventData.imageType);
```
McCallisterRomer: `const auto&`

source/engine/graphics/api/vulkan/ShaderStorage.cpp
```diff
@@ -0,0 +1,589 @@
...
+                eventData.slot,
+                eventData.currentFrameIndex
+            );
+            break;
+        }
+        case PpiaUpdateAction::Clear:
+        {
+            OPTICK_CATEGORY("PpiaUpdateAction::Clear", Optick::Category::Rendering);
+
+            for (auto i : std::views::iota(0u, MaxFramesInFlight))
```
McCallisterRomer: Would also be better as range-based for

source/engine/graphics/api/vulkan/ShaderStorage.cpp
```diff
@@ -0,0 +1,589 @@
...
+                {
+                    throw nc::NcError("Attempted to clear a Storage Buffer that doesn't exist.");
+                }
+
+                auto posIndex = static_cast<uint32_t>(std::distance(m_staticSsboStorage.uids.begin(), pos));
+                m_staticSsboStorage.buffers.at(posIndex)->Clear();
+                break;
+            }
+
+            for (auto i : std::views::iota(0u, MaxFramesInFlight))
```
McCallisterRomer: range-based for

source/engine/graphics/api/vulkan/ShaderStorage.cpp
```diff
@@ -0,0 +1,589 @@
...
+                {
+                    throw nc::NcError("Attempted to clear a Uniform Buffer that doesn't exist.");
+                }
+
+                auto posIndex = static_cast<uint32_t>(std::distance(m_staticUboStorage.uids.begin(), pos));
+                m_staticUboStorage.buffers.at(posIndex)->Clear();
+                break;
+            }
+
+            for (auto i : std::views::iota(0u, MaxFramesInFlight))
```
McCallisterRomer: another one

source/engine/graphics/shader_resource/ShaderResourceBus.h
```diff
@@ -1,21 +1,52 @@
...
 
 /** @brief Data transfer channels for shader resources. Signals are emitted from frontend graphics
  *         systems to be picked up by API-specific backend handlers. */
 struct ShaderResourceBus
 {
-    Signal<const EnvironmentData&> environmentChannel;
-    Signal<const std::vector<ObjectData>&> objectChannel;
-    Signal<const std::vector<PointLightData>&> pointLightChannel;
-    Signal<const std::vector<SkeletalAnimationData>&> skeletalAnimationChannel;
+    inline static uint32_t CubeMapArrayBufferUid = 0u;
```
McCallisterRomer: Why are these static?

source/engine/graphics/shader_resource/ResourceInstances.cpp
```diff
@@ -0,0 +1,99 @@
...
+#include "graphics/shader_resource/ShaderResourceBus.h"
+
+namespace nc::graphics
+{
+AssetResourcesConfig::AssetResourcesConfig(const config::MemorySettings& memorySettings)
+    : maxTextures{memorySettings.maxTextures},
+      maxCubeMaps{memorySettings.maxCubeMaps}
+{}
+
+AssetResources::AssetResources(AssetResourcesConfig config, ShaderResourceBus* resourceBus, ModuleProvider modules)
```
McCallisterRomer: In 'principle of least privelege' fashion, this should maybe just take `NcAsset`.

source/engine/graphics/api/vulkan/ShaderBindingManager.h
```diff
@@ -0,0 +1,78 @@
...
+
+#include <span>
+#include <unordered_map>
+
+namespace nc::graphics
+{
+using SetIndex = uint32_t;
+using BindingSlot = uint32_t;
+
+struct DescriptorSetLayout
```
McCallisterRomer, line 20: We should hash out a data layout for these that's more in line with the access pattern. `bindings` and `bindingFlags` are accessed at the same time by the same key, but stored in different maps. Same with `writesPerSet` and `isDirty` from `DescriptorSets`. There's some hoop jumping to get contiguous views over some of the data.

source/engine/graphics/system/EnvironmentSystem.h
```diff
@@ -1,12 +1,28 @@
...
 {
 struct CameraState;
 
+#ifdef _MSC_VER
+#pragma warning( disable : 4324 ) // Justification: We want to be explicit about alignment for the GPU buffer structs.
+#endif
+struct EnvironmentData
+{
+    alignas(16)Vector4 cameraWorldPosition = Vector4(0.0f, 4.0f, -6.4f, 0.0f);
+    alignas(16)uint32_t skyboxTextureIndex = std::numeric_limits<uint32_t>::max();
```
McCallisterRomer, line 20: Holy cow, do all scalar types need 16 byte alignment?

source/engine/graphics/GraphicsConstants.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
```
McCallisterRomer, line 1: Love this

source/engine/graphics/api/vulkan/buffers/CubeMapArrayBuffer.h
jaremieromer: Oh, we actually no longer use this field

jaremieromer: Removed

source/engine/graphics/api/vulkan/buffers/StorageBuffer.cpp
jaremieromer: Done

source/engine/graphics/api/vulkan/buffers/UniformBuffer.cpp
jaremieromer: Done

source/engine/graphics/api/vulkan/ShaderBindingManager.h
jaremieromer: Done

source/engine/graphics/api/vulkan/FrameManager.cpp
jaremieromer: Good to know! fixed

source/engine/graphics/api/vulkan/RenderGraph.h
jaremieromer: Nice, thanks. Yes - that's exactly why.

source/engine/graphics/api/vulkan/RenderGraph.cpp
jaremieromer: done

source/engine/graphics/api/vulkan/RenderGraph.cpp
jaremieromer: Fixed

source/engine/graphics/api/vulkan/RenderGraph.cpp
jaremieromer: fixed

source/engine/graphics/api/vulkan/RenderGraph.cpp
jaremieromer: fixed

source/engine/graphics/api/vulkan/RenderGraph.cpp
jaremieromer: Cleaned up a lot - although we can't do a range based for for the shadowpasses - we actually clear the vector and then recreate based on the count of m_activeShadowMappingPasses. However, I was able to clean this up a lot.

source/engine/graphics/api/vulkan/RenderGraph.cpp
jaremieromer: See above comment, I've cleaned up a lot though can't use a range based for for the shadowpasses

source/engine/graphics/api/vulkan/ShaderStorage.cpp
jaremieromer: Fixed

source/engine/graphics/api/vulkan/ShaderStorage.cpp
jaremieromer: fixed

source/engine/graphics/api/vulkan/RenderGraph.cpp
jaremieromer: done

source/engine/graphics/api/vulkan/ShaderStorage.cpp
jaremieromer: done

source/engine/graphics/api/vulkan/ShaderStorage.cpp
jaremieromer: done

source/engine/graphics/api/vulkan/ShaderStorage.cpp
jaremieromer: done

source/engine/graphics/api/vulkan/ShaderStorage.cpp
jaremieromer: done

source/engine/graphics/api/vulkan/ShaderStorage.cpp
jaremieromer: done

source/engine/graphics/shader_resource/ShaderResourceBus.h
jaremieromer: Ah - they don't have to be anymore. Holdover from one of the earlier designs

source/engine/graphics/shader_resource/ResourceInstances.cpp
jaremieromer: done

source/engine/graphics/system/EnvironmentSystem.h
jaremieromer: No - https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap15.html#interfaces-resources-layout
I'm still shaky on alignment, but I do know that we should avoid Vector3 and always use Vector4. Since Vector4 has an alignment of 16 - I thought that all members should have an alignment of 16.

McCallisterRomer: Ok, should be fine to give it extra alignment here, but FYI, both `nc::Vector3` and `nc::Vector4` have 4 byte alignment, not 16. `DirectX::XMVECTOR` is 16, though. Unresolving just so you see this, in case that matters anywhere.

McCallisterRomer (approved):
Awesome! I still have a validation layer problems, but everything else looks good. Feel free to use my PC for testing.

VUID-vkAcquireNextImageKHR-semaphore-01779(ERROR / SPEC): msgNum: 1461184347 - Validation Error: [ VUID-vkAcquireNextImageKHR-semaphore-01779 ] Object 0: handle = 0xee647e0000000009, type = VK_OBJECT_TYPE_SEMAPHORE; | MessageID = 0x5717e75b | vkAcquireNextImageKHR():  Semaphore must not have any pending operations. The Vulkan spec states: If semaphore is not VK_NULL_HANDLE it must not have any uncompleted signal or wait operations pending (https://vulkan.lunarg.com/doc/view/1.3.275.0/windows/1.3-extensions/vkspec.html#VUID-vkAcquireNextImageKHR-semaphore-01779)
    Objects: 1
        [0] 0xee647e0000000009, type: 5, name: NULL
VUID-vkAcquireNextImageKHR-semaphore-01779(ERROR / SPEC): msgNum: 1461184347 - Validation Error: [ VUID-vkAcquireNextImageKHR-semaphore-01779 ] Object 0: handle = 0xcad092000000000d, type = VK_OBJECT_TYPE_SEMAPHORE; | MessageID = 0x5717e75b | vkAcquireNextImageKHR():  Semaphore must not have any pending operations. The Vulkan spec states: If semaphore is not VK_NULL_HANDLE it must not have any uncompleted signal or wait operations pending (https://vulkan.lunarg.com/doc/view/1.3.275.0/windows/1.3-extensions/vkspec.html#VUID-vkAcquireNextImageKHR-semaphore-01779)
    Objects: 1
        [0] 0xcad092000000000d, type: 5, name: NULL

PR #623, opened 2024-03-03
Configurable wireframe colors

### LINKED ISSUE #573: Configurable Wireframe Colors
Add parameter to `WireframeRenderer` and underlying shader for configuring the wireframe color. Immediate use case is to more easily differentiate the selected `Entity` and `Collider`, but would probably come in handy in other scenarios, like physics debug rendering.

McCallisterRomer:
resolve #573

Also disabled wide lines feature since it isn't supported on molten vk/swiftshader.


jaremieromer (approved):
Really nice work!

PR #632, opened 2024-03-06
Vulkan Hardware Requirements Info

### LINKED ISSUE #627: Create Vulkan hardware minimum requirements document
We need to document (either in `VulkanConstants.h` or in our documentation) our hardware requirements for the GPU. 
Using Sascha Willems' gpuinfo.org, we need to come up with an accepted target Vulkan feature adoption rate for GPUs/drivers and not include features if they fall outside of that rate.
Also using gpuinfo.org, we need to find a good maximum limit for various Vulkan constructs (eg: 95% of drivers support n or more descriptor sets) and document them.

jaremieromer:
resolves #627


McCallisterRomer (approved):
NICE!

PR #638, opened 2024-03-18
Position/Freedom Constraints

### LINKED ISSUE #581: Constraint for Linear/Angular Restriction
Linear/Angular freedom is managed inside `PhysicsBody`, requiring a vector multiplication on each velocity write. Most objects don't use this property. Instead make a constraint that the solver can operate on (after other velocity-based constraints), and create it only when a body needs one.

McCallisterRomer:
resolves #581

- Factoring handling of degrees of freedom out of `PhysicsBody` into a `VelocityRestriction` component + constraint.
- Adding a `PositionClamp` component + constraint that is similar to linear restriction, but with a soft/springy response.
- Fixed incorrect `static_cast` bug in `PhysicsBody` factory (I thought I already did this, but maybe that branch was thrown away??)
- Added `Transform::RotateAround()`
- Updated `PhysicsTest` scene to use new components


include/ncengine/ecs/Transform.h
```diff
@@ -145,6 +145,9 @@ class Transform final : public ComponentBase
         /** @brief Apply a rotation about an axis to local rotation */
         void Rotate(const Vector3& axis, float radians);
 
+        /** @brief Apply a rotation about an axis passing through a point. */
```
jaremieromer, line 148: You might not have figured out exactly the 'how' yet in how the DX Math functions get this to work but - having a Vector3 as an axis doesn't seem possible and makes this method confusing. Can you add info on what is happening under the hood (if you have it?)

source/engine/physics/PhysicsBody.cpp
```diff
@@ -122,7 +108,7 @@ void PhysicsBody::ApplyImpulse(const Vector3& impulse)
 void PhysicsBody::ApplyImpulse(DirectX::FXMVECTOR impulse)
 {
     NC_PHYSICS_ASSERT(!m_properties.isKinematic, "Attempting to move immovable object");
```
jaremieromer: This is a nit, and I didn't notice it the first time around - but will you change these error messages to "Cannot move an immovable object" - I don't like "Attempting" here because it could be construed as: "we failed somewhere along the attempt to move an immovable object"

source/engine/physics/PhysicsPipelineBuilder.h
```diff
@@ -140,12 +140,21 @@ auto BuildMergeContactsTask(task::ExceptionContext& ctx, NarrowPhase& narrow, Co
...
+    });
+}
+
+inline auto BuildGenerateContactConstraintsTask(task::ExceptionContext& ctx, Solver& solver, NarrowPhase& narrow)
 {
     return task::Guard(ctx, [&solver, &narrow]
     {
-        OPTICK_CATEGORY("Solver::GenerateConstraints", Optick::Category::Physics);
-        solver.GenerateConstraints(narrow.Manifolds());
+        OPTICK_CATEGORY("Solver::GenerateFreedomConstraints", Optick::Category::Physics);
```
jaremieromer: Copy-paste typo from previous method, should be GenerateContactConstraints

jaremieromer (approved):
Whoah. Physics is very hairy. Nice work!

source/engine/physics/PhysicsBody.cpp
McCallisterRomer: Yes, fixed

PR #644, opened 2024-03-24
Safer buffer handles

McCallisterRomer:
Proposing changes to `StorageBufferHandle`/`UniformBufferHandle` that (IMO) can catch some potential bugs and are a bit clearer.
- The `Bind()`/`Update()` methods and the event data transmitted for buffer writes used non-const pointers, despite the fact that these are just sources for `memcpy`. I think we should be more explicit about the fact that the input data won't be modified. The event receiver doesn't know anything about the underlying data anyway, so it should never be modifying it.
- The interface for these can take a range, rather than a pointer/size, and pass along the pointer/size from the range to the implementation, minimizing bugs at the call site (maybe you have two vectors, and you accidentally call data on one and size on another - I have done this sort of thing before).
- The fact that there is a `memcpy` happening behind the scenes isn't explicit (arguably you could intuit that its happening), and nobody is verifying the type is safe to copy. In practice, the involved types are likely to be trivially copyable, but we can guarantee that at compile-time.
- The per-frame and static implementations can be combined pretty simply.


source/engine/graphics/shader_resource/StorageBufferHandle.h
```diff
@@ -33,8 +34,14 @@ class StorageBufferHandle
 {
     public:
         StorageBufferHandle(uint32_t uid, size_t size, shader_stage stage, Signal<const SsboUpdateEventData&>* backendPort, uint32_t slot, uint32_t set = 0u);
-        void Bind(void* data, size_t size, uint32_t currentFrameIndex);
-        void Bind(void* data, size_t size);
+
+        template<TriviallyCopyableRange Rng>
+        void Bind(const Rng& data, uint32_t currentFrameIndex = UINT32_MAX)
+        {
+            using T = std::remove_cvref_t<std::ranges::range_value_t<Rng>>;
```
jaremieromer, line 41: So nice!

jaremieromer (approved):
This is a really good technique to know. Thanks for doing this

PR #645, opened 2024-03-25
Particle instancing and alpha sorting

### LINKED ISSUE #619: Particle instancing
I thought we were already instancing particles. We're not. Do it.

McCallisterRomer:
Requires a ~1 line change once #644 is merged.

resolves #619 

### Adding instancing for particles
I followed the existing convention of preallocating based on a config value, however I didn't really want to throw if we exceed that size. Instead, I'm just not rendering particles beyond the specified maximum - might be worth revisiting in the future. It would be nice to be able to reallocate the buffer for the current frame without blocking on other frames in flight, but I'm not sure we have that capability.

### Adding approximate sorting for particles
I tried doing a full sort, but its too expensive. We either need to have more material options so we can skip sorting some things or move everything to the gpu. I'm down to dig into either.

Instead, we sort each system back to front based on its position the last time it emitted any particles. There are cases where this gets weird, but is good enough a lot of the time. Thanks for the suggestion! 

I also disabled depth writes in `ParticleTechnique` so particles don't occlude other particles. This greatly hides the fact that the sort isn't perfect. Some particles can still be rendered in the wrong order, but you don't get giant, blocky artifacts.


source/engine/graphics/system/ParticleEmitterSystem.cpp
```diff
@@ -1,129 +1,210 @@
 #include "ParticleEmitterSystem.h"
+#include "assets/AssetService.h"
```
McCallisterRomer: Sorry, my unindenting kind of hides relevant changes. Other than the new constructor member, changes should be in `UpdateParticles()`, `SortEmitters()`, and `Execute()`.

source/engine/graphics/NcGraphicsImpl.cpp
```diff
@@ -229,7 +230,7 @@ namespace nc::graphics
             std::move(objectState),
             std::move(lightingState),
             std::move(widgetState),
-            m_systemResources.particleEmitters.GetParticles()
+            std::move(particleState)
```
jaremieromer, line 233: Hooray!

source/engine/graphics/NcGraphicsImpl.cpp
```diff
@@ -203,6 +203,7 @@ namespace nc::graphics
         auto objectState = m_systemResources.objects.Execute(currentFrameIndex, MultiView<MeshRenderer, Transform>{m_registry}, MultiView<ToonRenderer, Transform>{m_registry},
                                                                         cameraState, environmentState, skeletalAnimationState);
         auto lightingState = m_systemResources.pointLights.Execute(currentFrameIndex, MultiView<PointLight, Transform>{m_registry});
+        auto particleState = m_systemResources.particleEmitters.Execute(currentFrameIndex);
```
jaremieromer, line 206: Hooray!

source/engine/graphics/PerFrameRenderState.h
```diff
@@ -18,6 +18,6 @@ struct PerFrameRenderState
     ObjectState objectState;
     LightingState lightingState;
     WidgetState widgetState;
-    std::span<const nc::particle::EmitterState> emitterStates;
+    ParticleState particleState;
```
jaremieromer, line 21: Hooray!

source/engine/graphics/api/vulkan/techniques/ParticleTechnique.cpp
```diff
@@ -105,28 +106,15 @@ namespace nc::graphics::vulkan
...
-            for (; index.Valid(); ++index)
-            {
-                pushConstants.model = models[index];
-                pushConstants.baseColorIndex = AssetService<TextureView>::Get()->Acquire(emitterState.GetInfo().init.particleTexturePath).index;
-
-                cmd->pushConstants(m_pipelineLayout.get(), vk::ShaderStageFlagBits::eFragment | vk::ShaderStageFlagBits::eVertex, 0, sizeof(ParticlePushConstants), &pushConstants);
-                cmd->drawIndexed(meshAccessor.indexCount, 1, meshAccessor.firstIndex, meshAccessor.firstVertex, 0); // indexCount, instanceCount, firstIndex, vertexOffset, firstInstance
-            }
-        }
+        const auto& meshAccessor = frameData.particleState.mesh;
```
jaremieromer, line 115: Much cleaner

jaremieromer (approved):
Really nice work!

PR #659, opened 2024-05-16
Enhancements for the PBR shader

jaremieromer:
Note: Most of the files are just assets. The important stuff is in Pbr.frag and shaders.ps1. The rest of the non-asset changes are removing the now-unneeded elements of Point Light. 

Recreated the PBR shader with proper D, G, F functions. 
Changed diffuse intensity in the point light to radius.
Added a parameter to the `shaders.ps1` script to allow for building the shaders in debug mode.
Added a bunch of sphere materials to demonstrate various metallic, roughness and albedo combinations.

Before:
image

After:
image


sample/source/scenes/GraphicsTest.cpp
```diff
@@ -46,14 +46,46 @@ void GraphicsTest::Load(ecs::Ecs world, ModuleProvider modules)
         "skeleton\\BaseColor.nca",
         "skeleton\\Roughness.nca",
         "skeleton\\Metallic.nca",
+        "denim\\BaseColor.nca",
```
McCallisterRomer: For linux to be able to run the sample, we'll have to start using '/' everywhere. I've fixed the existing ones in an in-progress branch. Depending on merge order, one of us will have to change these new ones.

McCallisterRomer (approved):
Slick!

sample/source/scenes/GraphicsTest.cpp
jaremieromer: I'll do it now

PR #664, opened 2024-05-20
Add smoke test to CI

### LINKED ISSUE #252: Automated Integration Tests

### LINKED ISSUE #601: Investigate SwiftShader for CI GFX Testing

McCallisterRomer:
resolves #252, resolves #601

Adding swiftshader + smoke test to windows CI builds. Also includes the remaining changes to make swiftshader (mostly) happy. It will still spit out a validation error due to push constant size, but it handles it fine.


jaremieromer (approved):
One step closer!

PR #665, opened 2024-05-20
Run smoke test on linux

### LINKED ISSUE #663: Run smoke test on linux
Remove the conditional checks in CI that gate the smoke test to windows builds. This requires being able to run the sample on linux (fix config paths, and I think something else with paths/directory iterators).

McCallisterRomer:
resolves #663 

Adding smoke test to linux CI builds. Also fixing remaining path separators in the sample for linux.


script/run_smoke_test.sh
```diff
@@ -15,23 +15,34 @@ echo "OUT_FILE: $OUT_FILE"
...
+  exit 1
+fi
+
+cat <<EOF > vk_swiftshader_icd.json
 {
   "file_format_version": "1.0.0",
   "ICD": {
-    "library_path": "./vk_swiftshader.dll",
+    "library_path": "./$SWIFTSHADER_LIB",
     "api_version": "1.3"
```
jaremieromer: Can we parameterize this? This will be easy to miss in the future

jaremieromer (approved):

script/run_smoke_test.sh
jaremieromer: Referring to the api_version

PR #667, opened 2024-05-22
Add NcWindow Module

### LINKED ISSUE #340: Move window and surface
The surface doesn't currently have a good home. It might make the most sense to keep this as part of the window, but the window lives one level above graphics. Future cross-plat stuff will probably look something like: window takes the instance -> creates window -> creates surface...

The window should be moved into the graphics layer, and move surface into the window. As the window also handles input, this is probably a larger effort, and there is probably other work to get this working correctly.

### LINKED ISSUE #363: Don't include module implementations in engine

McCallisterRomer:
resolves #363, resolves #340 

- Adding `NcWindow` so that `NcGraphics` can take a public interface instead of `WindowImpl` (now named `NcWindowImpl` for consistency). This also allowed moving the `BuildGraphicsModule()` factory to the public API.
- Moved the `BuildPhysicsModule()` factory to the public API.
- Updated window resize events to be handled by a `Signal` instead of a single callback to graphics. `IOnResizeReceiver` can then be updated/replaced by the `Signal` going forward, but for now, I just marked the registration functions as deprecated.
- Replaced the quit callback in `WindowImpl` with a `Signal` located `SystemEvents`.


include/ncengine/window/Window.h
```diff
@@ -19,9 +32,57 @@ auto GetScreenExtent() -> Vector2;
...
+/**
+ * @brief Allow an object to receive window resize events. Receivers must be unregistered before they are destroyed.
+ * @deprecated Prefer subscribing to NcWindow::OnResize().
+ */
 void RegisterOnResizeReceiver(IOnResizeReceiver* receiver);
 
-/** @brief Unregister an object from receiving window resize events. */
+/**
+ * @brief Unregister an object from receiving window resize events.
+ * @deprecated Prefer subscribing to NcWindow::OnResize().
```
jaremieromer: This comment should be updated to reflect unsubscribing

include/ncengine/window/Window.h
```diff
@@ -19,9 +32,57 @@ auto GetScreenExtent() -> Vector2;
...
+        explicit NcWindow() noexcept
+            : Module{NcWindowId} {}
+
+        /** @brief Get the dimensions of the entire window. */
+        auto GetDimensions() const noexcept -> const Vector2& { return m_dimensions; }
+
+        /** @brief Get the dimensions of the renderable window region. */
+        auto GetScreenExtent() const noexcept -> const Vector2& { return m_screenExtent; }
+
+        /** @brief Get the ration between the current DPI and the platform's default DPI. */
```
jaremieromer: *ratio

source/engine/engine/ModuleFactory.cpp
```diff
@@ -7,14 +7,13 @@
...
 #include "ncengine/ecs/NcEcs.h"
 #include "ncengine/ecs/Registry.h"
+#include "ncengine/graphics/NcGraphics.h"
+#include "ncengine/math/Random.h"
 #include "ncengine/module/ModuleRegistry.h"
+#include "ncengine/physics/NcPhysics.h"
 #include "ncengine/scene/NcScene.h"
 #include "ncengine/utility/Log.h"
 
-/** @todo #363 Move factories to public headers and include those instead. */
```
jaremieromer, line 14: Nice

jaremieromer (approved):

Issue #668, opened 2024-05-25
Fail CI on validation error in smoke test


McCallisterRomer:
Running the smoke test should fail CI on any validation layer output.

PR #669, opened 2024-05-25
Validation Layers in CI and VulkanCI Integration

### LINKED ISSUE #662: Build validation layers in CI
Build validation layers for running the smoke test in CI. These probably must be built ourselves rather than installing the whole SDK since we have to build the loader with a special flag.

~~Either as part of this issue, or a separate one, running the smoke test should fail on any validation layer output - one of the techniques uses large push constants, which swiftshahder will be unhappy with.~~

McCallisterRomer:
resolves #662 

Created [VulkanCI](https://github.com/NcStudios/VulkanCI) repo to house all CI-related Vulkan components. This will help minimize build times (specifically for the validation layers). Updating the setup-vulkan action to target a release from here.

Enabling validation layers in the smoke test. We still have errors due to push constant size, so we don't want to force failing the workflow yet. I've created #668 to cover this part of the work.


jaremieromer (approved):
This is fantastic, excellent job!

PR #671, opened 2024-05-26
Move View Projection to Environment Uniform Buffer

### LINKED ISSUE #358: Duplication of view projection across ObjectData instances
The `ObjectData::viewProjection` member is the same for every instance within a given frame. Refactor so that this is computed/set/mapped once per frame, instead of copying it for every object rendered.

McCallisterRomer:
resolves #358, unblocks #668

Adding the camera's view projection matrix to the environment uniform buffer so we can:
1. Drop the size of wireframe push constants below the guaranteed minimum.
2. Have one instance of the matrix instead of duplicating across each `ObjectData` instance.


resources/shaders/Pbr.frag
```diff
@@ -37,7 +37,6 @@ struct ObjectData
     // N MVP matrices
     mat4 model;
     mat4 modelView;
```
jaremieromer, line 39: I love seeing less data here! 
You wouldn't know this, but we also don't need modelView here. I'll create an issue to remove it as well.

resources/shaders/Toon.frag
```diff
@@ -40,14 +39,15 @@ layout (std140, set=0, binding=1) readonly buffer PointLightsArray
...
 layout (set = 0, binding = 5) uniform EnvironmentDataBuffer
 {
+    mat4 cameraViewProjection;
     vec4 cameraWorldPosition;
     int skyboxCubemapIndex;
 } environmentData;
 
+layout (set = 1, binding = 2) uniform sampler2D textures[];
+layout (set = 1, binding = 4) uniform samplerCube cubeMaps[];
 layout (location = 0) in vec3 inFragPosition;
```
jaremieromer, line 52: I like having a newline between the descriptor resources and the resources that come from the vertex shader.

resources/shaders/Wireframe.frag
```diff
@@ -4,7 +4,7 @@
 
 layout(push_constant) uniform PER_OBJECT
 {
-    layout(offset = 128) vec4 color;
+    layout(offset = 64) vec4 color;
```
jaremieromer, line 7: I was wondering if you were gonna remember to do this! nice job

jaremieromer (approved):
NICE!

PR #672, opened 2024-05-26
Fail Smoke Test On Validation Errors

### LINKED ISSUE #668: Fail CI on validation error in smoke test

McCallisterRomer:
resolves #668

Forcing smoke test to fail if there is any validation layer output. I also moved all smoke test files to `test/smoke_test` to keep the sample cleaner.


test/smoke_test/vk_layer_settings.txt
jaremieromer, line 1: Can you group these by whether or not the option is commented out? That will make it easier to read and edit later

jaremieromer (approved):

PR #676, opened 2024-05-31
Use VulkanCI

McCallisterRomer:
Replacing setup-vulkan action with VulkanCI


jaremieromer (approved):
This is SO awesome!! Nice work!!

PR #678, opened 2024-06-01
Added a SpotLight component. Swapped out the toon shader's overlay for outline width.

### LINKED ISSUE #217: Add spotlights

jaremieromer:
Added a new component `SpotLight.h`.
A spot light casts light in a single direction with a radius for attenuation and an inner and outer angle to specify both cone size and falloff.
In adding the spot light to the Toon Shader, I decided to fix the outline (in a separate PR now merged into this one). Because we are still restricted on our four material parameters per shader, I decided to nix the overlay parameter in favor of outline width because we can represent overlay in the base color texture. (We now no longer overload the hatchingTiling parameter to represent outline width).
The rest of the changes are largely boiler plate component add stuff.
closes #217
image

image


include/ncengine/graphics/SpotLight.h
```diff
@@ -0,0 +1,44 @@
...
+#pragma once
+
+#include "ncengine/ecs/Component.h"
+
+#include "ncmath/Vector.h"
+
+namespace nc::graphics
+{
+/** @brief Component representing a light pointing in a single direction with a radius for attenuation, and an inner and outer angle to specify both cone size and falloff. */
+struct SpotLight final : public ComponentBase
```
McCallisterRomer: Haven't looked myself yet, but is there any reason to derive from `ComponentBase`? IIRC, the only reason `PointLight` still does is for binary compatibility of the scene fragment in SS&DD.

include/ncengine/graphics/ToonRenderer.h
```diff
@@ -15,15 +15,15 @@ namespace nc::graphics
 struct ToonMaterial
 {
     std::string baseColor = asset::DefaultBaseColor;
-    std::string overlay = asset::DefaultBaseColor;
+    uint32_t outlineWidth = 2;
```
McCallisterRomer, line 18: This is a breaking change for serialization. We have a version number in the scene fragment, but, at least for now, no backwards compatible loading. Are we ok never updating SS&DD further? If so, this can stay, but the version in `SceneSerialization.h` needs to be bumped.

source/engine/graphics/api/vulkan/RenderGraph.cpp
```diff
@@ -164,13 +167,18 @@ void RenderGraph::SinkPostProcessImages()
...
     auto& shadowPasses = m_shadowMappingPasses[currentFrame->Index()];
 
-    for (auto i : std::views::iota(0u, m_activeShadowMappingPasses))
+    for (auto i : std::views::iota(0u, m_omniDirLightCount))
     {
-        perFrameShadowMapView[i] = shadowPasses[i].GetAttachmentView(0u);
+        perFrameShadowMapViews[i] = shadowPasses[i].GetAttachmentView(0u);
+    }
+
+    for (auto i : std::views::iota(m_omniDirLightCount, m_omniDirLightCount + m_uniDirLightCount))
```
McCallisterRomer, line 179: Can these two for loops be combined? To my eyes, its one sequential pass split into two steps.

source/engine/graphics/api/vulkan/RenderGraph.cpp
```diff
@@ -192,11 +200,18 @@ void RenderGraph::CommitResourceLayout()
     auto& perFramePasses =  m_shadowMappingPasses.at(frameIndex);
     perFramePasses.clear();
 
-    for (auto i : std::views::iota(0u, m_activeShadowMappingPasses))
+    for (auto i : std::views::iota(0u, m_omniDirLightCount))
```
McCallisterRomer, line 203: Definitely double check, but same?

source/engine/graphics/api/vulkan/RenderGraph.cpp
```diff
@@ -251,7 +266,13 @@ void RenderGraph::Resize(const Vector2& dimensions)
         shadowPasses.clear();
         shadowMapViews.clear();
 
-        for (auto shadowPassIndex : std::views::iota(0u, m_activeShadowMappingPasses))
+        for (auto shadowPassIndex : std::views::iota(0u, m_omniDirLightCount))
```
McCallisterRomer, line 269: ditto

source/engine/graphics/system/LightSystem.cpp
```diff
@@ -0,0 +1,105 @@
...
+    return state;
+}
+
+void LightSystem::Clear() noexcept
+{
+    m_pointLightData.clear();
+    m_pointLightBuffer.Clear();
+    m_spotLightData.clear();
+    m_spotLightBuffer.Clear();
+    for (auto i : std::views::iota(0u, MaxFramesInFlight))
```
McCallisterRomer: It was probably like this in the old file, but this just needs to be a range-based for over `m_syncedLightsCount`.

source/engine/graphics/system/LightSystem.h
```diff
@@ -0,0 +1,107 @@
...
+#include "DirectXMath.h"
+
+#include <vector>
+
+namespace nc::graphics
+{
+/** @brief Device-mapped properties of a PointLight. */
+struct PointLightData
+{
+    PointLightData() = default;
```
McCallisterRomer, line 22: Don't know if it was like this before, but a default constructor without default initializers on all primitives is unpredictable. You could probably just add a default for `radius` and remove both constructors since one is default and the other takes every member.

source/engine/graphics/system/LightSystem.h
```diff
@@ -0,0 +1,107 @@
...
+    Vector3 ambientColor;
+    int isInitialized = 1;
+    Vector3 diffuseColor;
+    float radius;
+};
+
+/** @brief Device-mapped properties of a SpotLight. */
+struct SpotLightData
+{
+    SpotLightData() = default;
```
McCallisterRomer, line 50: Same as above

source/engine/graphics/system/LightSystem.h
McCallisterRomer: Oh I see, its because of the `isInitialized` member. Still, if the default is here, `radius` needs an initializer.

source/engine/graphics/system/LightSystem.h
```diff
@@ -0,0 +1,107 @@
...
+    Vector3 direction = Vector3::Front();
+    float innerAngle;
+    float outerAngle;
+    float radius;
+};
+
+struct LightState
+{
+    std::vector<DirectX::XMMATRIX> viewProjections;
+    bool updateShadows;
```
McCallisterRomer: I'd feel better if this had a default too, even it is properly initialized on construction.

source/engine/serialize/ComponentSerialization.cpp
```diff
@@ -195,25 +195,26 @@ auto DeserializePhysicsMaterial(std::istream& stream, const DeserializationConte
     return physics::PhysicsMaterial{friction, restitution};
 }
 
-void SerializePointLight(std::ostream& stream, const graphics::PointLight& out, const SerializationContext& ctx, const std::any&)
+void SerializePointLight(std::ostream& stream, const graphics::PointLight& out, const SerializationContext&, const std::any&)
```
McCallisterRomer, line 198: Without the entity, serialize/deserialize can just be done directly for both `PointLight` and `SpotLight`:
```cpp
serialize::Serialize(stream, out);
```
and
```cpp
auto out = ...
serialize::Deserialize(stream, out);
return out;
```

McCallisterRomer (approved):

PR #680, opened 2024-06-16
RenderGraph Cleanup

jaremieromer:
Continuing work on cleaning up `RenderGraph`. 
This PR deals with collecting all the containers and operations that had to be duplicated for each frame in flight and merging them into a single `PerFrameRenderGraph` struct, so we were able to eliminate much of the code around selecting the right frame's data in each method.
The biggest functional change is how we create and bind pipelines to the render graph (for lit pass pipelines only, for now) now we create each `ITechnique` only once, and toggle it's `isActive` status to control whether or not it's executed.

A pipeline is now only executed if it has data to render. We track this in a new `PerFrameRenderStateData` struct that tracks counts each frame in `NcGraphicsImpl`

Because of the counts tracking, we no longer need any of the signals updating counts in `RenderGraph`

The other functional change is the control flow regarding sinking post process images (render pass outputs) has been inverted. Now `RenderGraph` sinks render pass outputs into `ShaderStorage`, as they are shader input. (Previously it stored them and `ShaderStorage` fetched them when needed.)

There was unfortunately no performance benefit to this cleanup - we hit the exact same benchmarks as prior to this change. (Measured on the loft PC).


source/engine/graphics/IGraphics.h
```diff
@@ -35,7 +36,7 @@ class IGraphics
         virtual ~IGraphics() = default;
 
         virtual auto BeginFrame() -> bool = 0;
-        virtual void CommitResourceLayout() = 0;
+        virtual void BuildRenderGraph(PerFrameRenderStateData state) = 0;
```
McCallisterRomer: const &

source/engine/graphics/api/vulkan/RenderGraph.cpp
```diff
@@ -95,237 +93,183 @@ auto CreateLitPass(const nc::graphics::vulkan::Device* device, nc::graphics::vul
...
-        perFrameShadowMapViews[i] = shadowPasses[i].GetAttachmentView(0u);
-    }
-}
+        return shadowPass.GetAttachmentView(0u);
+    });
 
-auto RenderGraph::GetPostProcessImages(PostProcessImageType imageType) -> const std::vector<vk::ImageView>&
-{
-    return m_postProcessImageViews.at(imageType).perFrameViews.at(m_frameManager->CurrentFrameContext()->Index());
+    m_shaderStorage->SinkPostProcessImages(shadowMapsSink, PostProcessImageType::ShadowMap, m_frameManager->CurrentFrameContext()->Index());
```
McCallisterRomer: Out of curiosity - `FrameManager` has a current index and the `PerFrameGpuContext` at that index also has an index. What's the difference between these?

source/engine/graphics/api/vulkan/RenderGraph.h
```diff
@@ -1,94 +1,65 @@
...
+        void Clear();
         void SinkPostProcessImages();
-        auto GetPostProcessImages(PostProcessImageType imageType) -> const std::vector<vk::ImageView>&;
-        auto GetLitPass() const noexcept -> const RenderPass& { return m_litPass.at(0u); };
-        void CommitResourceLayout();
-        void IncrementShadowPassCount(bool isOmniDirectional);
-        void DecrementShadowPassCount(bool isOmniDirectional);
-        void ClearShadowPasses() noexcept;
+        auto GetLitPass() const noexcept -> const RenderPass& { return m_perFrameRenderGraphs.at(0).litPass; };
+        void BuildRenderGraph(PerFrameRenderStateData stateData, uint32_t frameIndex);
```
McCallisterRomer: const &

source/engine/graphics/api/vulkan/ShaderStorage.cpp
```diff
@@ -517,4 +494,21 @@ void ShaderStorage::UpdateUniformBuffer(const UboUpdateEventData& eventData)
         }
     }
 }
+
+void ShaderStorage::SinkPostProcessImages(std::vector<vk::ImageView> postProcessImages, PostProcessImageType imageType, uint32_t frameIndex)
```
McCallisterRomer: `postProcessImages` should be const & or `std::span`.

source/engine/graphics/api/vulkan/VulkanGraphics.h
```diff
@@ -41,15 +42,13 @@ class VulkanGraphics : public IGraphics
...
                        asset::NcAsset* assetModule,
                        ShaderResourceBus& shaderResourceBus,
-                       uint32_t apiVersion, Registry* registry, 
-                       GLFWwindow* window, Vector2 dimensions, Vector2 screenExtent);
+                       uint32_t apiVersion, GLFWwindow* window, Vector2 dimensions, Vector2 screenExtent);
 
         ~VulkanGraphics() noexcept;
 
-        void CommitResourceLayout() override;
+        void BuildRenderGraph(PerFrameRenderStateData stateData) override;
```
McCallisterRomer: const &

source/engine/graphics/system/LightSystem.h
```diff
@@ -82,6 +82,8 @@ struct SpotLightData
 struct LightState
 {
     std::vector<DirectX::XMMATRIX> viewProjections = {};
+    uint32_t omniDirectionalLightCount;
```
McCallisterRomer: These should have defaults too

McCallisterRomer (approved):

source/engine/graphics/api/vulkan/RenderGraph.cpp
jaremieromer: Oops, I'd forgotten that I'd added that helper in FrameManager already. Thanks! No difference.

PR #681, opened 2024-06-19
Rendergraph refactor for sink views

jaremieromer:
### Note: Not actually as big as it seems. Most file changes are trickle down from the rename in item 5.) below.
More work on cleaning up `RenderGraph`. 
This work makes it simple to sink outputs from render passes into shaders. 

Changes:
1.) Moved `ShaderResourceBus` out of `NcGraphicsImpl` and into `VulkanGraphics`. This fixes a bug that was lurking: `ShaderResourceBus` was created in a helper method, then passed by ref into the `VulkanGraphics` CTOR- which held on to pieces of it beyond the lifetime of the CTOR, and then the helper method std::moved  `ShaderResourceBus` into `NcGraphicsImpl`. 

2.) Created a new descriptor set that will hold all render pass outputs, as keeping it in the same set that reads from it was causing problems.

3.) Deleted a few factory methods that created vk objects related to render passes based on parameters and switch statements and instead create the vk objects directly in the render pass - which makes the render passes much easier to parse.

4.) Render passes write to images that may or may not be used by shaders. I've added a method to fetch the `vk::ImageView`s that are shader inputs.

5.) Renamed the `PostProcessImageArrayBuffer` and its ilk to `RenderPassSinkBuffer`, as PostProcessImageArray was too specific and misleading.


source/engine/graphics/api/vulkan/renderpasses/RenderPass.cpp
```diff
@@ -178,4 +197,15 @@ void RenderPass::CreateFrameBuffers(std::span<const vk::ImageView> views, Vector
 
     m_frameBuffers.emplace_back(m_device.createFramebufferUnique(framebufferInfo));
 }
+
+auto RenderPass::GetSinkViews() const -> std::span<const vk::ImageView>
+{
+    if (m_sourceSinkPartition > m_sinkViews.size())
```
McCallisterRomer: I think we want >= here to avoid a span with a sketchy pointer below. Even though the size would be zero, I think its technically UB.

McCallisterRomer (approved):

PR #682, opened 2024-06-29
Pipeline cleanup and unification

jaremieromer:
### Problem
For shadow mapping and future rendering features, we need to pass per-pass or other types of data to invocations of `ITechnique::Record`. 
For shadow mapping, I had been passing that data through the CTOR of `ShadowMappingTechnique`. This had caused the need for a unique workflow in `RenderPass` for registering/unregistering shadow mapping techniques and state tracking regarding which technique was actively recording with regards to which data it had.

### Solution
It is a whole lot simpler just to pass any such data in a struct directly to `ITechnique::Record`.

This PR creates that struct `PerFrameInstanceData` and passes it into `ITechnique::Record`, and eliminates that special case stuff for shadow mapping in `RenderPass`. 

Since I already had to modify each `XTechnique` file, I also removed the redundant `CanRecord` and `CanBind` checks from the technique interface, and renamed techniques to pipelines as `Technique` was the incorrect term.


source/engine/graphics/api/vulkan/RenderGraph.cpp
```diff
@@ -229,15 +227,14 @@ void RenderGraph::BuildRenderGraph(const PerFrameRenderStateData& stateData, uin
...
-        renderGraph.shadowPasses.clear();
-        for (auto i : std::views::iota(0u, stateData.omniDirLightsCount))
+        while (renderGraph.shadowPasses.size() < stateData.omniDirLightsCount + stateData.uniDirLightsCount)
         {
-            renderGraph.shadowPasses.push_back(CreateShadowMappingPass(m_device, m_gpuAllocator, m_swapchain, m_shaderBindingManager, m_dimensions, i, true));
+            renderGraph.shadowPasses.push_back(CreateShadowMappingPass(m_device, m_gpuAllocator, m_swapchain, m_shaderBindingManager, m_dimensions));
         }
 
-        for (auto i : std::views::iota(stateData.omniDirLightsCount, stateData.omniDirLightsCount + stateData.uniDirLightsCount))
+        while (renderGraph.shadowPasses.size() > stateData.omniDirLightsCount + stateData.uniDirLightsCount)
```
McCallisterRomer: These while loops seem weird to me. Won't only one of them run, depending on whether the new size is greater or less than the current? To me, that'd be better expressed with an if statement. You could also do reserve + generate_n when we need to add, and resize when we need to shrink.

source/engine/graphics/api/vulkan/RenderGraph.cpp
```diff
@@ -284,10 +281,14 @@ void RenderGraph::Execute(const PerFrameRenderState &frameData, const Vector2& d
 
     SetViewportAndScissorFullWindow(cmd, dimensions);
 
-    for (auto& shadowMappingPass : renderGraph.shadowPasses)
+    auto instanceData = PerFrameInstanceData{};
+
+    for (auto [index, shadowMappingPass] : std::views::enumerate(renderGraph.shadowPasses))
     {
+        instanceData.isOmniDirectional = index < renderGraph.stateData.omniDirLightsCount ? true : false;
```
McCallisterRomer: Ternary is redundant here since the result of < is a bool.

McCallisterRomer (approved):

source/engine/graphics/api/vulkan/RenderGraph.cpp
jaremieromer: Yes, exactly. I like the reserve/generate_n and resize approach

source/engine/graphics/api/vulkan/RenderGraph.cpp
jaremieromer: Thanks!

PR #685, opened 2024-08-10
Use VMA C API

### LINKED ISSUE #684: vulkan memory allocator not compatible with latest vulkan sdk
vulkan sdk 1.3.290.0 is not compatible with the included vulkan memory allocator
image

to solve it you have to downgrade from 1.3.2.690.0 but this wont work forever because eventually the download for this version will disappear

McCallisterRomer:
resolves #684

We've been using a pinned version of the vma hpp bindings, which can't be assumed to compatible with the VulkanSDK across patch versions. Its also not trivial to identify the correct ref if we were to try and fetch a compatible version based on the detected sdk version. Instead, I've updated our usage of vma to use the c api. This allows us to prefer using the (optional) header installed with the VulkanSDK. If this is unavailable, we can fall back to a vendored c header.


cmake/FetchDependencies.cmake
```diff
@@ -2,10 +2,19 @@
...
+include(FetchContent)
+
 # Vulkan
 find_package(Vulkan REQUIRED)
 
-include(FetchContent)
+# Vulkan Memory Allocator
+if(EXISTS "${Vulkan_INCLUDE_DIR}/vma/vk_mem_alloc.h")
+    message(STATUS "Using VMA from Vulkan SDK: ${Vulkan_INCLUDE_DIR}")
+    set(VMA_INCLUDE_DIR ${Vulkan_INCLUDE_DIR})
```
jaremieromer, line 13: The FindVulkan CMake module uses such goofy casing

jaremieromer (approved):
Awesome job, thanks for doing this!

cmake/FetchDependencies.cmake
McCallisterRomer: I know. The rules for find modules are to name the script FindFoo.cmake and to output variables in the form Foo_VARIABLE. Since package names tend to not be all caps, you see this everywhere.

PR #698, opened 2024-08-30
Initial Jolt Integration

### LINKED ISSUE #688: Initial Jolt Implementation
Get basic Jolt POC implemented - build, init, simple rigid body type.

McCallisterRomer:
resolves #688 

Minimal `NcPhysics` implementation using Jolt. Sort of a doozy, but it took a fair bit to get meaningful feedback in the sample.

This only includes a bare bones module and `RigidBody` component (with box/sphere and static/dynamic options).

The new module (and tests) are always built, but the flavor of `NcPhysics` that gets registered at runtime is determined by the `NC_USE_JOLT` feature flag, defaulting to the old one.


cmake/FetchDependencies.cmake
```diff
@@ -87,13 +87,25 @@ FetchContent_Declare(optick
                      GIT_SHALLOW    TRUE
 )
 
+# Jolt
+FetchContent_Declare(JoltPhysics
+                     GIT_REPOSITORY https://github.com/jrouwe/JoltPhysics
+                     GIT_TAG        v5.1.0
+                     GIT_SHALLOW    TRUE
+                     SOURCE_SUBDIR  "Build"
```
jaremieromer, line 95: What's with this subdirectory? Isn't source_subdir already relative to the build directory? Won't this put jolt in build/<compiler_name>/Build

jaremieromer: Eg: Putting `SOURCE_DIR _deps/MyPackage` downloads the source to 
build/VS2022-amd64/_deps/MyPackage

source/engine/physics/collision/narrow_phase/Gjk.h
```diff
@@ -252,13 +257,13 @@ bool Gjk(const BVA& a, const BVB& b)
         if(XMVector3Less(XMVector3Dot(supportCSO_v, direction_v), g_XMZero))
             break;
 
+
```
jaremieromer: Nit, but this newline can go away as well as there is a typo on line 7 "potitive"

jaremieromer (approved):

cmake/FetchDependencies.cmake
McCallisterRomer: Insane naming on CMake's part, but `SOURCE_DIR` controls where things get unpacked and `SOURCE_SUBDIR` points to where the root CMakeLists.txt is. By default, that's expected to be in `SOURCE_DIR`, but jolt has it placed in a `Build` subdirectory for some reason. So I'm just telling CMake to look there when it runs the configure step.

PR #701, opened 2024-08-31
Extend ComponentHandler::drawUI

### LINKED ISSUE #689: Investigate Syncing Transform + RigidBody
Syncing RigidBody writes back to Transform is pretty easy, but the other way poses some problems - physics writes dirty the Transform, so we need to be able to differentiate client/physics Transform writes. Another option is to expose Transform functions directly through RigidBody and require client code to go through there (simpler to implement but clunkier to use).

Whatever the solution, we'll want to support updating through Transform in the editor.

McCallisterRomer:
part of #689

Passing the parent `Entity`, `EditorContext`, and `ComponentHandler::userData` to `ComponentHandler::drawUI` in addition to the component. This was a change I made in the game feature branch for one of the components over there, but its also needed for jolt integration - updating a `Transform` through the editor will need to know if the associated `Entity` also has a `RigidBody`.

resolves #689 (from #702)


include/ncengine/ecs/AnyComponent.h
```diff
@@ -74,12 +74,12 @@ class AnyComponent
         }
 
         /** @brief Invoke ComponentHandler::drawUI with the component instance, if it is set. */
-        void DrawUI()
+        void DrawUI(Entity self, ui::editor::EditorContext& ctx)
```
jaremieromer: Per our conversation - don't need 'self' passed in separately because 'selectedEntity' is already in `EditorContext`

jaremieromer (approved):

PR #702, opened 2024-09-02
Simulated Body Functions

### LINKED ISSUE #689: Investigate Syncing Transform + RigidBody

McCallisterRomer:
resolves #689 

PR is into feature branch because it heavily depends on changes over there. CI run [here](https://github.com/NcStudios/NcEngine/actions/runs/10671239127) since it won't trigger.

- Adding functions for updating a `Transform` and `RigidBody` in a synchronized way. Updating a `Transform` through the editor will use these when the `Entity` has a `RigidBody`.
- Added a flag for decoupling the scale of a collision shape from the `Transform`. This is because not all scales are valid for all shapes, so, by default, we constrain `Transform` scaling when necessary. This may not always be desirable - maybe we want to apply non-uniform scaling to a mesh, but we want sphere collider on it. The editor respects this flag.
- Unrelated to physics, but static `Transform`s automatically rebuild when updated through the editor. The debug menu has an option to disable this behavior.

Small rant on the first point: I originally wanted to add these to `Transform` so that they would be discoverable alongside all their related `Transform` methods. However, this caused a cascade of undefined symbols in tests. After thinking about it some more, I believe they belong with physics code. The rationale being based on previous conversations about writing modules as though they were separate libraries. Under this model, `Transform` would necessarily belong to a 'core' lib that everything else would depend on, and could/should not refence anything in `nc::physics`. My two cents, feel free to counter.

### DIFF
--- a/include/ncengine/ecs/Transform.h
+++ b/include/ncengine/ecs/Transform.h
@@ -67,7 +67,7 @@ class Transform final : public ComponentBase
         /** @brief Get local space position */
         auto LocalPosition() const noexcept -> Vector3 { return ToVector3(LocalPositionXM()); }
 
-        /** @brief Get loca space position as an XMVECTOR */
+        /** @brief Get local space position as an XMVECTOR */
         auto LocalPositionXM() const noexcept -> DirectX::FXMVECTOR { return m_localMatrix.r[3]; }
 
         /** @brief Get world space rotation */
--- a/include/ncengine/ecs/Transform.h
+++ b/include/ncengine/ecs/Transform.h
@@ -59,28 +59,40 @@ class Transform final : public ComponentBase
         Transform& operator=(const Transform&) = delete;
 
         /** @brief Get world space position */
-        auto Position() const noexcept -> Vector3 { return ToVector3(m_worldMatrix.r[3]); }
+        auto Position() const noexcept -> Vector3 { return ToVector3(PositionXM()); }
 
-        /** @brief Get world space position as XMVECTOR */
+        /** @brief Get world space position as an XMVECTOR */
         auto PositionXM() const noexcept -> DirectX::FXMVECTOR { return m_worldMatrix.r[3]; }
 
         /** @brief Get local space position */
-        auto LocalPosition() const noexcept -> Vector3 { return ToVector3(m_localMatrix.r[3]); }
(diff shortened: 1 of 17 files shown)


include/ncengine/ecs/Transform.h
```diff
@@ -59,28 +59,40 @@ class Transform final : public ComponentBase
...
 
-        /** @brief Get world space position as XMVECTOR */
+        /** @brief Get world space position as an XMVECTOR */
         auto PositionXM() const noexcept -> DirectX::FXMVECTOR { return m_worldMatrix.r[3]; }
 
         /** @brief Get local space position */
-        auto LocalPosition() const noexcept -> Vector3 { return ToVector3(m_localMatrix.r[3]); }
+        auto LocalPosition() const noexcept -> Vector3 { return ToVector3(LocalPositionXM()); }
+
+        /** @brief Get loca space position as an XMVECTOR */
```
jaremieromer: *local

include/ncengine/physics/RigidBody.h
```diff
@@ -1,53 +1,124 @@
...
+struct ComponentContext;
+class NcPhysicsImpl2;
+class RigidBody;
+
+/** @brief Handle to internal RigidBody state. */
 using BodyHandle = void*;
-using BodyInterface = void*;
 
+/** @brief Flags for configuring RigidBody behavior. */
+struct RigidBodyFlags
```
jaremieromer, line 22: I don't love this data structure:
Flags shouldn't be used for options that are mutually exclusive (`None` and `ScaleWithTransform`). 

As it stands a user could happily do something like:
`constexpr auto g_flags = nc::physics::RigidBodyFlags::ScaleWithTransform | nc::physics::RigidBodyFlags::None;` 

We can easily prevent by extracting the idea of scaling with transform to a boolean.

jaremieromer (approved):
Just the comment about the flags for RigidBody.
I agree about not letting RB stuff leak into transform

PR #705, opened 2024-09-03
Shape Scaling and Position

### LINKED ISSUE #692: Support Primitive Shape Types

### LINKED ISSUE #695: Support collision events
1. This may end up being a larger design change since the callback structure is wildly different. Consider if we want a separate type for sensors. May want a separate object layer too.
2. For the game, we'll need contact point info or maybe ray casting. In either case, contact info is valuable for the collision event.

### LINKED ISSUE #703: Support shape scaling + offset

McCallisterRomer:
resolves #692, resolves #703, resolves #695

- Adding capsule shape
- Adding local scale + position values for shapes.
- Updating `WidgetSystem` to render wireframes based on `RigidBody` with `NC_USE_JOLT` enabled.

#707 was merged into this

### DIFF
--- a/include/ncengine/physics/RigidBody.h
+++ b/include/ncengine/physics/RigidBody.h
@@ -134,9 +134,9 @@ class RigidBody
         void AddTorque(const Vector3& torque);
 
         auto IsInitialized() const noexcept -> bool { return m_handle; }
-        void SetPosition(const Vector3& position, bool wake = true);
-        void SetRotation(const Quaternion& rotation, bool wake = true);
-        auto SetScale(const Vector3& scale, bool wake = true) -> Vector3;
+        void SetBodyPosition(const Vector3& position, bool wake = true);
+        void SetBodyRotation(const Quaternion& rotation, bool wake = true);
+        auto SetBodyScale(const Vector3& previousScale, const Vector3& newScale, bool wake = true) -> Vector3;
 
     private:
         friend class NcPhysicsImpl2;
--- a/include/ncengine/physics/RigidBody.h
+++ b/include/ncengine/physics/RigidBody.h
@@ -4,10 +4,10 @@
  */
 #pragma once
 
+#include "Shape.h"
 #include "ncengine/ecs/Component.h"
 #include "ncengine/ecs/Transform.h"
-
-#include "DirectXMath.h"
+#include "ncengine/utility/MatrixUtilities.h"
 
 namespace nc::physics
 {
@@ -36,13 +36,6 @@ struct RigidBodyFlags
 
     /** @brief Default flag values. */
     static constexpr Type Default = ScaleWithTransform;
-};
-
-/** @brief RigidBody shape options. */
-enum class Shape : uint8_t
-{
-    Box,
-    Sphere
 };
 
 /** @brief Determines movement and collision behavior of a RigidBody. */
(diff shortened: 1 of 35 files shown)


include/ncengine/physics/RigidBody.h
```diff
@@ -4,10 +4,10 @@
  */
 #pragma once
 
+#include "Shape.h"
```
jaremieromer: we want our includes in the include dir to be relative to the include dir.
aka: change to `#include "ncengine/physics/Shape.h"`

include/ncengine/physics/RigidBody.h
```diff
@@ -93,7 +86,10 @@ auto SetSimulatedBodyScale(Transform& transform,
 class RigidBody
 {
     public:
-        RigidBody(Entity self, Shape shape, BodyType bodyType, RigidBodyFlags::Type flags = RigidBodyFlags::Default)
+        RigidBody(Entity self,
+                  const Shape& shape = Shape::MakeBox(),
```
jaremieromer, line 54: Can you clarify your intent with passing a value type to a const ref arg, and then assigning that const ref arg to a value type? It probably makes sense but it's confusing me.

include/ncengine/physics/RigidBody.h
```diff
@@ -128,18 +124,19 @@ class RigidBody
...
         void AddImpulse(const Vector3& impulse);
         void AddTorque(const Vector3& torque);
 
         auto IsInitialized() const noexcept -> bool { return m_handle; }
-        void SetPosition(const Vector3& position, bool wake = true);
-        void SetRotation(const Quaternion& rotation, bool wake = true);
-        auto SetScale(const Vector3& scale, bool wake = true) -> Vector3;
+        void SetBodyPosition(const Vector3& position, bool wake = true);
+        void SetBodyRotation(const Quaternion& rotation, bool wake = true);
+        auto SetBodyScale(const Vector3& previousScale, const Vector3& newScale, bool wake = true) -> Vector3;
```
jaremieromer: Should this be private? Looks like we only use it in the .cpp, and the way it requires a previous scale feels like an implementation detail that might not be apparent to someone at the scene level

source/engine/physics2/Shape.cpp
```diff
@@ -0,0 +1,80 @@
...
+    if (out.y <= 0.0f)
+        out.y = nc::physics::g_minimumShapeScale;
+    if (out.z <= 0.0f)
+        out.z = nc::physics::g_minimumShapeScale;
+}
+
+void FixSphereScale(const nc::Vector3& current, nc::Vector3& desired)
+{
+    if (!nc::FloatEqual(current.x, desired.x))
+        desired =  nc::Vector3::Splat(desired.x);
```
jaremieromer: Extreme nit, spacing

jaremieromer (approved):

include/ncengine/physics/RigidBody.h
McCallisterRomer: If move were meaningful to `Shape`, I'd take by value and move into the member here, but its full of primitives. Since R values and L values can bind to const &, taking by const & allows the caller to give us an temporary, but also avoids an extra copy in the case where an L value is passed in (copy once from call site to the parameter, then again from parameter to the member). If an R value is passed, we get the double copy, but we can't do anything about that. Although, I imagine the compiler can optimize that just fine. If we get the default param, its the same as if we were given an R value.

include/ncengine/physics/RigidBody.h
McCallisterRomer: I wasn't quite sure. `SetBodyPosition()` and `SetBodyRotation()` don't need to be called from anything except the simulated body functions, but `SetBodyScale()` is required by client code when `ScaleWithTransform` is disabled. I left them all here, because I don't love friend functions, but I've been adding a few here and there. Still unsure.

McCallisterRomer: Oh, I think I could make the simulated body functions members of `RigidBody` without too much hassle.

PR #707, opened 2024-09-06
Collision Events

### LINKED ISSUE #695: Support collision events

McCallisterRomer:
resolves #695 

Adding collision enter/exit events. I added a new `CollisionListener` component for this because the signature is different than those in the existing `CollisionLogic`, and we may end up wanting separate components for collision/trigger events for internal reasons.

### DIFF
--- a/include/ncengine/physics/EventListeners.h
+++ b/include/ncengine/physics/EventListeners.h
@@ -1,0 +1,31 @@
+/**
+ * @file EventListener.h
+ * @copyright Jaremie Romer and McCallister Romer 2024
+ */
+#pragma once
+
+#include "HitInfo.h"
+#include "ncengine/ecs/EcsFwd.h"
+
+#include <functional>
+
+namespace nc::physics
+{
+/** @brief Callback type for collision enter events. */
+using OnCollisionEnter_t = std::function<void(Entity self,
+                                              Entity other,
+                                              const HitInfo& hit,
+                                              ecs::Ecs world)>;
+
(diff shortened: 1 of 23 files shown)


include/ncengine/physics/EventListeners.h
```diff
@@ -0,0 +1,31 @@
+/**
+ * @file EventListener.h
+ * @copyright Jaremie Romer and McCallister Romer 2024
+ */
+#pragma once
+
+#include "HitInfo.h"
```
jaremieromer: change to be relative to the include dir
`#include "ncengine/physics/HitInfo.h"`

source/engine/physics2/jolt/ContactListener.cpp
```diff
@@ -0,0 +1,129 @@
...
+        );
+    }
+}
+
+// note: This called from multiple threads with all bodies locked.
+void ContactListener::OnContactRemoved(const JPH::SubShapeIDPair& pairID)
+{
+    const auto id1 = pairID.GetBody1ID();
+    const auto id2 = pairID.GetBody2ID();
+    if (m_physicsSystem->WereBodiesInContact(id1, id2))
```
jaremieromer, line 87: I really like question phrasing for bool methods

jaremieromer (approved):

PR #717, opened 2024-09-10
Jolt/Taskflow integration

### LINKED ISSUE #715: Job system to connect Jolt to Taskflow
Doesn't look too tough: make a JobSystem class derived from `JobSystemWithBarrier` and dispatch using `tf::Executor::silent_async()`. Probably worth a look through Taskflow docs to see if there are potentially better options (especially for dispatching batches).

McCallisterRomer:
resolves #715 

Adding our own `JPH::JobSystem` implementation so we can run physics jobs on the engine's thread pool. Jolt has a built in base class that does most of the dependency tracking, so we can just run the jobs as individual async tasks.


source/engine/engine/NcEngineImpl.h
```diff
@@ -19,6 +19,7 @@ namespace nc
             void Shutdown() noexcept override;
             auto GetComponentRegistry() noexcept -> ecs::ComponentRegistry& override;
             auto GetModuleRegistry() noexcept -> ModuleRegistry* override;
+            auto GetAsyncDispatcher() noexcept -> task::AsyncDispatcher override;
```
jaremieromer, line 22: Earlier in `ModuleFactory`, we pass the `AsyncDispatcher` as a const ref all the way down into `TaskRouter` until we copy it into a member variable at that point. Don't we want downstream callers of `GetAsyncDispatcher` to follow the same pattern? Why not const ref here?

source/engine/physics2/jolt/JobSystem.cpp
```diff
@@ -0,0 +1,90 @@
...
+#include "Jolt/Physics/PhysicsSettings.h"
+
+namespace
+{
+// JobSystem implementation that routes jobs to Taskflow
+class TaskRouter : public JPH::JobSystemWithBarrier
+{
+    public:
+        TaskRouter(const nc::task::AsyncDispatcher& dispatcher)
+            : JPH::JobSystemWithBarrier{JPH::cMaxPhysicsBarriers},
```
jaremieromer, line 18: In the sample implementation Jolt provides for their `JobSystemWithBarrier`, they initialize their job system with a number of threads for their thread pool. Don't we want to do the same and keep ours in sync with what the user specifies in EngineSettings?

https://github.com/jrouwe/JoltPhysics/blob/2292c415a9a94cd7b4706de6ac8f24c9b5af74a5/Jolt/Core/JobSystemThreadPool.h#L31

source/engine/physics2/jolt/JobSystem.cpp
```diff
@@ -0,0 +1,90 @@
...
+                std::this_thread::sleep_for(std::chrono::microseconds(100));
+            }
+
+            return JPH::JobHandle{&m_freeList.Get(index)};
+        }
+
+    protected:
+        void QueueJob(Job* job) override
+        {
+            m_dispatcher.SilentAsync([job]()
```
jaremieromer, line 53: Do we lose the ability to maintain the exception context when we go this route? I know you do with fire-and-forget async in .NET land

source/engine/physics2/jolt/JoltApi.h
```diff
@@ -31,30 +31,32 @@ struct JoltApi : public StableAddress
...
 
     static auto Initialize(const config::MemorySettings& memorySettings,
-                           const config::PhysicsSettings& physicsSettings) -> JoltApi;
+                           const config::PhysicsSettings& physicsSettings,
+                           const task::AsyncDispatcher& dispatcher) -> JoltApi;
 
     void Update(float dt, int steps = 1)
     {
-        const auto error = physicsSystem.Update(dt, steps, &tempAllocator, &jobSystem);
+        const auto error = physicsSystem.Update(dt, steps, &tempAllocator, jobSystem.get());
```
jaremieromer, line 39: Same question about exception context as above for the fire-and-forget methods

jaremieromer (approved):

source/engine/engine/NcEngineImpl.h
McCallisterRomer: I think its ok to return by value here, since its just a wrapper around a pointer. It should be treated as a value type by default. The const & in other places is so we can use it use it as an incomplete/forward declared type it in each of the functions it passes through on its way to the job system. The header has a taskflow include. If we took by value, each of those source files would depend on taskflow, which affects a bunch of tests.

PR #718, opened 2024-09-14
ConstraintFactory + ConstraintManager

### LINKED ISSUE #696: Physics Constraints
~~We need at least parity with our current joint, velocity restriction, position clamp, and orientation clamp. Not yet sure if these should continue to be separate components, or if it makes sense to combine them.~~ parity is tough - way may have to do some work for soft constraints.

McCallisterRomer:
resolves #696 

Full change set for constraints was getting outrageously large. This is a just a few of the pieces needed:
- Adding a few constraint-related types to API
- Adding `ConstraintFactory` to build `Jolt::Constraint`s from our `ConstraintInfo` types.
- Adding `ConstraintManager` to track constraints and maintain a mapping between constraints and body pairs.
- Updating `NormalizeScaleForShape()` to compute an average scale if the current and desired scales are equal. (same behavior if scales are different - if we resize in editor, scale based on which axis was scaled)
- Adding `JoltApiFixture` helper for tests. Physics tests are generally using the same base setup, so it cuts down a bit on repetition.

### DIFF
--- a/source/engine/physics2/jolt/ConstraintManager.cpp
+++ b/source/engine/physics2/jolt/ConstraintManager.cpp
@@ -1,0 +1,147 @@
+#include "ConstraintManager.h"
+
+#include "Jolt/Jolt.h"
+#include "Jolt/Physics/PhysicsSystem.h"
+
+namespace
+{
+void UntrackConstraint(nc::physics::EntityConstraints& state, nc::physics::ConstraintId constraintId, bool isOwner)
+{
+    auto idPos = std::ranges::find(state.ids, constraintId);
+    NC_ASSERT(idPos != state.ids.cend(), "ConstraintId not found");
+    *idPos = state.ids.back();
+    state.ids.pop_back();
+    if (!isOwner)
+    {
+        return;
+    }
+
+    auto infoPos = std::ranges::find(state.views, constraintId, &nc::physics::ConstraintView::id);
+    NC_ASSERT(infoPos != state.views.cend(), "ConstraintView not found");
+    state.views.erase(infoPos);
+}
+} // anonymous namespace
+
+namespace nc::physics
+{
+auto ConstraintManager::AddConstraint(const ConstraintInfo& createInfo,
+                                      Entity owner,
+                                      JPH::Body* ownerBody,
+                                      Entity referenced,
+                                      JPH::Body* referencedBody) -> ConstraintId
+{
+    auto handle = m_factory.MakeConstraint(createInfo, *ownerBody, *referencedBody);
+    m_physicsSystem->AddConstraint(handle);
+
+    const auto ownerId = owner.Index();
+    const auto referencedId = referenced.Index();
+    const auto index = [&]()
+    {
+        if (m_freeIndices.empty())
+        {
+            const auto index = static_cast<uint32_t>(m_handles.size());
+            m_handles.push_back(handle);
+            m_pairs.emplace_back(ownerId, referencedId);
+            return index;
+        }
+
+        const auto index = m_freeIndices.back();
+        m_freeIndices.pop_back();
+        m_handles[index] = handle;
+        m_pairs[index] = ConstraintPair{ownerId, referencedId};
+        return index;
+    }();
+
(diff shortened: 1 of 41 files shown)


source/engine/physics2/jolt/ConstraintManager.cpp
```diff
@@ -0,0 +1,219 @@
+#include "ConstraintManager.h"
+
+#include "Jolt/Jolt.h"
+#include "Jolt/Physics/PhysicsSystem.h"
+#include "Jolt/Physics/Constraints/TwoBodyConstraint.h"
+
+namespace nc::physics
+{
+auto ConstraintManager::AddConstraint(const ConstraintInfo& createInfo,
```
jaremieromer, line 9: I'm not a huge fan of the fact that a second body is optional, but the caller is responsible for making a dummy body to represent the "world" constraint.

How about an overload taking in a single body that sets the second body to `&JPH::Body::sFixedToWorld` under the hood?

source/engine/physics2/jolt/ConstraintManager.cpp
```diff
@@ -0,0 +1,219 @@
...
+{
+    constraint.m_otherBody = referenced;
+
+    const auto id = constraint.GetId();
+    auto oldHandle = GetConstraintHandle(id);
+    auto [ownerBody, _] = GetConstraintBodies(oldHandle);
+    ReplaceInternalConstraint(constraint, oldHandle, ownerBody, referencedBody);
+
+    auto& [trackedOwner, trackedReference] = m_pairs.at(id);
+    NC_ASSERT(trackedOwner != referenced.Index(), "Attempt to self assign constraint attachment");
```
jaremieromer, line 78: Good thinking!

source/engine/physics2/jolt/ConstraintManager.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+    trackedReference = referenced.Index();
+    if (referenced.Valid())
+    {
+        TrackConstraint(trackedReference, id);
+    }
+}
+
+void ConstraintManager::RemoveConstraint(ConstraintId constraintId)
+{
```
jaremieromer, line 92: There is a lot of state that has to be managed for constraint removal. What happens if any of these methods (like `UntrackConstraint`, for example, throw? I don't have a great suggestion to safeguard against this - but it might be worth adding some additional sanity checks at the beginning to make sure we know we aren't going to throw before we start exchanging and untracking things

source/engine/physics2/jolt/ConstraintManager.h
```diff
@@ -0,0 +1,88 @@
...
+class Constraint;
+class PhysicsSystem;
+} // namespace JPH
+
+namespace nc::physics
+{
+// Pair of entity indices associated with a constraint
+struct ConstraintPair
+{
+    uint32_t ownerId = UINT32_MAX; // Index of the Entity the constraint was added to
```
jaremieromer: weren't we on team `std::numeric_limits<uint32_t>::max()`?

jaremieromer (approved):

source/engine/physics2/jolt/ConstraintManager.cpp
McCallisterRomer: That seems fair, ~~`RigidBody` already is set up that way~~ (I don't know what I'm talking about).

source/engine/physics2/jolt/ConstraintManager.h
McCallisterRomer: Yes, I see your `std::numeric_limits::max()` and I raise you an `Entity::NullIndex`.

source/engine/physics2/jolt/ConstraintManager.cpp
McCallisterRomer: I can move the the assert if nullptr first always (you'r'e right that in `RemoveConstraint` and `RemoveConstraints` we could end up with nulling the pointer before throwing on a bad id!). As for any other functions throwing later, they could only throw if there's a bug in the tracking, which can't really be left in a valid/recoverable state.

PR #719, opened 2024-09-14
Initial Diligent Engine Integration

jaremieromer:
Added a new module - NcGraphicsImpl2. 
Added a method to `NcGraphics` to get supported graphics APIs.
Pulled in DiligentCore.
Added a new set of classes (`DiligentEngine`...) to initialize the device and context based on the target API and the platform.
Modified `NcWindow` to create a headless window if graphics is not enabled, and to provide a method to consumers to set the window.
Added some packages to the dev container and to the .yml for the pipeline.
Added two properties to `GraphicsSettings` in the config: `api` and `is_headless`


sample/CMakeLists.txt
```diff
@@ -1,6 +1,12 @@
 # Sample CMake Root
 add_executable(${NC_SAMPLE_EXE})
 
+set_target_properties(${NC_SAMPLE_EXE}
```
jaremieromer, line 4: Will update this before final review

source/engine/ecs/Transform.cpp
```diff
@@ -58,7 +58,9 @@ namespace nc
 
     void Transform::SetRotation(const Quaternion& quat)
     {
-        DirectX::XMVECTOR scl_v, rot_v, pos_v;
+        DirectX::XMVECTOR pos_v = DirectX::XMVectorZero();
```
jaremieromer, line 61: Should we ignore the "might be uninitialized" error instead of default initializing these?

source/engine/graphics2/diligent/DiligentEngineWin32.cpp
```diff
@@ -0,0 +1,170 @@
+#include "DiligentEngine.h"
+#include "ncengine/utility/Log.h"
+#include "ncengine/window/Window.h"
+#include "ncutility/NcError.h"
+
+#define GLFW_EXPOSE_NATIVE_WIN32 1
+#include "Graphics/GraphicsEngineD3D11/interface/EngineFactoryD3D11.h"
```
jaremieromer, line 7: These used to be wrapped in an #if PLATFORM_WIN32. I'll move these down to the other Diligent headers

source/engine/window/NcWindowImpl.cpp
```diff
@@ -160,6 +133,70 @@ namespace nc::window
...
+            m_window = glfwCreateWindow(width, height, m_windowName.c_str(), monitor, nullptr);
+        }
+
+        if(!m_window)
+        {
+            throw NcError("SetWindow failed");
+        }
+
+        glfwGetWindowContentScale(m_window, &m_contentScale.x, &m_contentScale.y);
+
```
jaremieromer, line 190: I'll remove this extra newline

test/graphics2/DiligentEngine_tests.cpp
```diff
@@ -0,0 +1,10 @@
+#include "gtest/gtest.h"
+#include "graphics2/DiligentEngine.h"
+
+using namespace nc::graphics;
+
+TEST(DiligentEngine_tests, AdjustDimensionsToAspectRatio_heightIsZero_doesntCrash)
```
jaremieromer, line 6: Ignore tests for now. I'll add them before marking the MR as 'non-draft'

source/engine/ecs/Transform.cpp
McCallisterRomer: ~~Its not the end of the world, but maybe? It is a false positive, but I wonder if gcc would think differently if we used our utility function instead: `[s, r, p] = DecomposeMatrix(m_localMatrix);`~~ stale

cmake/FetchDependencies.cmake
```diff
@@ -100,8 +100,19 @@ FetchContent_Declare(JoltPhysics
                      SOURCE_SUBDIR  "Build"
 )
 
+# Diligent
+set(DILIGENT_INSTALL_CORE OFF CACHE BOOL "" FORCE)
+set(DILIGENT_INSTALL_TOOLS  OFF CACHE BOOL "" FORCE)
+set(DILIGENT_INSTALL_SAMPLES  OFF CACHE BOOL "" FORCE)
```
McCallisterRomer: Probably want DILIGENT_BUILD_SAMPLES off as well.

Also, this doesn't seem like it applies to us:
>    By default Vulkan back-end is linked with glslang that enables compiling HLSL and GLSL shaders to SPIRV at run time. If run-time compilation is not required, glslang can be disabled with DILIGENT_NO_GLSLANG cmake option.

cmake/FetchDependencies.cmake
```diff
@@ -100,8 +100,19 @@ FetchContent_Declare(JoltPhysics
...
+set(DILIGENT_INSTALL_FX  OFF CACHE BOOL "" FORCE)
+
+FetchContent_Declare(DiligentCore
+GIT_REPOSITORY https://github.com/DiligentGraphics/DiligentCore.git
+SOURCE_DIR _deps/DiligentCore
```
McCallisterRomer: Definitely want a GIT_TAG here, and preferably GIT_SHALLOW TRUE.

include/ncengine/config/Config.h
```diff
@@ -69,17 +69,19 @@ struct MemorySettings
...
-    unsigned screenWidth = 1000;
-    unsigned screenHeight = 1000;
-    unsigned targetFPS = 60;
-    float nearClip = 0.5f;
-    float farClip = 400.0f;
-    bool useShadows = true;
-    unsigned antialiasing = 8u;
-    bool useValidationLayers = false;
+    bool enabled = true;              ///< enable the NcGraphics module
+    std::string api = "vulkan";       ///< possible values: vulkan, d3d12, d3d11, opengl
```
McCallisterRomer, line 79: We might want more detail - this is expected to be one of the values from `GetSupportedApis()`, and it may change in case of fallback. If you don't want to break the inline comment flow, maybe include a 'see GetSupportedApis()' or something, then have more detail there.

sample/CMakeLists.txt
```diff
@@ -29,6 +29,21 @@ target_link_libraries(${NC_SAMPLE_EXE}
...
+    copy_required_dlls(${NC_SAMPLE_EXE})
+
+    install(DIRECTORY      $<TARGET_FILE_DIR:${NC_SAMPLE_EXE}>/
+            DESTINATION    sample
+            FILES_MATCHING REGEX ".*\.(dll)" # Diligent DLLs and EXEs
+)
+elseif(UNIX AND NOT APPLE)
+    set_target_properties(${NC_SAMPLE_EXE}
+        PROPERTIES
+            INSTALL_RPATH "$<TARGET_FILE_DIR:${NC_SAMPLE_EXE}>/../_deps/diligentcore-build/Graphics/GraphicsEngineOpenGL;$<TARGET_FILE_DIR:${NC_SAMPLE_EXE}>/../_deps/diligentcore-build/Graphics/GraphicsEngineVulkan"
```
McCallisterRomer: I think `BUILD_RPATH` is what you want here. `INSTALL_RPATH` should not referencing within the build tree. Are we still setting `INSTALL_RPATH` elsewhere to point to the lib installation dir (and are the libs installed on linux)? If not, we'll need to.

I think you should be able to target the dll files themselves, rather than pathing from the sample.

Doesn't have to be done now, but we'll want a nicer way to expose these paths to a consuming game, as they'll have to set their rpath too.

source/engine/CMakeLists.txt
```diff
@@ -30,6 +30,24 @@ if(NC_PHYSICS_CONFIG)
     add_definitions(-DNC_PHYSICS_CONFIG="${NC_PHYSICS_CONFIG}")
 endif()
 
+# Add debug post fix of 'd' to NC_ENGINE_LIB targets
+set_target_properties(${NC_ENGINE_LIB}
+    PROPERTIES CMAKE_DEBUG_POSTFIX d
+)
+
+SET(DILIGENT_LIBRARIES
```
McCallisterRomer: Doesn't have to be done now, but we'll probably want this set with `PARENT_SCOPE` or something from `FetchDependencies`. I don't think this will be visible to tests.

source/engine/graphics2/NcGraphicsImpl2.cpp
```diff
@@ -0,0 +1,195 @@
...
+    );
+}
+
+void NcGraphicsImpl2::Update()
+{
+}
+
+void NcGraphicsImpl2::Run()
+{
+    OPTICK_CATEGORY("Render", Optick::Category::Rendering);
```
McCallisterRomer: Will you change this to `NC_PROFILE_TASK("Render", ProfileCategory::Rendering)` from `debug/Profile.h`?

source/engine/graphics2/NcGraphicsImpl2Linux.cpp
```diff
@@ -0,0 +1,37 @@
+
+#include "ncengine/graphics/NcGraphics.h"
+
+#include <dlfcn.h>
+
+namespace
+{
+auto CheckLibrary(const char* soName) -> bool
+{
+    void* handle = dlopen(soName, RTLD_LAZY);
```
McCallisterRomer, line 10: RTDL_LAZY is handy

source/engine/graphics2/NcGraphicsImpl2Win32.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include "ncengine/platform/win32/NcWin32.h"
+
+#include <string_view>
+#include <vector>
+
+namespace
+{
+auto CheckLibrary(char const* dllName) -> bool
+{
+    auto hModule = LoadLibrary(dllName);
```
McCallisterRomer: `LoadLibraryExA("foo", NULL, LOAD_LIBRARY_AS_DATAFILE)` will prevent calling DllMain.

source/engine/graphics2/diligent/DiligentEngineLinux.cpp
```diff
@@ -0,0 +1,112 @@
...
+            catch (const std::runtime_error& e)
+            {
+                EnsureContextFlushed(m_pImmediateContext);
+                NC_LOG_WARNING("Failed to initialize OpenGL.");
+                errorMessage = fmt::format("{0} Failed to initialize OpenGL: {1} \n", errorMessage, e.what());
+            }
+        }
+    }
+
+    if (m_renderApi == "" || errorMessage != "")
```
McCallisterRomer: If vulkan failed but falling back to opengl succeeded, won't this throw because errorMessage isn't empty?

source/engine/graphics2/diligent/DiligentEngineWin32.cpp
```diff
@@ -0,0 +1,170 @@
...
+            catch (const std::runtime_error& e)
+            {
+                EnsureContextFlushed(m_pImmediateContext);
+                NC_LOG_WARNING("Failed to initialize OpenGL.");
+                errorMessage = fmt::format("{0} Failed to initialize OpenGL: {1} \n", errorMessage, e.what());
+            }
+        }
+    }
+    
+    if (m_renderApi == "" || errorMessage != "")
```
McCallisterRomer: Same Q as with linux

McCallisterRomer (approved):

sample/CMakeLists.txt
McCallisterRomer: On the last point, I suppose if things are installed in a reasonable way, a game can just point to that directory.

source/engine/graphics2/diligent/DiligentEngineLinux.cpp
jaremieromer: Yes - great catch!

source/engine/graphics2/diligent/DiligentEngineWin32.cpp
jaremieromer: Yes - great catch!

sample/CMakeLists.txt
jaremieromer: My recent acquisition of CMake skills is coming up short here. I don't _think_ we need to set the BUILD_RPATH at all. This step only ensures that the installed executable has knowledge of the libs. That being said, I do see now why making that path relative to the build dir is wrong. I've changed it, and I think my new solution will work, testing now

PR #744, opened 2024-10-06
Warning Fixes + Cleanup

McCallisterRomer:
- Fixed gcc warnings with assimp and Transform.cpp.
  - Added cmake functions for this to cut down on repetition
  - Added warning suppression macros (and a few other minor things) to Platform.h
- Moved Platform.h and win32 stuff over to ncutility so it can be shared by other libs
Moved MatrixUtilities.h to ncmath b/c that's its natural habitat


cmake/FetchDependencies.cmake
```diff
@@ -1,9 +1,24 @@
...
+function(disable_warnings_for_headers target)
+    get_target_property(target_include_dirs ${target} INTERFACE_INCLUDE_DIRECTORIES)
+    if (target_include_dirs)
+        target_include_directories(${target} SYSTEM INTERFACE ${target_include_dirs})
+    endif()
+endfunction()
+
+include(FetchContent)
+
 #############################
```
jaremieromer: It feels a little weird to have this banner style comment not at the top of the file anymore. Maybe just make it a standard comment?

cmake/FetchDependencies.cmake
```diff
@@ -143,6 +151,9 @@ if(NC_BUILD_NCCONVERT)
     )
 
     FetchContent_MakeAvailable(assimp)
+
+    disable_warnings_for_headers(assimp)
```
jaremieromer, line 151: Can these not go above in the block where all the other `disable_warnings` function calls are?

jaremieromer (approved):

cmake/FetchDependencies.cmake
McCallisterRomer: Yeah, I like that

cmake/FetchDependencies.cmake
McCallisterRomer: No, dependent on assimp being 'made available', which is guarded on an option.

PR #746, opened 2024-10-08
Include DiligentTools

McCallisterRomer:
- Added DiligentTools and additional dependencies for linux
- Fixed segfault in sample when `NC_USE_DILIGENT` is enabled due to null camera
- Switching to ninja in linux CI build & adding parallel flag (run time went from ~42 min to ~23 min!)


.devcontainer/ubuntu.dockerfile
```diff
@@ -1,27 +1,54 @@
...
+    python3.12-venv \
+    libvulkan-dev \
+    xorg-dev \
+    xvfb
+
+# setup python virtual environment & get setuptools
+RUN python3 -m venv /opt/venv && /opt/venv/bin/pip install --upgrade setuptools
+
+# Gfx Packages (split out for faster rebuilds, as this list is more in flux)
+RUN apt install -y \
```
jaremieromer, line 35: Will you audit these to see how many we can get rid of now that we dropped OpenGL support?

jaremieromer (approved):
Awesome!

PR #747, opened 2024-10-10
Added unit tests around initialization of the DiligentEngine

jaremieromer:
** Changes from Code Review Feedback:

I initially split out the tests into headless tests and ones requiring a display. However, much to my horror I realized GLFW has an unresolved issue where headless GLFW windows still require displays: https://github.com/glfw/glfw/issues/1882.
In the meantime, I've moved all tests into a new folder `integration_test`.

I moved a bunch of common code into a shared file.

I removed OpenGL support.

*****

The tests cover the following scenarios on Linux and Windows, on the following APIs (opengl, vulkan, d3d12 (Windows only), d3d11 (Windows only)):

1. Initializing DiligentEngine with a window
2. Initializing DiligentEngine with a headless window
3. Rendering a triangle for one second

We also test that falling back works in failure cases.

Note: We really should scrap OpenGL. I experimented with getting the window to render a triangle with OpenGL on Linux, and I couldn't figure out how to do it. We are able to render a triangle on Linux with Vulkan.

The tests are not currently running on the pipeline, I'm not passing NC_USE_DILIGENT as a build param. I want to get your input on expanding the matrix.


McCallisterRomer:
We should only have to expand the matrix once we want a new build of the sample/smoke test. For unit tests in the pipeline we have two options:
1. They probably just work without the definition, so just build and run them all the time (with NC_USE_DILIGENT ON or OFF). This is what I did with jolt.
2. If its needed, add the definition to each of the tests.

.devcontainer/ubuntu.dockerfile
```diff
@@ -1,27 +1,57 @@
+# To run the sample/smoke test, an X11 server needs to be setup after container launch.
```
McCallisterRomer, line 1: Will you take just what you needed from my changes (prob gdb + ninja)? I don't want to take on the rest of this unless we end up needing DiligentTools.

source/ncengine/graphics2/diligent/DiligentEngineLinux.cpp
```diff
@@ -24,7 +24,7 @@ void EnsureContextFlushed(Diligent::IDeviceContext* context)
 
 namespace nc::graphics
 {
-DiligentEngine::DiligentEngine(const config::GraphicsSettings& graphicsSettings, window::NcWindow& window_, std::span<const std::string_view> supportedApis)
+DiligentEngine::DiligentEngine(const config::GraphicsSettings& graphicsSettings, Diligent::EngineCreateInfo engineCreateInfo, GLFWwindow* window_, std::span<const std::string_view> supportedApis)
```
McCallisterRomer: Fella wants to be a const &

test/ncengine/graphics2/win32/DiligentEngineWin32_tests.cpp
```diff
@@ -0,0 +1,434 @@
+#include "gtest/gtest.h"
```
McCallisterRomer: There is a bunch of common code between the platforms. Can we extract out into some stub/collateral helpers?

McCallisterRomer: Take a look at `test/ncengine/physics/jolt`. In there, `JoltApiFixture.inl`, `ContactListener_stub.inl`, and `JobSystem_stub.inl` get repurposed by most of the tests.

test/ncengine/graphics2/win32/DiligentEngineWin32_tests.cpp
```diff
@@ -0,0 +1,434 @@
...
+    PSOCreateInfo.pVS = pVS;
+    PSOCreateInfo.pPS = pPS;
+    engine->Device()->CreateGraphicsPipelineState(PSOCreateInfo, ppPso);
+}
+
+void RenderSquare(nc::graphics::DiligentEngine* engine, Diligent::IPipelineState* pPso)
+{
+    using namespace Diligent;
+
+    const float ClearColor[] = {0.350f, 0.350f, 0.350f, 1.0f};
```
McCallisterRomer: I wonder if we should gate 'interactive' graphics tests on a variable or something. I don't love the idea of forcing window creation and rendering just by running the unit tests. 99% of my lifetime interaction with linux has been via ssh, and I don't want to put a hard requirement that debugging a non-graphics test on linux requires setting up an x11 server. Similar story for ubuntu runners, although these tests may be able to just use the display we set up for the smoke test.

test/ncengine/graphics2/win32/DiligentEngineWin32_tests.cpp
```diff
@@ -0,0 +1,434 @@
...
+    drawAttrs.NumVertices = 4;
+    engine->Context()->Draw(drawAttrs);
+}
+} // anonymous namespace
+
+using namespace nc::graphics;
+
+TEST(DiligentEngineWin32_tests, CreateDiligentEngine_VulkanNotHeadless_Succeeds)
+{
+    ExpectDiligentInitSucceeds(false, "vulkan", false);
```
McCallisterRomer: Are we fine saying that tests will only pass on environments that support all apis? For example, what if the surface can't initialize D3D12 or something?

test/ncengine/graphics2/win32/DiligentEngineWin32_tests.cpp
```diff
@@ -0,0 +1,434 @@
...
+}
+
+TEST(DiligentEngineWin32_tests, CreateDiligentEngine_VulkanHeadless_Succeeds)
+{
+    ExpectDiligentInitSucceeds(true, "vulkan", false);
```
McCallisterRomer: These tests, especially vulkan, dump an absurd amount to stdout. I suppose that's kind of nice in the case of failure, but I'm finding it to be quite difficult to sift through.

McCallisterRomer (commented):
The following 4 tests fail on my machine:

DiligentEngineWin32_tests.CreateDiligentEngine_VulkanNotHeadless_Succeeds
DiligentEngineWin32_tests.CreateDiligentEngine_OpenGLNotHeadless_Succeeds
DiligentEngineWin32_tests.CreateDiligentEngine_OpenGLHeadless_Succeeds
DiligentEngineWin32_tests.CreateDiligentEngine_OpenGLRenderTriangle_Succeeds

I would post output, but its hefty. Probably easier to just pop on my box.

test/ncengine/graphics2/win32/DiligentEngineWin32_tests.cpp
jaremieromer: I haven't found a sure-fire way to silence them yet, but still digging.

source/ncengine/graphics2/NcGraphicsImpl2.cpp
```diff
@@ -74,16 +74,13 @@ namespace nc::graphics
 
         if (graphicsSettings.enabled)
         {
-            auto ncAsset = modules.Get<asset::NcAsset>();
             auto ncWindow = modules.Get<window::NcWindow>();
-            NC_ASSERT(ncAsset, "NcGraphics requires NcAsset to be registered before it.");
```
McCallisterRomer, line 79: This assert should stay. We can discard the value like with NcScene, if its currently unused.

test/ncengine/window/NcWindowImpl.inl
```diff
@@ -0,0 +1,69 @@
+#include "window/Window.h"
```
McCallisterRomer: We don't rely directly on the symbol `NcWindowImpl` in graphics, correct? I'd bet its just the base `NcWindow`. If true, this file and type should be named something like `NcWindowStub`.

McCallisterRomer (approved):

PR #748, opened 2024-10-12
Add meshoptimizer

McCallisterRomer:


jaremieromer (approved):
Looks like no real change in amount of cubes we can render, but there was a significant improvement with more complicated meshes!

We can render 2.5x more ogre meshes with this change!

Nice work

McCallisterRomer:
@jaremieromer Are there potential problems with animation here due to vertex reordering (or deduping)? I suspected they would get wonky, but both skelington and ogre animations seemed normal after optimizing. If so, I can turn off optimizations for animated meshes.

jaremieromer:
> @jaremieromer Are there potential problems with animation here due to vertex reordering (or deduping)? I suspected they would get wonky, but both skelington and ogre animations seemed normal after optimizing. If so, I can turn off optimizations for animated meshes.

No, this was my concern leading up to the MR, but refreshing my memory of the animation code and process addressed my worries.

I had two concerns, initially:

1. What if the data store of vertex to bone gets out of sync with the animated bones data store?
2. What if we cull the only vertex that was influenced by bones?

For 1:
Not a concern. We grab the data of which vertex is mapped to which bones in `GeometryConverter` and store those bones' IDs in the `Vertex` before we cull vertices. Then, animation is all done at the bone level, and we're not culling bones/bone keyframe data. We pass a buffer of animated bone transforms and a mapping between _bone ID_ and _bone transform_ to the shader. The individual vertices already have this _boneID_ as part of `VertexData`.

For 2:
It could be a concern if we abuse weight painting to make the error occur - but in our workflow I don't think we'd ever run into it:

Animation roughly works like:
Each vertex has the four closest bones' weights (influences) painted on to it in 3d modeling software. This is done by locality to the bone.
We then get the animated transform for the bone each frame, and in the vertex shader the vertex then calculates it's transform per frame based on its four bones' transforms (more or less). 
Because our vertex weight painting workflow is done by locality to the bone - we don't have to worry about a scenario where we culled our only vertex with bone influences - because  meshoptimizer isn't going to cull a lone vertex (I'm thinking a really long and thin spike pyramid thingy) because it would be so significant it would not meet the exclusion criteria, and even if it was culled it'd be a wonky mesh from the get-go because that bone _should_ also be influencing the four vertices on the base of the pyramid. 

TLDR: I really don't think we're going to hit this case because of interpolation and the nature of weight painting meaning that similar vertices generally are influenced by similar bones at a similar weight.

PR #754, opened 2024-10-17
Initial Texture + Shader Loading

McCallisterRomer:
These are the changes I was after:
- Added `GlobalTextureBufferResource` which is sort of an analog to our current `TextureArrayBuffer`.
- Added `ShaderFactory` - currently only has runtime compilation; issue created for loading bytecode.

Plumbing and other stuff that came along for the ride:
- Added `AssetDispatch` to route events from asset managers to graphics.
- Added `GlobalResourceSignature` for sharing of resources across PSOs.
- Added `ShaderBindings` to house signatures/resources.
- Added `TestPipeline` so we can get some visual feedback from the sample (I believe nothing in this file is permanent).
- Added option to reroute output from diligent.


source/ncengine/graphics2/diligent/DiligentEngine.h
```diff
@@ -19,17 +23,23 @@ namespace graphics
 class DiligentEngine
 {
     public:
-        DiligentEngine(const config::GraphicsSettings& graphicsSettings, const Diligent::EngineCreateInfo& engineCreateInfo, GLFWwindow* window, std::span<const std::string_view> supportedApis);
+        DiligentEngine(const config::GraphicsSettings& graphicsSettings,
+                       const Diligent::EngineCreateInfo& engineCreateInfo,
+                       GLFWwindow* window,
+                       std::span<const std::string_view> supportedApis,
+                       Diligent::DebugMessageCallbackType logCallback = nullptr);
```
jaremieromer, line 30: Nice

cmake/FetchDependencies.cmake
```diff
@@ -83,9 +83,13 @@ set(DILIGENT_INSTALL_SAMPLES OFF CACHE BOOL "" FORCE)
 set(DILIGENT_INSTALL_FX OFF CACHE BOOL "" FORCE)
 set(DILIGENT_BUILD_SAMPLES OFF CACHE BOOL "" FORCE)
 set(DILIGENT_NO_RENDER_STATE_PACKAGER OFF CACHE BOOL "" FORCE)
-set(DILIGENT_NO_GLSLANG OFF CACHE BOOL "" FORCE)
 set(DILIGENT_NO_OPENGL ON CACHE BOOL "" FORCE)
 set(DILIGENT_NO_METAL ON CACHE BOOL "" FORCE)
+if(NC_RUNTIME_SHADER_COMPILATION)
+    set(DILIGENT_NO_GLSLANG OFF CACHE BOOL "" FORCE)
```
jaremieromer, line 89: We also want to have DILIGENT_NO_HLSL as part of this, or maybe just have _only_ DILIGENT_NO_HLSL in this conditional and always not build GLSLang at runtime - 
GLSLLang is the OpenGL/Vulkan/Spir-V toolchain, which is what we have in our codebase with all the legacy shaders.
HLSL is what Diligent recommends we use and what we will use going forward/what you have in `TestPipeline`

For runtime shader compiling Diligent can only use HLSL as the cross-API language. It can compile it to SPIRV or DX compatible bytecode. Diligent can runtime compile GLSL too, but can't convert GLSL to D3D compatible bytecode. Only HLSL goes both ways.

Any runtime shader compilation we do will be on shaders written in HLSL so we don't have to duplicate shaders, so we'll probably want to write our offline compiled shaders in HLSL as well

source/ncengine/graphics2/diligent/resource/GlobalResourceSignature.h
```diff
@@ -0,0 +1,43 @@
...
+
+namespace nc::graphics
+{
+class GlobalTextureBufferResource;
+
+class GlobalResourceSignature
+{
+    public:
+        static constexpr auto BindingIndex = 0u;
+        static constexpr auto SignatureName = "GlobalTextureBuffer";
```
jaremieromer: This signature will hold all of our global resources, not just Textures. I'd rename to something like "Global Resource Signature"

jaremieromer: If this isn't just a cleanup/old code thing and you're confused by my comment, see this discussion:
https://github.com/DiligentGraphics/DiligentEngine/issues/161
I'm imagining something like ^ this for this class.

Disregard if you are too

source/ncengine/graphics2/diligent/resource/GlobalTextureBufferResource.h
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        auto GetMaxTextureCount() const -> size_t
+        {
+            return m_maxTextures;
+        }
+
+    private:
+        std::vector<Diligent::RefCntAutoPtr<Diligent::ITexture>> m_textures;
+        std::vector<Diligent::IDeviceObject*> m_objects;
```
jaremieromer: I'm curious about why m_objects is of type `IDeviceObject` rather than `ITextureView`? I'm not %100, but that seems to be the type we're returning here. If it can't be/isn't an `ITextureView` for some reason we should probably still call this member `m_views`

source/ncengine/graphics2/AssetDispatch.h
```diff
@@ -0,0 +1,54 @@
...
+
+#include "Graphics/GraphicsEngine/interface/RenderDevice.h"
+#include "Graphics/GraphicsEngine/interface/DeviceContext.h"
+
+namespace nc::graphics
+{
+class AssetDispatch
+{
+    public:
+        AssetDispatch(Diligent::IDeviceContext& context,
```
jaremieromer, line 16: This class does exactly what I expect it to do and nothing more. Nice!

source/ncengine/graphics2/diligent/DiligentEngineWin32.cpp
```diff
@@ -76,6 +85,7 @@ DiligentEngine::DiligentEngine(const config::GraphicsSettings& graphicsSettings,
                     throw nc::NcError("Failed to create the D3D12 swapchain.");
                 }
 
+                m_shaderFactory = MakeShaderFactory(*pFactoryD3D12, *m_pDevice);
```
jaremieromer, line 88: I think we want to check for null and throw here for `m_shaderFactory` as well, in each of the API/Platform combos

jaremieromer (commented):
This is really great!!!! Nice job

jaremieromer (approved):

source/ncengine/graphics2/diligent/resource/GlobalResourceSignature.h
McCallisterRomer: Oops, copy paste error

source/ncengine/graphics2/diligent/resource/GlobalTextureBufferResource.h
McCallisterRomer: `m_views` is a better name. The type is `IDeviceObject` because that's what the `IShaderResourceVariable` functions operate on. `ITextureView` is derived from it.

source/ncengine/graphics2/diligent/DiligentEngineWin32.cpp
McCallisterRomer: That's our own type/function. It won't return null, and it throws if it fails to create a `IShaderSourceInputStreamFactory`.

cmake/FetchDependencies.cmake
McCallisterRomer: We need GLSLang (at least for vulkan) even if the source language is HLSL. For runtime compilation, the GLSL compiler is still used, which calls a `GLSLangUtils::HLSLtoSPIRV()` function internally to convert it. It barfs if this is disabled.

Also, we need the HLSL option enabled to support the SPV_GOOGLE_hlsl_functionality1 extension with vulkan backend.

PR #757, opened 2024-10-19
Add Validation Layers to Dockerfile

### LINKED ISSUE #749: Add full VulkanSDK to ubuntu container
Either add the official sdk or the artifacts from VulkanCI (if we want swiftshader too) to the ubuntu container. Enable validation layers in the integration tests.

McCallisterRomer:
resolves #749 

Adding validation layers to ubuntu dockerfile + enabling validation in integration tests. Instead of using libvulkan-dev and vulkan-validationlayers-dev, I added the latest sdk from the lunarg repo. This avoids having to deal with versioning issues between packages from the ubuntu repos.


jaremieromer (approved):
Nice! Just a note: I've made an [issue](https://github.com/NcStudios/NcEngine/issues/760) to wire up our `useValidationLayers` in the config.

PR #761, opened 2024-10-21
Logging Improvements + Rerouting Diligent Messages

McCallisterRomer:
Updates to logging to mostly address logging issues when `NcEngine`'s constructor throws and limited logging extensibility

Summary
- Logging macros now go through a configurable callback, which defaults to printing to stdout.
- Added `FileLogger` for basic file logging. This is almost exactly what we have now, but is opt-in and can operate outside of the engine's lifetime.
- Moved log-related options from config into `FileLogger`.
- Added a command line argument to the sample to use file logging. The default is stdout, but the smoke test writes to a file so we can publish it.
- Since things are a bit more publicly exposed now, I replaced the naked prefix characters with a `LogCategory` enum.
- Routing diligent messages through our logging.
  - Diligent messages don't always have file/line info, so I've added a subsystem indicator to the front of log entries to make the source less mysterious. Looks like:
    >> V [NcEngine] NcGraphicsImpl2.cpp:121: Building NcGraphics module
V [Diligent] User-defined allocator is not provided. Using default allocator.

### DIFF
--- a/include/ncengine/config/Config.h
+++ b/include/ncengine/config/Config.h
@@ -13,8 +13,6 @@ namespace nc::config
 struct ProjectSettings
 {
     std::string projectName = "Project Name";
-    std::string logFilePath = "NcEngine.log";
-    size_t logMaxFileSize = 1000000ull;
 };
 
 /** @brief Settings for configuring the engine run loop and executor. */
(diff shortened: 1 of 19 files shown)


jaremieromer:
This is ...really... awesome.
image

jaremieromer (approved):

PR #763, opened 2024-10-23
Add GlobalEnvironmentResource

McCallisterRomer:
Adding a resource for environment data to the global signature. It just has the camera, for now. This also required making the an analog of `PerFrameRenderState` and some initial 'system' handling. Names are different to avoid conflicts, but otherwise its pretty similar.


source/ncengine/graphics2/NcGraphicsImpl2.cpp
```diff
@@ -247,7 +251,9 @@ void NcGraphicsImpl2::Run()
     context.ClearRenderTarget(pRTV, &ClearColor.x, Diligent::RESOURCE_STATE_TRANSITION_MODE_TRANSITION);
     context.ClearDepthStencil(pDSV, Diligent::CLEAR_DEPTH_FLAG, 1.f, 0, Diligent::RESOURCE_STATE_TRANSITION_MODE_TRANSITION);
 
+    m_shaderBindings.Update(renderState, context);
```
McCallisterRomer, line 254: I'm not totally sure how we want to handle getting render state to its destination(s). There is very little happening now, but maybe this needs to get more sophisticated.

source/ncengine/graphics2/diligent/resource/GlobalResourceSignature.h
```diff
@@ -10,15 +10,18 @@
...
 class GlobalTextureBufferResource;
+class GlobalEnvironmentResource;
 
 class GlobalResourceSignature
 {
     public:
         static constexpr auto BindingIndex = 0u;
         static constexpr auto SignatureName = "GlobalResourceSignature";
         static constexpr auto GlobalTextureBufferShaderVariableName = "g_textures";
+        static constexpr auto GlobalEnvironmentShaderVariableName = "EnvironmentData";
```
McCallisterRomer, line 21: Also, I'm not sure what the best way to manage these variables is. Just spitballing, but maybe we could keep them in their own own file, and inject them into shaders with macros and include them here. There would at least be a single source of truth that way.

test/integration_test/ncengine/graphics2/GlobalEnvironmentResource_tests.cpp
```diff
@@ -0,0 +1,67 @@
...
+    EXPECT_NE(0, Diligent::RESOURCE_STATE_CONSTANT_BUFFER & actualBuffer->GetState());
+}
+
+TEST_P(GlobalEnvironmentResourceTest, Update_succeeds)
+{
+    auto cameraState = nc::graphics::CameraRenderState{
+        .viewProjection = DirectX::XMMatrixPerspectiveFovRH(90.0f, 16.0f / 9.0f, 0.1f, 100.0f)
+    };
+
+    // Buffer is dynamic/cpu write only, so we don't have a way of inspecting actual contents.
```
McCallisterRomer, line 60: Any ideas on how I can test this more thoroughly? I tried injecting a CPU_ACCESS_READ flag during tests, but it requires additional changes. Felt like the object under test was too far removed from the real deal. Diligent does seem pretty capable at detecting issues, so a 'no error output' test doesn't seem entirely trivial.

source/ncengine/graphics2/frontend/subsystem/CameraSubsystem.cpp
```diff
@@ -0,0 +1,45 @@
...
+namespace
+{
+const auto g_defaultViewProj = DirectX::XMMatrixMultiply(
+    DirectX::XMMatrixLookAtRH(
+        DirectX::g_XMIdentityR3,
+        DirectX::g_XMIdentityR2,
+        DirectX::g_XMNegIdentityR1
+    ),
+    DirectX::XMMatrixPerspectiveFovRH(
+        1.0472f,
```
jaremieromer: Near clip and far clip exist in the config, we should pull them from there. We currently have `AspectRatio` defined in `GraphicsUtilities.h`. We probably want to make that a config item too, but I'm fine with creating an issue for that. 

With FovAngleY, we should either put it in the config as well, or make a constant for it in `GraphicsUtilities.h`.
I'm fine with whatever you want to do now VS creating an issue for.

source/ncengine/graphics2/NcGraphicsImpl2.cpp
```diff
@@ -228,14 +234,12 @@ void NcGraphicsImpl2::OnBuildTaskGraph(task::UpdateTasks& update, task::RenderTa
...
 
-void NcGraphicsImpl2::Update()
-{
-}
-
 void NcGraphicsImpl2::Run()
 {
     NC_PROFILE_TASK("Render", Optick::Category::Rendering);
 
+    auto renderState = m_frontend.BuildRenderState(m_world);
```
jaremieromer, line 241: I think we should move this out of Run() and into a method called Update() that we create a new TaskFlow job for, with the expectation that everything in BuildRenderState is CPU-side only. The stuff in there is going to get pretty CPU intensive. We can still ensure that graphics update happens after physics update, but if we head in that direction we can move towards doing some of BuildRenderState in parallel for the individual subsystems

source/ncengine/graphics2/frontend/subsystem/AssetDispatch.h
jaremieromer, line 1: I didn't catch this before, but now that I have a better idea of the full picture, we'll need a layer in between the asset dispatch and the graphics buffer being modified/updated. 

I know you only had the original graphics implementation to go off of, but this was something that was just waiting to bite us in the butt (and had on many occasions) with the way I designed it there. 

The heart of the problem is that we don't know at asset event fire time whether the GPU is done using the buffer we are trying to modify. We _mostly_ got away with it by abusing the UPDATE_AFTER_BIND flag: https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkDescriptorBindingFlagBits.html, and it didn't show it's face all that often because we didn't yet have texture and mesh loading re-implemented in our new editor. But if you remember all the crashes with point lights being added and removed at runtime via the editor, this is that same issue. 
There is a very important difference between updating a buffer and and changing the size of an array. We could update the contents of a constant buffer without blowing (doesn't require an update of a descriptor set), but when adding or removing items from a mapped array (does require updating a descriptor set) it will blow.

Proposal:
1. Pull the `DiligentEngine` stuff out of `AssetDispatch`, but keep the signal forwarding there.
2. Add a class called `AssetLoader` (workshopping the name, I don't like that one) into NcGraphicsImpl2.
3. `AssetLoader` will receive the signal from `AssetDispatch`, and will temporarily cache the data (`asset::TextureUpdateEventData`, for example) and will flip a bit property called: `HasQueue` or `QueueNotEmpty` upon receiving the event.
4. Inside of `NcGraphicsImpl2::Run()` or preferably `NcGraphicsImpl2::Update()`, (called before Run), we check each frame   that: both the swapchain is done rendering AND the bit is set indicating that `AssetLoader` has items to process in its queue. If both true, we do the `m_textureBuffer->Load`/`m_textureBuffer->Unload` and clear the queue.

jaremieromer: We don't need to do this as part of this PR, by the way. Just something we want to do before we get to deep into other resource types.

source/ncengine/graphics2/NcGraphicsImpl2.cpp
```diff
@@ -247,7 +251,9 @@ void NcGraphicsImpl2::Run()
     context.ClearRenderTarget(pRTV, &ClearColor.x, Diligent::RESOURCE_STATE_TRANSITION_MODE_TRANSITION);
     context.ClearDepthStencil(pDSV, Diligent::CLEAR_DEPTH_FLAG, 1.f, 0, Diligent::RESOURCE_STATE_TRANSITION_MODE_TRANSITION);
 
+    m_shaderBindings.Update(renderState, context);
     m_shaderBindings.GetGlobalSignature().Commit(context);
+
     m_testPipeline.Render(context);
     swapChain.Present();
```
jaremieromer, line 258: From a quick glance, it looks like Diligent does wait on this Present() call until the commands in the buffer have finished execution. It seems safe to assume that we can map resources freely to our command queues/the GPU once this call is done or before `m_shaderBindings.GetGlobalSignature().Commit(context)`. That's awesome

source/ncengine/graphics2/diligent/resource/GlobalResourceSignature.h
jaremieromer: I like that idea - though I'm fuzzy on this part still but it seems that Diligent, at least when loading pipelines from DRSN, will patch the compiled byte code with the bindings specified in the signatures, so if you get it wrong or there is a collision in the HLSL it will be adjusted at DRSN package time.

jaremieromer, line 21: since we have to follow naming conventions of "g_" for textures, we should do the same for all our global shader resources, so I'd change this name to "g_environmentData"

test/integration_test/ncengine/graphics2/GlobalEnvironmentResource_tests.cpp
jaremieromer: I think it's really time for me to wire up an offscreen render target. The idea is you write to a RenderTarget that isn't the backbuffer, and now you have that memory to poke and prod at, as a texture. It will take some doing to get the image layouts to transition back to a place where we can interrogate it's contents, but I need to do that basic offscreen renderpass setup for shadows and the other shader anyway.

We would still need the full graphics pipeline though  to produce the output.

I think a 'test no throw' is fine for now

jaremieromer (approved):
Nice! It's really coming together. I have one major point re: synchronizing the asset loading. In general, I think if we aim for keeping the GPU out of `GraphicsFrontend` we're going to maintain a separation that we'll be thankful for later.

I'm approving because the big things are fine to do in a separate issue, and the other things are small. Up to you how you want to divvy up the PRs. I'm also fine if you want to roll them into this one.

Really nice work, it's all coming together in a nice way.

source/ncengine/graphics2/NcGraphicsImpl2.cpp
jaremieromer: This is great - provided we can guarantee we only map constant buffers here and don't require an update of a descriptor set (which happens when you change the size of an array). See my other comment on the layer between AssetDispatch and the GPU for more context.

test/integration_test/ncengine/graphics2/GlobalEnvironmentResource_tests.cpp
McCallisterRomer: Cool, issue created.

source/ncengine/graphics2/frontend/subsystem/CameraSubsystem.cpp
McCallisterRomer: I'm going to remove near/far clip from the config. They have been properties of the `Camera` for a while, so the config values are not used anywhere. FOV doesn't need be constant or config-based. Different camera's can have different values. I've also done some experimentation with having dynamic FOV based on zoom, which is quite nice.

Aspect ratio probably makes sense here. The expectation is that this matrix won't get used in a shipped product, but just exists as a fallback in case the camera is deleted from the editor or something. Its not a `Camera` property though, so I think we should try to keep it consistent. I'll see if I can sneak it in here.

source/ncengine/graphics2/NcGraphicsImpl2.cpp
McCallisterRomer: Yeah, that's the direction I was thinking as well. Renamed/moved an existing issue to cover this.

source/ncengine/graphics2/diligent/resource/GlobalResourceSignature.h
McCallisterRomer: I'm not aware of any convention requiring the "g_" prefix (there is the "_sampler" suffix). I just left the prefix when copy pasting from some sample code. Correct me if there's something I don't know, but I think we can remove it if we want.

source/ncengine/graphics2/frontend/subsystem/AssetDispatch.h
McCallisterRomer: Hmm, this is potentially a larger design issue. As it stands we can't queue up the texture data, since the event contains a non-owning view, and is destroyed after the event is fired. We can change that, but I've started down that path, and its a nightmare to also maintain the existing graphics implementation.

Also, it isn't immediately obvious what other implications this might have. If loading an asset isn't always completed after returning from the call, maybe the flow should be restructured into an async operation. Then again, maybe that can be ignored as an implementation detail. The underlying views wouldn't be used in a draw call until the buffers were actually updated.

Will make an issue.

McCallisterRomer: You know, if we're going to wait on the swapchain during the next go round in `NcGraphicsImpl2`, can we just move that block into `AssetDispatch` and call it a day?

source/ncengine/graphics2/diligent/resource/GlobalResourceSignature.h
jaremieromer: Nope - I thought the "g_" was required as well. If it isn't, let's name them whatever we want, so long as they're consistent with one another

PR #770, opened 2024-10-26
Add MeshBuffer

### LINKED ISSUE #764: Vertex/Index Buffers

McCallisterRomer:
resolves #764 

Adding `MeshBuffer` to manage vertex/index buffers.


source/ncengine/graphics2/diligent/resource/MeshBuffer.h
```diff
@@ -0,0 +1,35 @@
...
+namespace asset
+{
+struct MeshVertex;
+} // namespace asset
+
+namespace graphics
+{
+auto GetMeshVertexLayoutElements(uint32_t slot = 0u, uint32_t indexOffset = 0u) -> std::array<Diligent::LayoutElement, 7>;
+
+class MeshBuffer
```
jaremieromer: I'm leaning towards naming this the same as the other global resources, `GlobalMeshBufferResource` or adding a folder in resources called `global` and changing the other names so it looks like:
```
-resources
    -global
         -MeshBuffer
         -TextureBuffer
         -EnvironmentBuffer
```
Even though `MeshBuffer` is not bound the same as the other global resources, it makes the most sense to me to categorize as a global resource because functionally that's what it is, it is shader input that will be shared by many pipelines.

I don't think its necessary to draw the distinction (its implementation being different from the others) at the folder or filename level, because API-readability-wise, if I want to know what common shader inputs are available/predefined at the system level, I'd want them all to be grouped together and named something similar.

jaremieromer: If or when we get to the point of having user-defined buffers and shader reflection, I'd imagine a folder structure like this:

```
-resources
    -global
         -MeshBuffer
         -TextureBuffer
         -EnvironmentBuffer
     -user
         -ConstantBuffer
         -StructuredBuffer
```

jaremieromer (approved):

PR #773, opened 2024-10-27
Add Diligent-Imgui

### LINKED ISSUE #771: Add Diligent-Imgui to build
There's a few things to work out with this.
- We want to target our own version b/c there are some breaking changes and I'm pretty sure we're using some features that require 1.9-ish.
- ImTextureID was changed at some point from void* to int64. I think this will be a small problem.
- We'll at least need to bring in imgui_impl_win32 - probably makes sense to fix the current strategy for how imgui is included. (its a bit of a pain to update version; would be nice to just get through fetch_content.)

McCallisterRomer:
resolve #771

Changes to get Diligent-Imgui building:
- replaced vendored dear imgui source with fetch_content
  - updated to latest imgui version, with a few tweaks for backwards compatibility with diligent's expectations
- updated imgui includes to match actual repo structure
- renamed our `Imgui.h` to `ImguiVulkan.h`. Without the `imgui/` parent directory, `#include "imgui.h"` and `#include "Imgui.h"` are not distinguishable.


jaremieromer (approved):
Very nice work!

PR #780, opened 2024-11-01
Add Dearchiver

McCallisterRomer:
Adding `Dearchiver` wrapper. Its fairly simple; mostly its just doing error checking. Next, I'll probably start looking at incorporating this into some kind of factory that assembles our core resource + pso objects.


source/ncengine/graphics2/diligent/Dearchiver.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+namespace
+{
+void UnpackGraphicsPipelineCB(Diligent::PipelineStateCreateInfo& ci, void* userData)
+{
+    NC_ASSERT(ci.PSODesc.PipelineType == Diligent::PIPELINE_TYPE_GRAPHICS, "Pipeline type mismatch");
+    auto& upcastCI = static_cast<Diligent::GraphicsPipelineStateCreateInfo&>(ci);
+    auto& graphicsPipeline = upcastCI.GraphicsPipeline;
+    auto* swapChain = static_cast<Diligent::ISwapChain*>(userData);
+    graphicsPipeline.NumRenderTargets = 1;
```
jaremieromer, line 23: Are the count and format of render targets not settable in the DRSN? You won't know this, but for shadow mapping, normal mapping and depth passes we will need configurable render targets. If they're not settable, you can leave this and I will hardcode those conditions when we come to that bridge.

jaremieromer (approved):

Issue #782, opened 2024-11-05
Parameterize Signature + Variable Names


McCallisterRomer:
For discoverability and such, we probably want all resource signature, shader variable, and binding slots defined/configurable in one place. I'd argue this should also go in `ShaderTypes`. Parameterizing `ShaderBindings` with something like the following would do:

```
struct ResourceBindingConfig
{
    std::string bufferVariableKey;
    uint32_t maxInstances;
    uint32_t initialSize;
};

struct ShaderBindingConfig
{
    std::string signatureKey; // EG: "MaterialResourceSignature";
    uint8_t bindingIndex = 1;
   std::vector<ResourceBindingConfig> resourceConfigs; // TBD
    // ...
};
```

We could instead continue to use constants, but they should be in one place. Either way, each downstream resource/signature needs to be parameterized, if it isn't already.

If we don't want them in the engine's config, this would be a good place to put size hints as well:
```
uint32_t materialBufferInitialCountHint = 1000u;
```

PR #783, opened 2024-11-05
Add MaterialInstance

McCallisterRomer:


include/ncengine/graphics/Material.h
```diff
@@ -0,0 +1,109 @@
...
+
+namespace nc
+{
+namespace graphics
+{
+class MaterialRegistry;
+} // namespace graphics
+
+/** @brief Identifier for a MaterialInstance. */
+using MaterialInstanceHandle = uint16_t;
```
jaremieromer, line 21: What's the thought for not having these in the graphics namespace?

source/ncengine/graphics2/diligent/resource/ShaderBindings.h
```diff
@@ -13,10 +14,20 @@ struct FrontendRenderState;
...
-            : m_globalSignature{device, context, maxTextures}
+                                uint32_t maxTextures,
+                                uint32_t initialMaterialSizeHint)
+            : m_globalSignature{device, context, maxTextures},
+              m_materialSignature{
+                context,
+                device,
+                "MaterialInstanceSignature",
+                "MaterialDataBuffer",
+                1,
```
jaremieromer: Let's move this binding index to 2. It makes sense to me to put ComponentResourceSignature at 1, because it's the second most general grouping.
Also, let's extract these to a constant and put them in `ShaderBindings`

jaremieromer (approved):

source/ncengine/graphics2/diligent/resource/ShaderBindings.h
McCallisterRomer: Updated to 2, but see the comment above about #782. I want to parameterize `ShaderBindings` with a struct to hold names/indices (or it could just be a bunch of constants). Those also feel like part of the interface and shouldn't be buried in diligent/backend, but probably in `ShaderTypes`. (??) Anyway, there is an issue to collect all of those variables together once this and your component resource are in.

include/ncengine/graphics/Material.h
McCallisterRomer: Same reason for removing the physics namespace from things in include/. One namespace for most things in the API is sufficient, and we can organize/prevent name collisions with per-module namespaces in source/. When actually using the engine for a while, the additional namespaces (`physics` and `graphics` in particular) felt redundant and cumbersome:
```
world.Emplace<nc::graphics::MeshRenderer>(
    entity,
    myMesh,
    nc::graphics::MaterialDesc{
        "myMaterial",
        nc::graphics::MaterialPass::Toon | nc::graphics::MaterialPass::Outline,
        ...
    }
);
```

PR #788, opened 2024-11-08
Adjustments to SwapChain + Preferred Api

### LINKED ISSUE #734: Write the supported render API back to GraphicsSettings
Since MR #719 covers so much ground, I chose not to include the work of removing `const` from the config passed through to `NcGraphicsImpl2` for writing the supported render API back to the config.

I was planning on un-const-qualifying `GraphicsSettings` in `NcGraphicsImpl2`, but that would propagate all the way up through `ModuleFactory` into `NcEngineImpl` (and possibly further).

Maybe we want to take a different approach. Either way, I've decided not to do that as part of the initial Diligent integration MR.

### LINKED ISSUE #778: Improve Headless Mode/Avoid Null SwapChain
Leaving the swapchain null when `isHeadless == true` is going to break a bunch of downstream things. We want a better strategy. It probably makes more sense to make headless mean a headless window with `useGraphics == false`, so there are no graphics objects at all.

McCallisterRomer:
resolves #778, resolves #734

- Removing 'headless' mode from graphics + unconditionally creating the swapchain. (to me, `useGraphics == false` is a proper enough headless mode, I think we talked about this a bit).
- Adding method to `NcGraphics` to expose the actual selected api and renaming the `GraphicsSettings` api member to `preferredApi`.

Let thy concerns be voiced.

### DIFF
--- a/include/ncengine/graphics/NcGraphics.h
+++ b/include/ncengine/graphics/NcGraphics.h
@@ -95,6 +95,13 @@ struct NcGraphics : public Module
      * is not cleared as it can be set on a persistent Entity.
      */
     virtual void ClearEnvironment() = 0;
+
+    /**
+     * @brief Get the graphics api being used.
+     * @note Typically, this will be equal to GraphicsSettings::preferredApi, but insufficient device support or
+     *       initialization failure may result in a fallback api being selected.
+     */
+    virtual auto GetApi() const noexcept -> std::string_view = 0;
 };
 
 /**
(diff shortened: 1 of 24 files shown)


include/ncengine/graphics/NcGraphics.h
```diff
@@ -95,6 +95,13 @@ struct NcGraphics : public Module
      * is not cleared as it can be set on a persistent Entity.
      */
     virtual void ClearEnvironment() = 0;
+
+    /**
+     * @brief Get the graphics api being used.
+     * @note Typically, this will be equal to GraphicsSettings::preferredApi, but insufficient device support or
```
jaremieromer, line 101: Since this comment will be seen in our public interface will you capitalize API throughout comments in this file?

source/ncengine/graphics/NcGraphicsImpl.h
```diff
@@ -31,6 +31,7 @@ class NcGraphicsImpl : public NcGraphics
                        std::unique_ptr<IGraphics> graphics,
                        window::NcWindow& window);
 
+        auto GetApi() const noexcept -> std::string_view override { return "vulkan"; }
```
jaremieromer: Anywhere we reference the API by a string that needs to match the config value, we should use `graphics::api::Vulkan` (see `NcGraphics.h`). Feel free to remove the `api` namespace there as part of scrubbing the namespaces out of our public include.

jaremieromer (approved):

PR #790, opened 2024-11-10
Remove D3D12

McCallisterRomer:
Axing code. I also combined the platform-specific `DiligentEngine` source files, and instead added platform-specific window handling - other than the glfw calls, engine creation is identical.


source/ncengine/graphics2/diligent/DiligentEngine.cpp
```diff
@@ -0,0 +1,59 @@
...
+    {
+        pFactoryVk->SetMessageCallback(logCallback);
+    }
+
+    auto engineCI = EngineVkCreateInfo{engineCreateInfo};
+    pFactoryVk->CreateDeviceAndContextsVk(engineCI, &m_pDevice, &m_pImmediateContext);
+    if (!m_pDevice || !m_pImmediateContext)
+    {
+        EnsureContextFlushed(m_pImmediateContext);
+        throw nc::NcError("Failed to create the Vulkan device or context.");
```
jaremieromer: I don't think we really need "Vulkan" in any of these error messages anymore

jaremieromer (approved):
Really nice work. I'm amazed you did this so quickly!

PR #795, opened 2024-11-11
Cleanup and parameterize resources 

### LINKED ISSUE #782: Parameterize Signature + Variable Names

jaremieromer:
resolves #782

Cleaned up the interfaces for our resource signatures so all relevant variables are parameterized and exposed in `ShaderBindings`.


source/ncengine/graphics2/diligent/resource/ShaderBindings.h
```diff
@@ -15,29 +16,29 @@ struct FrontendRenderState;
...
                                 uint32_t initialMaterialSizeHint)
-            : m_globalSignature{device, context, maxTextures},
-              m_componentSignature{context,
-                device,
+            : m_globalSignature{
+                context, device,
+                "GlobalResourceSignature",
+                0,
+                TextureBufferResourceDesc{"TextureBufferData", Diligent::SHADER_TYPE::SHADER_TYPE_PIXEL, maxTextures},
+                UniformBufferResourceDesc{"EnvironmentBufferData", Diligent::SHADER_TYPE::SHADER_TYPE_VERTEX}
```
McCallisterRomer, line 29: Is there a better way for us to deal with the shader types? I don't love that here we specify `SHADER_TYPE_VERTEX`, but `GlobalEnvironmentResource::MakeResourceDesc()` sets `SHADER_TYPE_VS_PS`. There is the variable access problem to deal with though... What if we moved the `XXXResource::MakeDesc()` functions from each resource into `ShaderTypes.h` as free functions that take in our desc types. Ex:
```
inline auto MakeResourceDesc(const UniformBufferResourceDesc& desc) -> Diligent::PipelineResourceDesc;`
```

Then we had another free function in there to extract a single shader stage from the `SHADER_TYPE` flags for variable access:
```
auto GetShaderStageForVariableAccess(SHADER_TYPE stages)
{
    if (stages & SHADER_TYPE_VERTEX)
        return SHADER_TYPE_VERTEX;
    else if(stages & SHADER_TYPE_PIXEL)
        return SHADER_TYPE_PIXEL;
    // probably don't care about others for now...
    else
        return SHADER_STAGE_UNKNOWN;
}
```

source/ncengine/graphics2/diligent/resource/ResourceTypes.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include "Graphics\GraphicsEngine\interface\GraphicsTypes.h"
+
+#include <string>
+
+namespace nc::graphics
+{
+struct StructuredBufferResourceDesc
```
McCallisterRomer, line 10: Desc members should probably have defaults, even if they're 'invalid' states like SHADER_TYPE_UNKNOWN.

source/ncengine/graphics2/diligent/resource/ShaderBindings.h
```diff
@@ -15,29 +16,29 @@ struct FrontendRenderState;
...
-        /** @todo: 782 Fully parameterize names, indices, and size hints */
         explicit ShaderBindings(Diligent::IRenderDevice& device,
                                 Diligent::IDeviceContext& context,
                                 uint32_t maxTextures,
                                 uint32_t maxMeshRenderers,
                                 uint32_t initialMaterialSizeHint)
-            : m_globalSignature{device, context, maxTextures},
-              m_componentSignature{context,
-                device,
+            : m_globalSignature{
```
McCallisterRomer, line 24: Much better!

McCallisterRomer (approved):

PR #799, opened 2024-11-13
MaterialPass Frontend/Backend

### LINKED ISSUE #797: MaterialPass System
Design system for routing draw call information to specific material passes. Initial work can just reimplement `TestPipeline` as a pass (it'll be the only pass for now).

McCallisterRomer:
resolves #797 

Updates to issue draw calls based on material pass:
- `MaterialPassCache` tracks the association of `MeshRenderer2` properties and enabled passes.
- `MaterialPassBackend` owns PSOs and makes draw calls based on the render state produced by `MaterialPassCache`.

 For now, there's only one enabled pass, which is copy-pasted from `TestPipeline`.


jaremieromer (approved):
I've been looking for a few hours, and cannot find anything to say about this one. Fantastic job, I can't believe you did it!

PR #800, opened 2024-11-16
Fix Bug in RigidBody Batching

McCallisterRomer:
During the `RigidBody` batch add process, `AddBodiesPrepare()` can modify the input array. Updating to pass a copy of the new ids, instead of a pointer into our map.


jaremieromer (approved):
This is a very impressive find and fix. nice job!

Issue #798, opened 2024-11-12
Static MeshRenderer Optimizations


McCallisterRomer:
Add persistent tracking of `MeshRenderer`/`Material` information to `MeshRendererSubsystem` and friends. At the very least, static objects should be tracked on add/remove and ignored in the per-frame loop. We could consider tracking everything via on add/remove. GPU updates should only be done for dirty buffer subranges. `MeshRenderer` will need a way to notify the backend when its mesh is changed.

PR #801, opened 2024-11-16
Disallow Changing a Material's Passes + add SetMaterial()

McCallisterRomer:
prereq for #798 

Making passes immutable after material construction and adding `MeshRenderer2::SetMaterial()` as an alternative way to update passes for an object. Also splitting out shader-visible values from `MaterialDesc` into a `MaterialProperties` struct so API getters/setters are little clearer with respect to mutability.

Once we start caching instance/static instance data, it becomes really obnoxious to remap objects to new passes when they change on the material side. Somebody (like the material) needs to know all of the `MeshRenderer`s using a particular material, so that on update, we can go and shuffle around each affected object's cached state. How we actually want to do this is not by changing the material's passes, but simply assigning a new material to a `MeshRenderer`.

### DIFF
--- a/include/ncengine/graphics/Material.h
+++ b/include/ncengine/graphics/Material.h
@@ -103,6 +103,8 @@ class MaterialInstance
         static inline graphics::MaterialRegistry* s_registry = nullptr;
 
         MaterialInstanceHandle m_handle;
+
+        void Release() noexcept;
 };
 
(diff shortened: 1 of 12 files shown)


jaremieromer (approved):
It's really starting to gel for me now. Nice work!

PR #810, opened 2024-11-22
Add InstanceCache

McCallisterRomer:
part of #808 

Adding `InstanceCache`, which manages state required for instancing `MeshRenderer`s. Just adding the type + tests for now because its a lot of code just for that.

### DIFF
--- a/source/ncengine/graphics2/frontend/subsystem/InstanceCache.h
+++ b/source/ncengine/graphics2/frontend/subsystem/InstanceCache.h
@@ -182,7 +182,7 @@ class InstanceCache
         sparse_map<uint32_t> m_entityToIndex;
         std::vector<uint32_t> m_indexToEntity;
         InstanceCacheStaging m_stagingArea;
-        uint32_t m_dirtyBegin = UINT_MAX;
+        uint32_t m_dirtyBegin = NullBatchIndex;
         uint32_t m_initialBatchSize;
 
         void InsertInstance(uint32_t index, const StagedInstance& instance);
--- a/source/ncengine/graphics2/frontend/subsystem/InstanceCache.h
+++ b/source/ncengine/graphics2/frontend/subsystem/InstanceCache.h
@@ -1,0 +1,198 @@
+#pragma once
+
+#include "graphics2/ShaderTypes.h"
+#include "graphics2/frontend/subsystem/MeshRendererRenderState.h"
+
+#include "ncengine/asset/AssetViews.h"
+#include "ncengine/utility/SparseMap.h"
+
+#include <limits>
+#include <ranges>
+#include <vector>
+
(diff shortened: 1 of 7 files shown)


source/ncengine/graphics2/frontend/subsystem/InstanceCache.h
```diff
@@ -0,0 +1,198 @@
...
+class InstanceCacheStaging
+{
+    public:
+        // API-Facing Functions
+        void AddInstance(uint32_t entityId,
+                         MaterialPasses passes,
+                         const asset::MeshView& mesh,
+                         const InstanceData& instanceData);
+
+        void RemoveInstance(uint32_t entityId,
```
jaremieromer, line 77: I don't have a problem with it just curious: what's the reasoning behind leaving the batches around when they've been removed of all their members?

source/ncengine/graphics2/frontend/subsystem/InstanceCache.h
```diff
@@ -0,0 +1,198 @@
...
+            return m_batches.size();
+        }
+
+    private:
+        std::vector<InstanceData> m_buffer;
+        std::vector<BatchRegion> m_batches;
+        sparse_map<uint32_t> m_entityToIndex;
+        std::vector<uint32_t> m_indexToEntity;
+        InstanceCacheStaging m_stagingArea;
+        uint32_t m_dirtyBegin = UINT_MAX;
```
jaremieromer: What did we decide on UINT_MAX vs std::numeric_limits<uint32_t>::max()? Obviously one is fewer characters, so I'm fine with it. I'm actually curious why both exist.

jaremieromer:
This is the work that I'm most impressed with that you've ever done. Really cool

jaremieromer (approved):

source/ncengine/graphics2/frontend/subsystem/InstanceCache.h
McCallisterRomer: Its to minimize work. Thinking being that if a batch existed once, it'll probably get used again. By leaving it around, those spaces on the gpu are marked for reuse later, and we don't have to send anything over the wire on removal. If/when a new object is added to the batch in the future, there's already a space reserved for it.

They are removed on scene change though, because a batch could be based on a mesh that's specific to a scene.

source/ncengine/graphics2/frontend/subsystem/InstanceCache.h
McCallisterRomer: Fixed. I usually use `UINT_MAX` when POCing things and fix later. There's a small emotional burden to breaking flow state to include `<limits>` that I just can't afford :)

PR #811, opened 2024-11-24
Add HostBuffer

McCallisterRomer:
required for #808 and #798

Adding a `HostBuffer` type to manage cpu-side of a `StructuredBuffer` for the common case (where buffer has stable/unsorted elements and can be sparse). In my in-progress instancing work, I had to duplicate most of the code in `MaterialRegistry` for managing matrices. Decided to pull the common stuff out into its own class. I've also updated `MaterialRegistry` accordingly here.

In addition to existing buffer code, I've also adding a 'staging' feature, which is required for decoupling the update/render graphs.


jaremieromer (approved):
This is great. How do you feel about calling it HostStructuredBuffer?

McCallisterRomer:
> This is great. How do you feel about calling it HostStructuredBuffer?

Thanks! Yes, that name more narrowly describes the thing. Will change it.

Issue #809, opened 2024-11-22
Pass Material Updates Through to InstanceCache


McCallisterRomer:
- need a way for material pass changes to get into `InstanceCache`
- current bug: since passes are not updated here, modifying a mesh when your passes have changed is probably really bad
- don't know if we can handle empty passes currently, we don't really want to make a batch in this case

PR #813, opened 2024-11-25
Finish Instancing

### LINKED ISSUE #798: Static MeshRenderer Optimizations

### LINKED ISSUE #808: Instanced Draw Calls

McCallisterRomer:
resolves #808, resolves #798

Instancing all draw calls + minimizing gpu transfers of static matrices.
- Added `TransformCache` to track up update matrices.
  - Swapped our `StructuredBuffer<MeshRendererData>` with a `StructuredBuffer<TransformData>`
- Integrated `InstanceCache` into the frontend and `Batch` into the pass backend.
- Updated `MeshRenderer2` to only hold its mesh id, instead of the whole `MeshView` (among other changes for instancing).
- Updated `Benchmark` scene to work with new `MeshRenderer2`.


include/ncengine/graphics/MeshRenderer2.h
```diff
@@ -10,31 +10,70 @@
...
         auto GetMaterial() const -> const MaterialInstance& { return m_material; }
         auto GetMaterial() -> MaterialInstance& { return m_material; }
         void SetMaterial(const MaterialDesc& materialDesc) { m_material = MaterialInstance{materialDesc}; }
 
     private:
-        asset::MeshView m_mesh;
+        friend class graphics::MeshRendererSubsystem;
+        inline static graphics::MeshRendererSubsystem* s_subsystem = nullptr;
+        Entity m_self;
+        uint64_t m_meshId;
```
jaremieromer, line 72: `uint64_t` might not be big enough to hold all of our meshes with our new instancing combined with the culling I'm going to do.
It only represents 18,446,744,073,709,551,615 values

source/ncengine/graphics2/MeshRenderer2.cpp
```diff
@@ -0,0 +1,50 @@
...
+{
+    m_transformDataHandle = s_subsystem->AddInstance(
+        self,
+        m_material.GetHandle(),
+        materialDesc.passes,
+        mesh
+    );
+}
+
+void MeshRenderer2::SetMesh(const asset::MeshView& mesh)
```
jaremieromer, line 22: What is the plan with the UI for the Material -> Passes section? I know we were changing material passes to be uneditable after creation, so should we change the UI so that the Material -> Passes section is read only? Right now, if you uncheck any pass in that section, and then change the mesh, we crash

source/ncengine/graphics2/NcGraphicsImpl2.cpp
```diff
@@ -249,15 +254,15 @@ void NcGraphicsImpl2::Run()
 {
     NC_PROFILE_TASK("Render", Optick::Category::Rendering);
 
-    auto renderState = m_frontend.BuildRenderState(m_world);
-
     auto& context = m_engine.GetContext();
```
jaremieromer, line 258: Unrelated to this work, but so I don't forget: since this is our frame loop do we get any benefit from caching the pointers to context, device and swapchain here?

source/ncengine/graphics2/ShaderTypes.h
```diff
@@ -24,18 +24,16 @@ struct GlobalEnvironmentData
...
 };
 
-// Object model for MeshRenderers (type: StructuredBuffer element type).
-struct MeshRendererData
+// Object model for Transforms (type: StructuredBuffer element type).
+struct TransformData
 {
-    DirectX::XMMATRIX modelMatrix = DirectX::XMMatrixIdentity(); // Transforms the object from object space to world space.
-    uint32_t materialIndex = NullMaterialInstanceHandle;
+    DirectX::XMMATRIX matrix = DirectX::XMMatrixIdentity();
```
jaremieromer: If we're going to have just one - I prefer `model` over `matrix` here

source/ncengine/graphics2/diligent/pass/MaterialPassBackend.cpp
```diff
@@ -1,47 +1,49 @@
...
-        target.indexCount,
+        batch.indexCount,
         Diligent::VT_UINT32,
         drawFlags,
-        1,
-        target.firstIndex,
-        target.firstVertex,
-        target.instance
+        batch.instanceCount,
+        batch.indexOffset,
```
jaremieromer: I like sticking with VK conventions for naming for `Batch`. Thoughts?
```
    uint32_t    indexCount;
    uint32_t    instanceCount;
    uint32_t    firstIndex;
    int32_t     vertexOffset;
    uint32_t    firstInstance;
```

source/ncengine/graphics2/diligent/pass/MaterialPassBackend.cpp
```diff
@@ -1,47 +1,49 @@
...
 {
 void MaterialPassBackend::Render(Diligent::IDeviceContext& context,
-                                 const std::vector<PassRenderState>& passStates)
+                                 const std::vector<std::vector<Batch>>& passBatches)
 {
-    NC_ASSERT(m_passes.size() == passStates.size(), "Frontend/Backend passes out of sync.");
-    for (auto [pass, state] : std::views::zip(m_passes, passStates))
+    NC_PROFILE_SCOPE("MaterialPassBackend::Render()", ProfileCategory::Rendering);
+    NC_ASSERT(m_passes.size() == passBatches.size(), "Frontend/Backend passes out of sync.");
+    for (auto [pass, batches] : std::views::zip(m_passes, passBatches))
```
jaremieromer, line 43: This really came together nicely

source/ncengine/graphics2/diligent/pass/Pass.cpp
```diff
@@ -125,12 +132,14 @@ cbuffer EnvironmentBufferData
...
 void main(in  VSInput VSIn, uint InstanceID : SV_InstanceID,  out PSInput PSIn)
 {
-    float4 TransformedPos = mul(float4(VSIn.Pos, 1.0), MeshRendererBufferData[InstanceID].model);
+    uint transformIndex = InstanceBufferData[InstanceID].transformIndex;
+    uint materialIndex = InstanceBufferData[InstanceID].materialIndex;
+    float4 TransformedPos = mul(float4(VSIn.Pos, 1.0), TransformBufferData[transformIndex].model);
     PSIn.Pos = mul(TransformedPos, cameraViewProjection);
     PSIn.UV  = VSIn.UV;
-    PSIn.Normal = normalize(mul(MeshRendererBufferData[InstanceID].model, VSIn.Normal)); // @TODO #805, compute inverse model matrix CPU-side
+    PSIn.Normal = normalize(mul(TransformBufferData[transformIndex].model, VSIn.Normal)); // @TODO #805, compute inverse model matrix CPU-side
```
jaremieromer: With this new architecture - should extra 'Per Instance' data live in `TransformData` now? (In this case it is a transform that would need to be added: the inverse transpose of the model matrix) I may shift to using the TBN matrix and doing lighting in tangent space anyway- so if it throws a wrench in the gears I can go that route.

source/ncengine/graphics2/diligent/resource/ShaderBindings.h
```diff
@@ -26,12 +26,13 @@ class ShaderBindings
                 context, device,
                 "PerFrameResourceSignature",
                 0,
-                StructuredBufferResourceDesc{"MeshRendererBufferData",     Diligent::SHADER_TYPE::SHADER_TYPE_VS_PS, maxMeshRenderers,     maxMeshRenderers},
-                StructuredBufferResourceDesc{"DirectionalLightBufferData", Diligent::SHADER_TYPE::SHADER_TYPE_VS_PS, maxDirectionalLights, maxDirectionalLights},
-                StructuredBufferResourceDesc{"PointLightBufferData",       Diligent::SHADER_TYPE::SHADER_TYPE_VS_PS, maxPointLights,       maxPointLights},
-                StructuredBufferResourceDesc{"SpotLightBufferData",        Diligent::SHADER_TYPE::SHADER_TYPE_VS_PS, maxSpotLights,        maxSpotLights},
-                StructuredBufferResourceDesc{"MaterialBufferData",         Diligent::SHADER_TYPE::SHADER_TYPE_PIXEL, maxMeshRenderers,     initialMaterialSizeHint},
-                TextureBufferResourceDesc{"TextureBufferData",             Diligent::SHADER_TYPE::SHADER_TYPE_PIXEL, maxTextures},
+                StructuredBufferResourceDesc{"TransformBufferData",        Diligent::SHADER_TYPE::SHADER_TYPE_VERTEX, maxMeshRenderers,     maxMeshRenderers / 2},
```
jaremieromer: what's with the /2 here? It seems a little arbitrary to me.

source/ncengine/graphics2/frontend/subsystem/TransformCache.h
```diff
@@ -0,0 +1,46 @@
...
+class Entity;
+class Transform;
+
+namespace graphics
+{
+// Unique identifier for a TransformData instance.
+using TransformDataHandle = HostStructuredBufferHandle;
+
+// CPU-side buffer for Transform matrices. Effectively has the same behavior as a StructuredHostBuffer, but needs to be
+// synced with the registry each frame via UpdateMatrices(), which may NOT run concurrently with game logic.
```
jaremieromer, line 20: You've really been bringing the heat lately.

jaremieromer (approved):

source/ncengine/graphics2/MeshRenderer2.cpp
McCallisterRomer: See #809. I can sort this out next after benchmark crash. Material passes cannot be edited through the instance, but they can be quasi-edited by recreating the instance with the desired passes. I think we want to use that workflow in the editor. I just need to get that event into the `InstanceCache` somehow.

source/ncengine/graphics2/NcGraphicsImpl2.cpp
McCallisterRomer: No, really just context is referenced a bunch of times.

source/ncengine/graphics2/diligent/pass/Pass.cpp
McCallisterRomer: Generally, its a toss up between `TransformData` and `InstanceData`. As discussed, we'll probably end up with variants of `InstanceData` (mesh vs. skinned renderer, etc) with slightly different properties, so it probably depends on whether it pertains to all instance types or a subset. For an additional matrix, I'd assume `TransformData`.

source/ncengine/graphics2/diligent/resource/ShaderBindings.h
McCallisterRomer: Eh, I wasn't sure what to do here. Its the initial buffer size. It seemed unnecessary to allocate the max buffer, since we can resize. It would be silly to allocate 0, because there will certainly be some data. It also didn't seem like a meaningful value to set through the config. Maybe I'll make it a parameter and pass it in. It should probably be set the same as initialMaterialSizeHint for now.

source/ncengine/graphics2/diligent/pass/MaterialPassBackend.cpp
McCallisterRomer: That's fine. I didn't do it because I can't reason about why both 'offset' and 'first' are used.

source/ncengine/graphics2/ShaderTypes.h
McCallisterRomer: reverted

PR #816, opened 2024-11-26
Extend ShaderFactory For Bytecode

### LINKED ISSUE #753: Add bytecode loading to ShaderFactory

McCallisterRomer:
resolves #753


jaremieromer (approved):
NICE!!!

source/ncengine/graphics2/diligent/ShaderFactory.cpp
```diff
@@ -46,12 +46,29 @@ auto ShaderFactory::MakeShaderFromSource(std::span<const char> source,
     createInfo.SourceLanguage = language;
     createInfo.CompileFlags = Diligent::SHADER_COMPILE_FLAG_PACK_MATRIX_ROW_MAJOR |
                               Diligent::SHADER_COMPILE_FLAG_ENABLE_UNBOUNDED_ARRAYS;
+    return CreateShader(createInfo);
+}
 
+auto ShaderFactory::MakeShaderFromByteCode(std::span<const char> byteCode,
```
jaremieromer, line 52: Does this assume SPIR-V? I know we're planning on shifting back towards GLSL, but since we're suggesting support for both HL and GL with these methods - we should test if this succeeds with DXIL/DXBC. 

If not - DXC has an argument to compile into SPIR-V, but we should make a note or name the method accordingly

McCallisterRomer: Expectation is set based on the backend, so spirv-only for bytecode version. Will add comments.

PR #819, opened 2024-11-27
Notify InstanceCache On MaterialPass Update

### LINKED ISSUE #809: Pass Material Updates Through to InstanceCache

McCallisterRomer:
resolves #809 

`InstanceCache` needs to know when a new `MaterialInstance` is assigned to a `MeshRenderer` so it can update the index in the instance buffer. I could have done this in the initial PR - thought there was more to it for some reason.


jaremieromer (approved):
Nice!

PR #822, opened 2024-11-29
Cal/post process pass

McCallisterRomer:
@jaremieromer Here's my idea for handling post processing passes and their properties. There are a few things left to implement and cleanup. Completely untested as of now, but the general structure is all there. Let me know if you have any thoughts.


source/ncengine/graphics2/diligent/pass/PostProcessPassBackend.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+    for (const auto& [effectId, passId, properties] : postProcessState.modifiedProperties)
+    {
+        CopyProperties(properties, FindInstance(m_passes, effectId, passId));
+    }
+}
+
+void PostProcessPassBackend::Render(Diligent::IDeviceContext& context, PostProcessBufferResource& resource)
+{
+    for (auto& pass : m_passes)
```
McCallisterRomer, line 105: @jaremieromer  Can you verify this loop fits with what you're doing?

source/ncengine/graphics2/GraphicsUtility.cpp
```diff
@@ -1,10 +1,11 @@
 #include "ncengine/graphics/GraphicsUtility.h"
 
 #include <array>
+#include <vector>
 
 namespace
 {
-constexpr auto g_passNames = std::array{
+constexpr auto g_materialPassNames = std::array{
```
jaremieromer: See comment on API MR about not needing these

source/ncengine/graphics2/diligent/pass/PostProcessPassBackend.cpp
jaremieromer: Yes - this is right. Each `...PassBackend` that we end up having will need to call this before the draw:

`BindRenderTarget(context, swapChain, perPassResourceSignature.GetPostProcessSinkBufferResource(), pass.colorRenderTargetIndex, pass.depthRenderTargetIndex);`

Where `BindRenderTarget` will either bind to the SwapChain or a render target taken from the `PostProcessSinkBufferResource`, which holds a vector of ITextures for Color and a vector of ITextures for Depth.

The backend chooses which RTs it will bind to based on `pass.colorRenderTargetIndex` and `pass.depthRenderTargetIndex`, so every pass, regardless of what pass type it is (Material, PP, etc) will also need to hold this information.

source/ncengine/graphics2/diligent/pass/PostProcessPassBackend.cpp
```diff
@@ -0,0 +1,129 @@
...
+            {
+                continue;
+            }
+
+            if (instance.buffer.has_value())
+            {
+                resource.SetVariable(pass.id, instance.buffer->GetBuffer());
+            }
+
+            // draw stuff
```
jaremieromer: The draw call for every PPPass will be identical:

```
            Diligent::DrawAttribs drawAttribs;
            drawAttribs.NumVertices = 4;
            drawAttribs.Flags = Diligent::DRAW_FLAG_VERIFY_ALL;
            context.Draw(drawAttribs);
```

source/ncengine/graphics2/diligent/resource/PostProcessBufferResource.h
```diff
@@ -0,0 +1,33 @@
...
+
+// 
+struct PostProcessDataVariable
+{
+    Diligent::IShaderResourceVariable* variable = nullptr;
+    PostProcessPass::type passId = PostProcessPass::None;
+};
+
+// 
+class PostProcessBufferResource
```
jaremieromer: We should name this `PostProcessPropertyBufferResource` to distinguish it from the resource that binds the render targets to the shader as an SRV, which I will change to be named `PostProcessSinkBufferResource`

jaremieromer (approved):
I think this is exactly what we need!

PR #823, opened 2024-11-30
Add UniformBuffer

McCallisterRomer:
Adding `UniformBuffer` type and updating `EnvironmentBufferResource` accordingly. Post process passes will optionally use it for their properties as well.

### DIFF
--- a/source/ncengine/graphics2/diligent/resource/base/UniformBuffer.h
+++ b/source/ncengine/graphics2/diligent/resource/base/UniformBuffer.h
@@ -1,0 +1,60 @@
+#pragma once
+
+#include "graphics2/ShaderTypes.h"
+
+#include "Common/interface/RefCntAutoPtr.hpp"
+#include "Graphics/GraphicsEngine/interface/RenderDevice.h"
+#include "Graphics/GraphicsEngine/interface/DeviceContext.h"
+
+#include <span>
+#include <string_view>
+
+namespace nc::graphics
+{
+class UniformBuffer
+{
+    public:
+        template<TriviallyCopyable T>
+        explicit UniformBuffer(Diligent::IDeviceContext& context,
+                               Diligent::IRenderDevice& device,
+                               const T& initialData,
+                               std::string_view name = "UniformBuffer")
+            : UniformBuffer{
+                context,
+                device,
+                static_cast<const void*>(&initialData),
+                sizeof(T),
+                name
+              }
+        {
+        }
+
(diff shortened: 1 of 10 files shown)


source/ncengine/graphics2/diligent/resource/base/UniformBuffer.h
```diff
@@ -0,0 +1,60 @@
...
+#include <span>
+#include <string_view>
+
+namespace nc::graphics
+{
+class UniformBuffer
+{
+    public:
+        template<TriviallyCopyable T>
+        explicit UniformBuffer(Diligent::IDeviceContext& context,
```
jaremieromer: We should probably name this `DynamicUniformBuffer`, or just `DynamicBuffer`. If we do that, then we can eventually make a buffer type called `ConstantUniformBuffer`, or just `ConstantBuffer`.  The difference being for constant buffers we  have to do the whole staging buffer/transition state thing every write. 

Once we finally have some breathing room we may want to test both types out for performance - as we may find ourselves out of dynamic pool space if we only have a dynamic type.

jaremieromer (approved):
It doesn't have to be done now, but we will want to implement a strategy to safeguard against overflowing physical device limits for dynamic uniform buffers. When I added an array of 50,000 uint64_t to EnvironmentBufferData, I received the following validation error, which will crash some users.


```
 Dynamic heap buffer, type = VK_OBJECT_TYPE_BUFFER; | MessageID = 0x8a540de7 | vkUpdateDescriptorSets(): pDescriptorWrites[0].pBufferInfo[0].range (400096) is greater than maxUniformBufferRange (65536) for descriptorType VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER_DYNAMIC. The Vulkan spec states: If descriptorType is VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER or VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER_DYNAMIC, the range member of each element of pBufferInfo, or the effective range if range is VK_WHOLE_SIZE, must be less than or equal to VkPhysicalDeviceLimits::maxUniformBufferRange (https://vulkan.lunarg.com/doc/view/1.3.275.0/windows/1.3-extensions/vkspec.html#VUID-VkWriteDescriptorSet-descriptorType-00332)
```

We probably can't realistically do anything clever like create a new buffer and bring the limit exceeders over to that one, but we should detect if sizeof (buffer data) >= the physical device limits for maxUniformBufferRange, and crash with an NcError so we know as devs not to exceed.

PR #824, opened 2024-11-30
PostProcess API

McCallisterRomer:
Adding API types and functions for post processing. I have a bit of cleanup + testing to do on the implementation still. Posting separately in case you have feedback and to cut down on scope of the next PR.

### DIFF
--- a/source/ncengine/graphics2/GraphicsUtility.cpp
+++ b/source/ncengine/graphics2/GraphicsUtility.cpp
@@ -10,20 +10,12 @@ namespace
 {
 constexpr auto g_materialPassNames = std::array{
     std::string_view{"Shadow"},
-    std::string_view{"Toon"},
-    std::string_view{"Alpha"},
-    std::string_view{"Depth"},
-    std::string_view{"Normals"},
-    std::string_view{"Outline"}
+    std::string_view{"Toon"}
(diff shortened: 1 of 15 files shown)


source/ncengine/graphics2/GraphicsUtility.cpp
```diff
@@ -1,10 +1,14 @@
...
+
 #include <array>
+#include <vector>
+#include <ranges>
 
 namespace
 {
-constexpr auto g_passNames = std::array{
+constexpr auto g_materialPassNames = std::array{
     std::string_view{"Shadow"},
```
jaremieromer, line 12: Should this only be "Shadow" and "Toon" now?

source/ncengine/graphics2/GraphicsUtility.cpp
```diff
@@ -13,7 +17,7 @@ constexpr auto g_passNames = std::array{
     std::string_view{"Outline"}
 };
 
-constexpr auto g_passFlags = std::array{
+constexpr auto g_materialPassFlags = std::array{
```
jaremieromer, line 16: Should this only be Shadow and Toon now?

jaremieromer (approved):

PR #826, opened 2024-12-03
Migrate WireframeRenderer

### LINKED ISSUE #814: Migrate WireframeRenderer

McCallisterRomer:
resolves #814 

Adding system + pass for `WireframeRenderer`.

### DIFF
--- a/source/ncengine/graphics2/NcGraphicsImpl2.cpp
+++ b/source/ncengine/graphics2/NcGraphicsImpl2.cpp
@@ -179,6 +179,12 @@ NcGraphicsImpl2::NcGraphicsImpl2(const config::GraphicsSettings& graphicsSetting
             m_engine.GetShaderFactory(),
             m_shaderBindings
           )},
+          m_wireframePass{
+            m_engine.GetDevice(),
+            m_engine.GetSwapChain(),
+            m_engine.GetShaderFactory(),
+            m_shaderBindings
+          },
           m_postProcessPassBackend{
             MakePostProcessPasses(
                 m_engine.GetContext(),
@@ -322,6 +328,7 @@ void NcGraphicsImpl2::Run()
     m_shaderBindings.GetMeshBuffer().SetBuffers(context);
 
     m_materialPassBackend.Render(context, renderState.meshRendererState.passBatches);
+    m_wireframePass.Render(context, renderState.wireframeRenderState);
     /** @todo Post process PSOs are currently null. Add this call in somewhere once implemented. */
     // m_postProcessPassBackend.Render(context, m_shaderBindings.GetPerFrameSignature().GetPostProcessPropertyBuffer());
     m_ui.Render(context);
--- a/source/ncengine/graphics2/diligent/pass/WireframePass.cpp
+++ b/source/ncengine/graphics2/diligent/pass/WireframePass.cpp
@@ -1,0 +1,172 @@
+#include "WireframePass.h"
+#include "graphics2/diligent/ShaderFactory.h"
+#include "graphics2/diligent/resource/MeshBuffer.h"
+#include "graphics2/diligent/resource/ShaderBindings.h"
+#include "graphics2/diligent/resource/WireframeBufferResource.h"
+
+#include "ncutility/NcError.h"
+
+namespace
+{
+constexpr auto g_pixelShader = std::string_view{
+R"(
+cbuffer EnvironmentBufferData
+{
+    float4x4 cameraViewProjection;
+    float3 cameraPosition;
+    uint dirLightsCount;
+    uint pointLightsCount;
+    uint spotLightsCount;
+    float2 padding;
+};
+
+cbuffer WireframeBufferData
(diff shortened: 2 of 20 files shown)


source/ncengine/graphics2/NcGraphicsImpl2.cpp
```diff
@@ -322,6 +328,7 @@ void NcGraphicsImpl2::Run()
     m_shaderBindings.GetMeshBuffer().SetBuffers(context);
 
     m_materialPassBackend.Render(context, renderState.meshRendererState.passBatches);
+    m_wireframePass.Render(context, renderState.wireframeRenderState);
```
McCallisterRomer, line 331: All of these probably belong in some sort of pass manager thing. I don't know if you've started anything like that, and it would probably just be more merge conflicts for you, so I've left it alone. Can do once you're current changes are in, if needed.

source/ncengine/graphics2/diligent/pass/WireframePass.cpp
```diff
@@ -0,0 +1,172 @@
...
+    if (state.wireframeData.empty())
+    {
+        return;
+    }
+
+    context.SetPipelineState(m_pso);
+    for (const auto& [data, mesh] : state.wireframeData)
+    {
+        m_buffer->Update(context, data);
+        const auto attribs = Diligent::DrawIndexedAttribs{
```
McCallisterRomer, line 159: Decided not to instance these because the common case right now is editor selections, which is just 0/1/2 instances. Might be worth revisiting if we add debug rendering for jolt or mass collision rendering.

source/ncengine/graphics2/frontend/subsystem/WireframeRendererSubsystem.cpp
```diff
@@ -0,0 +1,151 @@
+#include "WireframeRendererSubsystem.h"
```
McCallisterRomer, line 1: This whole file is almost copy paste from the existing implementation. Changed renderer type to `MeshRenderer2` and slightly tweaked a few `emplace_back` calls to match `WireframeInstanceState`.

McCallisterRomer: Oh yeah, we also previously disabled everything here if `NC_EDITOR_ENABLED` isn't on. I didn't do that, but maybe we want to? Its awkward because I don't think we really care about wireframes in prod builds, but the component still exists, so it isn't obvious that the component is conditionally ignored. Thoughts?

source/ncengine/graphics2/NcGraphicsImpl2.cpp
jaremieromer: I'll incorporate this pass backend into the master pass backend I'm working on

source/ncengine/graphics2/diligent/pass/WireframePass.cpp
```diff
@@ -0,0 +1,172 @@
...
+{
+    if (state.wireframeData.empty())
+    {
+        return;
+    }
+
+    context.SetPipelineState(m_pso);
+    for (const auto& [data, mesh] : state.wireframeData)
+    {
+        m_buffer->Update(context, data);
```
jaremieromer, line 158: I saw that you didn't want to instance to small count of elements, and that makes sense, but why not do bindless here? (use a structured buffer with a vector of `WireFrameBufferData`)

source/ncengine/graphics2/diligent/resource/WireframeBufferResource.cpp
```diff
@@ -0,0 +1,24 @@
+#include "WireframeBufferResource.h"
+
+namespace nc::graphics
+{
+WireframeBufferResource::WireframeBufferResource(Diligent::IDeviceContext& context,
+                                         Diligent::IRenderDevice& device,
```
jaremieromer: Column alignment here

source/ncengine/graphics2/frontend/subsystem/WireframeRendererSubsystem.cpp
jaremieromer: If possible, I think we should not have any editor-specific logic in the engine itself, and we should leave it up to the user whether or not to use the component.

jaremieromer (approved):

source/ncengine/graphics2/diligent/pass/WireframePass.cpp
McCallisterRomer: Basically because of unknowns.
- I have an issue to add debug rendering for physics, at which point we'll probably want to add instancing as well, but I don't know exactly what other changes will be necessary for that.
- I didn't know how we wanted to handle the max size of that buffer. It can't just be component count, because (at least with previous physics rendering) you would render every contact point with a single `WireframeRenderer` component. It probably needs to be a very large buffer for dev builds, but then do we just waste that space in prod builds?

PR #827, opened 2024-12-04
Templatize InstanceCache

McCallisterRomer:
Making `InstanceCache` a class template so we can have distinct instance types for static/skinned meshes. Also tweaked a few names ~~that would be too generic once `SkinnedRenderer` comes along~~ for `StaticMesh`/`SkinnedMesh` renaming.


source/ncengine/graphics2/ShaderTypes.h
```diff
@@ -32,12 +32,20 @@ struct OutlinePassData
 };
 
 // Object model for MeshRenderers (type: StructuredBuffer element type).
-struct InstanceData
+struct MeshRendererInstanceData
```
McCallisterRomer: Just a thought, but are we set on the names `MeshRenderer`/`SkinnedRenderer` as opposed to something like `StaticMesh`/`SkinnedMesh`? Not a big deal either way, but having 'renderer' on the component has always felt a tad weird to me. Double weird when you end up with names like `MeshRendererRenderState`.

source/ncengine/graphics2/NcGraphicsImpl2.cpp
```diff
@@ -321,7 +321,7 @@ void NcGraphicsImpl2::Run()
     m_shaderBindings.GetPerFrameSignature().Commit(context);
     m_shaderBindings.GetMeshBuffer().SetBuffers(context);
 
-    m_materialPassBackend.Render(context, renderState.meshRendererState.passBatches);
+    m_materialPassBackend.Render(context, renderState.meshRenderState.staticMeshBatches);
```
jaremieromer, line 330: much more expressive

source/ncengine/graphics2/frontend/subsystem/InstanceCache.h
```diff
@@ -41,38 +41,41 @@ struct StagedBatchRegion
     asset::MeshView mesh = asset::MeshView{};
 };
 
-// InstanceData for a MeshRenderer waiting to added to a batch.
+// Instance data for a mesh component waiting to added to a batch.
```
jaremieromer: *to be added to a batch

jaremieromer (approved):

PR #830, opened 2024-12-06
Add BoneCache

McCallisterRomer:
Reverted back to draft - might want to tweak some things.

Precursor to skeletal animation migration. We need a way to manage offsets/counts in the animated bone buffer. These have to be stable so the `InstanceCache` for `SkinnedMesh`es doesn't have to get thrashed every frame. Enter the `BoneCache`.

### DIFF
--- a/source/ncengine/graphics2/frontend/subsystem/BoneCache.h
+++ b/source/ncengine/graphics2/frontend/subsystem/BoneCache.h
@@ -1,0 +1,84 @@
+#pragma once
+
+#include "graphics2/ShaderTypes.h"
+
+#include "ncengine/utility/SparseMap.h"
+
+#include <span>
+#include <vector>
+
+namespace nc::graphics
+{
+// Identifier for a range of bones in the BoneCache.
+using BoneCacheHandle = uint32_t;
+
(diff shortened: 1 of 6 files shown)


source/ncengine/graphics2/frontend/subsystem/BoneCache.h
```diff
@@ -0,0 +1,84 @@
...
+        // Commit allocations to the buffer.
+        // IMPORTANT: Neither the staging area nor buffer may be accessed while this in progress.
+        void CommitPendingChanges();
+
+        // Build info specifying modified buffer ranges
+        auto BuildUpdateInfo() -> BufferUpdateInfo<BoneData>;
+
+    private:
+        std::vector<BoneData> m_data;
+        BoneCacheStaging m_staging;
```
McCallisterRomer, line 82: note to self: Synchronization and ownership is getting a bit difficult to track between mesh/animation stuff. It might be better to pull staging out of here and have `MeshSubsystem` own it (maybe call it something else), then `SkeletalAnimationSubsystem` can own `BoneCache`. Keeping them independent makes things clearer within the systems. Somebody (`GraphicsFrontend`?) will still have to to sync between them at some point.

PR #838, opened 2024-12-12
Post Process System

jaremieromer:
Note: I fixed the tests and made the change to the constants for `PassDesc` specification like we mentioned, along with some various cleanup.

Added `PassBackend` that coordinates execution of passes and their render targets.


sample/config.ini
```diff
@@ -51,7 +51,7 @@ target_fps=60
 use_shadows=1
 antialiasing=8
 initial_batch_size=1
-use_validation_layers=0
+use_validation_layers=1
```
McCallisterRomer: remove

source/ncengine/graphics2/NcGraphicsImpl2.cpp
```diff
@@ -318,24 +359,15 @@ void NcGraphicsImpl2::Run()
...
-
-    constexpr auto ClearColor = Vector4{0.050f, 0.050f, 0.050f, 1.0f};
-    context.ClearRenderTarget(pRTV, &ClearColor.x, Diligent::RESOURCE_STATE_TRANSITION_MODE_TRANSITION);
-    context.ClearDepthStencil(pDSV, Diligent::CLEAR_DEPTH_FLAG, 1.f, 0, Diligent::RESOURCE_STATE_TRANSITION_MODE_TRANSITION);
-
-    m_postProcessPassBackend.Update(context, renderState.postProcessState);
+    m_passBackend.Update(renderState.postProcessState);
     m_shaderBindings.Update(context, device, renderState);
     m_shaderBindings.GetPerFrameSignature().Commit(context);
+    m_shaderBindings.GetPerPassSignature().Commit(context);
```
McCallisterRomer, line 371: Just doing a quick pass now, but if its a 'per pass' resource, wouldn't we want to commit later?

source/ncengine/graphics2/diligent/ShaderFactory.cpp
```diff
@@ -24,6 +24,31 @@ auto ReadShaderFile(std::string_view filePath) -> std::vector<char>
     return buffer;
 }
 
+auto ShaderFactory::MakeShaderFromPath(std::span<const char> path,
```
McCallisterRomer: What's this guy for? We should have tests covering usage of `ReadShaderFile()` + `MakeShaderFromSource()` to make a shader from source file.

source/ncengine/graphics2/NcGraphicsImpl2.cpp
jaremieromer: No - commit is only needed once per frame and additionally once whenever the underlying resource handle (buffer, texture) on the GPU that the SRB is pointing to changes. (When the pointer changes, not when the stuff the pointer is pointing at changes.) We don't currently change the pointers to underlying resources outside of this initial set here.

The PerPass distinction is still required because we need to call `TransitionShaderResources` at SRB scope for each pass (it transitions the render targets from `RenderTargetView` to `ShaderResourceView`

source/ncengine/graphics2/NcGraphicsImpl2.cpp
```diff
@@ -347,5 +379,15 @@ void NcGraphicsImpl2::OnResize(const Vector2& dimensions, bool isMinimized)
 {
     (void)isMinimized;
     m_engine.GetSwapChain().Resize(static_cast<uint32_t>(dimensions.x), static_cast<uint32_t>(dimensions.y));
+    m_resizeNeeded = true;
+    m_dimensions = dimensions;
+}
+
+void NcGraphicsImpl2::Resize()
```
McCallisterRomer, line 392: I don't think this is being called.

source/ncengine/graphics2/diligent/ShaderFactory.cpp
jaremieromer: Missed that, will fix now

source/ncengine/graphics2/diligent/pass/PassManifest.cpp
```diff
@@ -0,0 +1,93 @@
...
+        throw nc::NcError("The pass was already registered");
+    }
+
+    switch (desc.type)
+    {
+        case PassType::Material:
+            m_materialPassDescs.emplace_back(std::move(desc));
+            break;
+        case PassType::Wireframe:
+            m_wireframePassDesc = desc;
```
McCallisterRomer: can move

McCallisterRomer (approved):

McCallisterRomer:
Awesome!

source/ncengine/graphics2/NcGraphicsImpl2.cpp
jaremieromer: Forgot to wire it back up. Fixed.

source/ncengine/graphics2/diligent/pass/PassManifest.cpp
jaremieromer: Done

Issue #802, opened 2024-11-16
Shader Cleanup


jaremieromer:
## Objectives

- Convert the shader for the Toon pass to GLSL
- Move the shader to the shaders folder
- Compile offline
- Conditionally include skinning w/ macros
- Hook compiler into CMake (?)

McCallisterRomer:
parts depend on #753

PR #843, opened 2024-12-15
Add SkinnedMesh Passes

### LINKED ISSUE #834: Add Animated Pass Variants

McCallisterRomer:
resolves #834 

Adding passes for rendering `SkinnedMesh`es. Made a duplicate shader for it - will handle macrofying it in #802 when we add offline compilation.

### DIFF
--- a/source/ncengine/graphics2/NcGraphicsImpl2.cpp
+++ b/source/ncengine/graphics2/NcGraphicsImpl2.cpp
@@ -194,6 +194,14 @@ NcGraphicsImpl2::NcGraphicsImpl2(const config::GraphicsSettings& graphicsSetting
                     .depthSink = MainDepth
                 },
                 PassDesc{
+                    .id = MaterialPassFlag::Toon,
+                    .name = "ToonSkinned",
+                    .type = PassType::SkinnedMaterial,
+                    .shaderPaths = ShaderPaths{"Toon.psh", "ToonSkinned.vsh"},
+                    .colorSink = MainColor,
+                    .depthSink = MainDepth
+                },
+                PassDesc{
                     .id = MaterialPassFlag::Normals,
                     .name = "Normals",
                     .type = PassType::Material,
@@ -201,11 +209,19 @@ NcGraphicsImpl2::NcGraphicsImpl2(const config::GraphicsSettings& graphicsSetting
                     .colorSink = NormalsColor
                 },
                 PassDesc{
+                    .id = MaterialPassFlag::Normals,
+                    .name = "NormalsSkinned",
+                    .type = PassType::SkinnedMaterial,
+                    .shaderPaths = ShaderPaths{"Normals.psh", "ToonSkinned.vsh"},
+                    .colorSink = NormalsColor
+                },
+                PassDesc{
                     .id = MiscPassFlag::Wireframe,
                     .name = "Wireframe",
                     .type = PassType::Wireframe,
                     .shaderPaths = ShaderPaths{"Wireframe.psh", "Wireframe.vsh"},
-                    .colorSink = MainColor
+                    .colorSink = MainColor,
+                    .depthSink = MainDepth
                 },
                 PassDesc{
                     .id = PostProcessPassFlag::Wave,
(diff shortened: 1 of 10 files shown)


source/ncengine/graphics2/NcGraphicsImpl2.cpp
```diff
@@ -193,19 +193,35 @@ NcGraphicsImpl2::NcGraphicsImpl2(const config::GraphicsSettings& graphicsSetting
...
+                    .colorSink = NormalsColor
+                },
                 PassDesc{
                     .id = MiscPassFlag::Wireframe,
                     .name = "Wireframe",
                     .type = PassType::Wireframe,
                     .shaderPaths = ShaderPaths{"Wireframe.psh", "Wireframe.vsh"},
-                    .colorSink = MainColor
+                    .colorSink = MainColor,
+                    .depthSink = MainDepth
```
McCallisterRomer, line 224: Will you double check this is legit? Wireframes were rendering on top of standard meshes, so I figured this was the ticket, but IDK.

source/ncengine/graphics2/diligent/ShaderFactory.h
```diff
@@ -46,7 +46,7 @@ class ShaderFactory
     private:
         Diligent::IRenderDevice* m_device;
         Diligent::RefCntAutoPtr<Diligent::IShaderSourceInputStreamFactory> m_streamFactory;
-        std::string_view m_shadersPath;
+        std::string m_shadersPath;
```
McCallisterRomer, line 49: As a `string_view` this pointed inside the global config object. Seemed like a UB timebomb if we ever change how that is stored.

source/ncengine/graphics2/diligent/pass/PassBackend.cpp
```diff
@@ -174,23 +182,39 @@ void PassBackend::Update(const PostProcessState& postProcessState)
...
-    for (auto [pass, batches] : std::views::zip(m_materialPasses, passBatches))
+    for (auto [staticPass, skinnedPass, staticBatches, skinnedBatches] : passView)
     {
+        // PassManifest verifies static/skinned pass pairs specify the same render targets, so we can just choose from either here.
         BindRenderTarget(context, swapChain,
                          perPassResourceSignature.GetPostProcessColorSinkBufferResource(),
                          perPassResourceSignature.GetPostProcessDepthSinkBufferResource(),
-                         pass.colorRTIndex, pass.depthRTIndex);
+                         staticPass.colorRTIndex, staticPass.depthRTIndex);
         ClearRenderTarget(context, swapChain,
```
McCallisterRomer, line 209: Quick Q: Are material passes pretty much guaranteed to always use different render targets? Was wondering if this clear could potentially wipe out a previous pass.

source/ncengine/graphics2/diligent/pass/PassTypes.h
```diff
@@ -28,11 +28,12 @@ struct ShaderPaths
 
 enum class PassType : uint8_t
 {
-    None,
-    Material,
-    Wireframe,
-    PostProcess,
-    UI
+    None            = 0,
```
McCallisterRomer, line 31: Made these flags so we can combine misc passes for filtering in the future, if needed. Wireframe registration was specifying != Material and != PostProccess. Easier in my mind to add an inclusion when we add a new misc pass, rather than add an exclusion if we add something else (like SkinnedMaterial here).

source/ncengine/graphics2/diligent/ShaderFactory.h
jaremieromer: GREAT CATCH!

source/ncengine/graphics2/diligent/pass/PassBackend.cpp
```diff
@@ -174,23 +182,39 @@ void PassBackend::Update(const PostProcessState& postProcessState)
...
 {
     NC_PROFILE_SCOPE("PassBackend::RenderMaterial()", ProfileCategory::Rendering);
-    NC_ASSERT(m_materialPasses.size() == passBatches.size(), "Frontend/Backend passes out of sync.");
+    NC_ASSERT(
+        m_staticMaterialPasses.size() == staticPassBatches.size() &&
+        m_skinnedMaterialPasses.size() == skinnedPassBatches.size(),
+        "Frontend/Backend passes out of sync."
+    );
+
+    auto passView = std::views::zip(
```
jaremieromer, line 195: 12/15/2024. A historic day. The emergence of the QuadZip™

source/ncengine/graphics2/diligent/pass/PassBackend.cpp
jaremieromer: This is still a layer of fragility I don't have a great solution for. You've found the fragility, though are backwards on why it's a problem:

Material passes were designed to always use the **_same_** render targets. I didn't know how to express this in the code, but any contiguous (the pass descs are contiguous in the manifest) passes that share the same sinks will be merged in the API into the same `vkRenderPass`. 

All passes considered a `MaterialPass` have the backends call the 'ClearRenderTarget' instruction because they get merged into a single pass, and this pass clears the `MainColor` render target before drawing, which effectively wipes the last frame's pixels.
 
The "Misc Passes" don't clear the render target at all, so they're intended to be 'additive' passes that overlay on top of   an existing pass.

The post process passes all clear their render targets, because they're expected to each have their own pass. 

I would love a sit down with you to talk about a better way to enforce this in code.

Here are the underlying behaviors/invariants:

- All passes considered a `MaterialPass` (currently skinned and static renderers) with the same Sink must be ordered contiguously, because the contiguous block of passes will be merged, and the merged pass will clear the Sink target. A material pass can be defined to output to a separate sink, it will clear that sink before rendering to it and is given its own `vkRenderPass`.
- All passes considered a 'MiscPass' do not clear their Sink target.
- All passes considered a 'PostProcessPass' clear their Sink target. They must have separate sinks, or they will overwrite the previous sink.

This is as robust as I could make it without a proper render graph: the hardcoded final pass always has a new render target, and it clears this before reading from (if any) the sources output by any previous pass (as a separate render target shader input), so if we have no material passes the post process passes all clear themselves so we don't get any compositing.

Thinking about this now - if we have no material passes but did have a Misc Pass, the Misc Passes _would_ continually add on to themselves so we'd get the output of the previous frames endlessly added on. I'll add an issue for Misc Passes to check and clear the target if there are no material passes.

Passes

jaremieromer (approved):

PR #847, opened 2024-12-17
Finish Skeletal Animation Migration

### LINKED ISSUE #832: Migrate Skeletal Animation System

McCallisterRomer:
resolves #832

The last (hopefully) of skeletal animation work: adding animation calculations and wiring up transition handling for animation controllers.

I had to rework the task graph to fix the issue with being briefly in an unanimated state. It works correctly on scene load and when adding/modifying meshes in game logic, but there are cases where it still happens in the editor due to when those operations run. I should be able to noodle out how to fix that, but I think its minor enough to not have to block PR any further.


jaremieromer (approved):
You did it!! I might have more feedback once I start working with it, but I really like the new organization

PR #848, opened 2024-12-17
Remove Legacy Graphics/Registry/Net

McCallisterRomer:
Deleting:
- `NcGraphics` 1 (2?) + Diligent feature flag
- Legacy `Registry`
- `NetworkDispatcher` and friends.

I moved the few important things from source/graphics to graphics2. Will rename the directory in a future PR. I also updated the smoke test scene to use new graphics types.

CI drivers are showing a few validation errors. Here, I've added the `NO_DYNAMIC_BUFFERS` flag for structured buffer resources and reduced the smoke test texture count to get it running on swiftshader, but there are a few more issues. Temporarily disabled validation error checking. To be enabled again with #850.


jaremieromer (approved):
NICE!!! After this we can rename all the graphics2 stuff to graphics, and change the 'diligent' folder to 'backend'

PR #854, opened 2024-12-21
Id-Based Asset Acquire Functions

McCallisterRomer:
- Adding functions to acquire asset views based on their id. I've hit an annoyance once or twice (and again with in-progress particles) where you have an asset id, and need to get the view. Without these you have to first get the path from the id. Then when you acquire the view, the path is converted back to the id. Silly.
- There's some mixed usage of `size_t`/`uint64_t` for ids. Adding an `AssetId` alias and updating everything I found.


source/ncengine/asset/AssetService.h
```diff
@@ -34,6 +34,7 @@ class IAssetService : public IAssetServiceBase
         virtual bool Unload(const InputType& input, asset_flags_type flags = AssetFlags::None) = 0;
         virtual void UnloadAll(asset_flags_type flags = AssetFlags::None) = 0;
         virtual auto Acquire(const InputType& input, asset_flags_type flags = AssetFlags::None) const -> data_type = 0;
+        virtual auto Acquire(AssetId, asset_flags_type  = AssetFlags::None) const -> data_type = 0;
```
jaremieromer: Extra space

jaremieromer (approved):

PR #855, opened 2024-12-21
MSAA

jaremieromer:
Added MSAA support.

We cap the count of samples requested by `antialising` in the config to whatever the device supports via a new struct and helper in `DiligentEngine`.

All the passes that aren't post process are multisampled. The post process render call then resolves all of these multisampled render targets down to single sampled targets before using the new single sampled targets for rendering.

The single sampled and multisampled targets share the same indices which makes it much cleaner when creating `PassDesc` in the `PassManifest` interface.
Depth targets don't need to be resolved. (I don't know why). 

Tested with an `antialising` value of 8, 1, 16, and 912 (in the >8 cases, we cap at 8).

### DIFF
--- a/source/ncengine/graphics2/diligent/pass/PassManifest.cpp
+++ b/source/ncengine/graphics2/diligent/pass/PassManifest.cpp
@@ -82,32 +82,20 @@ void PassManifest::RegisterPass(PassDesc desc)
         throw nc::NcError("The pass was already registered");
     }
 
-    if (desc.colorSink != NoTarget && desc.colorSink != SwapChainColorRTIndex)
-        m_colorSinkCount = std::max(m_colorSinkCount, desc.colorSink + 1);
-    if (desc.depthSink != NoTarget && desc.depthSink != SwapChainDepthRTIndex)
-        m_depthSinkCount = std::max(m_depthSinkCount, desc.depthSink + 1);
+    SetMaxIndices(desc.colorSink, desc.depthSink, false);
 
     switch (desc.type)
     {
         case PassType::Material:
-            if (desc.colorSink != NoTarget && desc.colorSink != SwapChainColorRTIndex)
-                m_colorSinkCountMsaa = std::max(m_colorSinkCountMsaa, desc.colorSink + 1);
-            if (desc.depthSink != NoTarget && desc.depthSink != SwapChainDepthRTIndex)
-                m_depthSinkCountMsaa = std::max(m_depthSinkCountMsaa, desc.depthSink + 1);
+            SetMaxIndices(desc.colorSink, desc.depthSink, true);
             m_staticMaterialPassDescs.emplace_back(std::move(desc));
             break;
         case PassType::SkinnedMaterial:
-            if (desc.colorSink != NoTarget && desc.colorSink != SwapChainColorRTIndex)
-                m_colorSinkCountMsaa = std::max(m_colorSinkCountMsaa, desc.colorSink + 1);
-            if (desc.depthSink != NoTarget && desc.depthSink != SwapChainDepthRTIndex)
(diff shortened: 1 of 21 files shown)


source/ncengine/graphics2/diligent/pass/PassManifest.cpp
```diff
@@ -80,15 +82,32 @@ void PassManifest::RegisterPass(PassDesc desc)
         throw nc::NcError("The pass was already registered");
     }
 
+    if (desc.colorSink != NoTarget && desc.colorSink != SwapChainColorRTIndex)
```
McCallisterRomer: All the ifs here do the same checks. I think it'd be a lot less noisy if they were stored in local bools and/or the count incrementing in the switch cases was extracted to a function.

source/ncengine/graphics2/diligent/resource/PostProcessColorSinkBufferResource.cpp
```diff
@@ -31,14 +32,26 @@ auto PostProcessColorSinkBufferResource::Add(Diligent::IRenderDevice& device,
         return addedIndices;
     }
 
-    if (numColorRenderTargets + m_colorRenderTargets.size() > m_maxTextures)
+    if (numSamples > 1)
```
McCallisterRomer, line 35: This function does this check a few times and basically has duplicated code each time, just targeting different vectors. Maybe something like this with a function/lamba to create it based on sample count would cut things down a bit:
```cpp
struct RTStuf
{
    std::vector<Diligent::RefCntAutoPtr<Diligent::ITexture>& x;
    std::vector<Diligent::IDeviceObject*>& y;
    std::vector<Diligent::IDeviceObject*>& z;
};
```

source/ncengine/graphics2/diligent/resource/PostProcessColorSinkBufferResource.cpp
```diff
@@ -64,23 +78,46 @@ auto PostProcessColorSinkBufferResource::Add(Diligent::IRenderDevice& device,
...
     }
 
-    SetArrayRegion(m_colorRenderTargetViewsSR, 0u, m_colorRenderTargetViewsSR.size());
+    if (numSamples > 1)
+    {
+        SetArrayRegion(m_colorRenderTargetViewsSRMsaa, m_colorRenderTargetViewsSR.size(), m_colorRenderTargetViewsSRMsaa.size());
+    }
+    else
+    {
+        SetArrayRegion(m_colorRenderTargetViewsSR, 0u, m_colorRenderTargetViewsSR.size());
```
McCallisterRomer: Didn't look all that closely, but is this end correct? Can this overwrite the first msaa entry?

source/ncengine/graphics2/diligent/resource/PostProcessDepthSinkBufferResource.cpp
```diff
@@ -27,9 +28,26 @@ auto PostProcessDepthSinkBufferResource::Add(Diligent::IRenderDevice& device,
         throw NcError{"Max texture count exceeded"};
     }
 
-    m_depthRenderTargets.reserve(m_depthRenderTargets.size() + numDepthRenderTargets);
-    m_depthRenderTargetViewsRT.reserve(m_depthRenderTargetViewsRT.size() + numDepthRenderTargets);
-    m_depthRenderTargetViewsSR.reserve(m_depthRenderTargetViewsSR.size() + numDepthRenderTargets);
+    if (numSamples > 1)
```
McCallisterRomer, line 30: Maybe the same idea as suggested in color. Not needed now, but at a glance, I think almost all of this code is shared with the color variety. Probably a good candidate for extracting a base class.

McCallisterRomer (approved):

PR #861, opened 2024-12-28
Outline Pass

jaremieromer:
Added an outline pass. I still need to come up with a screen-space anti-aliasing solution to get rid of the jaggies. 
I removed `outlineWidth` and `outlineColor` from `MaterialProperties` and replaced it with `normalIntensity` so we can control the normal artifacts at a per-material instance level.

image

image


sample/source/scenes/GraphicsTest.cpp
```diff
@@ -154,9 +154,14 @@ void GraphicsTest::Load(ecs::Ecs world, ModuleProvider modules)
...
         .tag = "Main Camera"
     });
+
     auto& camera = world.Emplace<SceneNavigationCamera>(cameraHandle);
     world.Emplace<FrameLogic>(cameraHandle, InvokeFreeComponent<SceneNavigationCamera>{});
     modules.Get<NcGraphics>()->SetCamera(&camera);
+
+    auto ncGraphics = modules.Get<NcGraphics>();
+    ncGraphics->SetPostProcessEffectEnabled(nc::OutlinedToonEffectId, true);
+    ncGraphics->SetPostProcessEffectProperties(nc::OutlinedToonEffectId, PostProcessPassFlag::Outline, PostProcessPassProperties{OutlinePassProperties{.width = 1.0f}});
```
McCallisterRomer: Very minor, but you can just pass `OutlinePassProperties` here and leave out the `PostProcessPassProperties`. Also, namespace is unnecessary here.

McCallisterRomer: 160 also uses `NcGraphics`, so maybe move line 162 above that.

source/ncengine/graphics2/GraphicsUtility.cpp
```diff
@@ -12,14 +12,14 @@ constexpr auto g_materialPassNames = std::array{
...
-    std::string_view{"Alpha"}
+    std::string_view{"Depth"}
 };
 
 constexpr auto g_materialPassFlags = std::array{
     nc::MaterialPassFlag::Shadow,
     nc::MaterialPassFlag::Toon,
     nc::MaterialPassFlag::Normals,
-    nc::MaterialPassFlag::Alpha,
+    nc::MaterialPassFlag::Depth,
```
McCallisterRomer, line 20: The order of these determines the ordering in which everything is processed (pass order, batch order, etc). The order is intended to be by ascending flag value. If this is correct, then the value of `MaterialPassFlags` needs to be adjusted so that depth is the highest value.

source/ncengine/graphics2/diligent/pass/PassBackend.cpp
```diff
@@ -230,8 +230,8 @@ void PassBackend::RenderMaterial(Diligent::IDeviceContext& context,
     for (auto [staticPass, skinnedPass, staticBatches, skinnedBatches] : passView)
     {
         // PassManifest verifies static/skinned pass pairs specify the same render targets, so we can just choose from either here.
-        BindRenderTarget(context, swapChain, colorSinkBuffer, depthSinkBuffer, staticPass.colorRTIndex, staticPass.depthRTIndex, m_numSamples > 1);
-        ClearRenderTarget(context, swapChain, colorSinkBuffer, depthSinkBuffer, staticPass.colorRTIndex, staticPass.depthRTIndex, m_numSamples > 1);
+        BindRenderTarget(context, swapChain, colorSinkBuffer, depthSinkBuffer, staticPass.colorRTIndex, staticPass.depthRTIndex, staticPass.isMsaa && m_numSamples > 1);
```
McCallisterRomer, line 233: Not suggesting a change now, but I'd like to fix this in the near future. Identifying a resource based on the intersection of three variables is beyond what I, as a mere mortal, can reason about.

source/ncengine/graphics2/diligent/pass/PassUtilities.cpp
```diff
@@ -132,6 +132,27 @@ void BindRenderTarget(Diligent::IDeviceContext& context,
     context.SetRenderTargets(colorRenderTargetIndex == NoTarget ? 0 : 1, &pRTV, pDSV, Diligent::RESOURCE_STATE_TRANSITION_MODE::RESOURCE_STATE_TRANSITION_MODE_TRANSITION);
 }
 
+auto CreateShaderFromSourceIfInitialized(ShaderFactory& shaderFactory, Diligent::SHADER_TYPE shaderType, const nc::graphics::ShaderPaths& shaderPaths) -> Diligent::RefCntAutoPtr<Diligent::IShader>
```
McCallisterRomer, line 135: I don't love that this helper function also handles the empty checking/null construction. To me, its better to express that directly in the PSO construction, if nullness is valid for that pipeline. I can live with it, though.

source/ncengine/graphics2/diligent/pass/PassUtilities.cpp
```diff
@@ -132,6 +132,27 @@ void BindRenderTarget(Diligent::IDeviceContext& context,
...
+{
+    Diligent::RefCntAutoPtr<Diligent::IShader> shader;
+    const auto& shaderPath = shaderType == Diligent::SHADER_TYPE_VERTEX ? shaderPaths.vertexShaderPath : shaderPaths.pixelShaderPath;
+
+    if (shaderPath == "Uninitialized")
```
McCallisterRomer: Either way, "Uninitialized" is a surprising 'empty' state to me. Can we default initialize to an empty string instead?

source/ncengine/graphics2/frontend/subsystem/CameraSubsystem.cpp
```diff
@@ -43,13 +43,22 @@ auto CameraSubsystem::BuildState(ecs::ExplicitEcs<Transform> ecs) -> CameraRende
                 m_mainCamera->ViewMatrix(),
                 m_mainCamera->ProjectionMatrix()
             ),
-            .position = transform.Position()
+            .invProjection = m_mainCamera->InverseProjectionMatrix(),
+            .position = transform.Position(),
+            .nearClip = g_defaultProperties.nearClip,
```
McCallisterRomer: The camera has configurable near/far clip values that are probably needed here. I don't think there's a getter yet.

source/ncengine/ui/editor/impl/ComponentWidgets.cpp
```diff
@@ -108,8 +108,7 @@ auto MaterialColorWidget(nc::MaterialProperties& properties) -> bool
 
 auto MaterialOutlineWidget(nc::MaterialProperties& properties) -> bool
 {
-    auto modified = nc::ui::InputColor3(properties.outlineColor, "color");
-    modified = nc::ui::DragFloat(properties.outlineWidth, "width", 0.1f, 0.0f, 10.0f) || modified;
+    auto modified = nc::ui::DragFloat(properties.normalIntensity, "normal intensity", 0.01f, 0.0f, 5.0f);
```
McCallisterRomer: lowerCamelCase is conventional for widget names.

source/ncengine/ui/editor/impl/windows/dialogs/PostProcessDialog.cpp
```diff
@@ -23,7 +23,9 @@ void DrawPostProcessPassInfo(nc::PostProcessEffectId effectId,
                     IMGUI_SCOPE(nc::ui::Indent);
                     auto copy = unpacked;
                     auto modified = nc::ui::InputColor3(copy.color, "color");
-                    modified = nc::ui::DragFloat(copy.width, "width", 0.05f, 0.1f, 10.0f) || modified;
+                    modified = nc::ui::DragFloat(copy.width, "width", 1.0f, 0.0f, 15.0f) || modified;
+                    modified = nc::ui::DragFloat(copy.depthThreshold, "depth threshold", 0.01f, 0.0f, 2.0f) || modified;
```
McCallisterRomer: same

test/ncengine/graphics2/CMakeLists.txt
```diff
@@ -224,6 +224,7 @@ add_executable(PassManifest_tests
 PassManifest_tests.cpp
     ${NC_SOURCE_DIR}/graphics2/GraphicsUtility.cpp
     ${NC_SOURCE_DIR}/graphics2/PostProcess.cpp
+    ${NC_SOURCE_DIR}/graphics2/diligent/ShaderFactory.cpp
```
McCallisterRomer: the diligent directory should really only appear in the integration tests. Can we stub whatever was required here?

source/ncengine/graphics2/GraphicsUtility.cpp
jaremieromer: Depth should come after shadow - fixed. (Looks like I'd fixed this as part of an experiment, then stashed it and forgot to reapply just this

source/ncengine/graphics2/diligent/pass/PassBackend.cpp
jaremieromer: Yar, it be gettin' a bit wild.

source/ncengine/graphics2/diligent/pass/PassUtilities.cpp
jaremieromer: That's really the only thing it does - Nullness is valid for the pipeline. I think the right move is to consolidate all the pipeline creation state into a method as we have a lot of duplicate code everywhere. I'll probably do that after we move to deferred because much will change

source/ncengine/graphics2/diligent/pass/PassUtilities.cpp
jaremieromer: Yes - you're right that's much more sane

source/ncengine/graphics2/frontend/subsystem/CameraSubsystem.cpp
jaremieromer: Done

test/ncengine/graphics2/CMakeLists.txt
jaremieromer: It relied on too many symbols from Diligent, so I've moved it to integration_tests

source/ncengine/graphics2/diligent/pass/PassBackend.cpp
jaremieromer: I'm going to leave it for now but with the intention to revisit when we move to deferred

include/ncengine/graphics/Camera.h
```diff
@@ -60,6 +60,12 @@ class Camera : public FreeComponent
         /** @brief Unproject 2D normalized device coordinates into 3D points on the near and far planes. */
         auto CastToNearAndFarPlanes(const Vector2& normalizedDeviceCoords) const -> NearFarPoints;
 
+        /** @brief Get the camera properties. */
+        auto GetProperties() const -> const CameraProperties& { return m_properties; };
+
+        /** @brief Set the camera properties */
+        void SetProperties(CameraProperties properties) { m_properties = std::move(properties); };
```
McCallisterRomer: I'm pretty sure the projection matrix is rebuilt only on resize, not each frame, so this doesn't quite work. If you don't need it right now, its probably easiest to just remove this.

McCallisterRomer (approved):

include/ncengine/graphics/Camera.h
McCallisterRomer: removed

PR #862, opened 2024-12-29
FXAA

jaremieromer:
Implemented FXAA using https://blog.simonrodriguez.fr/articles/2016/07/implementing_fxaa.html.

We still see some jaggies but the effect is GREATLY improved.


resources/shaders/PPFxaa.psh
```diff
@@ -0,0 +1,256 @@
+// Much credit to Timothy Lottes and Simon Rodriguez https://blog.simonrodriguez.fr/articles/2016/07/implementing_fxaa.html
+
+#ifdef VULKAN
```
McCallisterRomer: btw, you don't need to be doing this since we're 100% vulkan now.

McCallisterRomer (approved):

PR #867, opened 2025-01-08
Add Vehicle

McCallisterRomer:
Adding a vehicle constraint option for `RigidBody`. I added a character dropdown in the `PhysicsTest` scene so you can try it out without having to go through the hassle of building a vehicle, if you want.


include/ncengine/physics/RigidBody.h
```diff
@@ -182,6 +186,8 @@ class RigidBody
         auto ScalesWithTransform() const -> bool { return !IgnoreTransformScaling(); }
         auto IgnoreTransformScaling() const -> bool { return m_info.flags & RigidBodyFlags::IgnoreTransformScaling; }
         void IgnoreTransformScaling(bool value);
+        auto DisableSleeping() const -> bool { return m_info.flags & RigidBodyFlags::DisableSleeping; }
```
jaremieromer, line 189: We keep weaving in and out of how we should name our getters and setters. I am back on the "GetDisableSleeping" "SetDisableSleeping" train. (No need to change, just think we should converge and then stick there)

include/ncengine/physics/Vehicle.h
```diff
@@ -0,0 +1,230 @@
...
+        }
+
+        /**
+         * @name Vehicle Input
+         * @note Vehicle input does not wake the owning RigidBody. The RigidBody can be awakened when needed, or
+         *       RigidBodyFlags::DisableSleeping may be set.
+         * 
+         * Throttle values range from [-1, 1] indicating reverse/forward direction and throttle amount.
+         * Steering values range from [-1, 1] indicating left/right steering angle and amount.
+         * Brake values range from [0, 1] indicating brake amount.
```
jaremieromer, line 171: This is a great interface

source/ncengine/physics/jolt/VehicleManager.cpp
```diff
@@ -0,0 +1,155 @@
...
+        m_handles[i] = std::move(vehicle);
+        return i;
+    }();
+
+    const auto ownerId = owner.Index();
+    auto& state = m_entityToVehicle.contains(ownerId)
+        ? m_entityToVehicle.at(ownerId)
+        : m_entityToVehicle.emplace(ownerId, {});
+
+    state = std::make_unique<Vehicle>(std::move(createInfo), vehiclePtr, index);
```
jaremieromer, line 73: I see your comment about vehicle being moved and the std::move - how is this pointer valid at this point? Can you explain?

test/ncengine/physics/Vehicle_unit_tests.cpp
```diff
@@ -0,0 +1,183 @@
...
+                DestroyBody(body);
+            }
+        }
+
+    public:
+        nc::physics::ConstraintFactory constraintFactory;
+        std::vector<JPH::Body*> bodies;
+        std::vector<JPH::Ref<JPH::VehicleConstraint>> constraints;
+
+        auto CreateVehicle(nc::VehicleInfo&& info) -> nc::Vehicle
```
jaremieromer, line 32: Why && here? I'm sure you did it for a reason but I can't puzzle it out

jaremieromer (approved):
An unbelievable number of parameters. Amazing job!
I love the little car. The suspension makes me think of a June Bug when it jumps.

source/ncengine/physics/jolt/VehicleManager.cpp
McCallisterRomer: Its a ref counted pointer, so its just like:
```cpp
auto smart = std::make_unique<Foo>();
auto raw = smart.get();
Consume(std::move(smart));
raw->Use(); // ok
smart->Use(); // nullptr dereference
```
where `move` changes the owner, but the address on the heap is the same.

test/ncengine/physics/Vehicle_unit_tests.cpp
McCallisterRomer: Eh, it doesn't have to be here for the test, but I was probably just trying to be consistent with `Vehicle`'s constructor. I made that one take by `VehicleInfo&&` because its moderately expensive to copy. Recommendation at [F.18](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-consume) of core guidelines. We sort of treat the 'exception' to that rule as the default because it makes for a more friendly interface.

PR #871, opened 2025-01-11
Update Hull/Concave Assets

### LINKED ISSUE #693: Support for new convex mesh format
- Update nca format to hold jolt's cooked data
- On load, create shape object mapped to asset id
  - Possible option: Add signal to asset manager that sends the raw bytes to `ShapeFactory` to do the conversion/storage.
  - Update `ConvexHullView` accordingly
- Add support for convex shapes to `RigidBody` and the editor.
- Update nc-convert version (no reason to continue v4 support, since legacy physics is gone)
- Can we add back inspect option to nc-convert?
- ? normalize naming to `ConvexHull` - `AssetType::HullCollider` and nc-convert `hull-collider` options are still around.

### LINKED ISSUE #694: Support for new concave mesh format

McCallisterRomer:
resolves #693, resolves #694

~~Updating `HullCollider` (renamed `ConvexHull`) for Jolt. The asset now just holds binary data in the form Jolt wants. This will have to get loaded into a shape object somewhere in `NcPhysics`. This change updates most of `nc-convert`/`NcAsset` and the asset manager, but there's no way to use the asset yet. I'll do `RigidBody`/`NcPhysics`/`ConvexHullView` updates in another PR (or 2).~~


source/ncconvert/converters/GeometryConverter.cpp
```diff
@@ -430,10 +436,11 @@ class GeometryConverter::impl
                 LOG("Warning: Bad values detected in mesh. {} values have been set to 0.", count);
             }
 
-            return asset::HullCollider{
+            const auto shape = jolt::BuildConvexHull(convertedVertices);
```
jaremieromer, line 441: I appreciate changing the naming to match Jolt

source/ncjolt/ncjolt/ShapeUtility.h
```diff
@@ -0,0 +1,32 @@
...
+class Shape;
+} // namespace JPH
+
+namespace nc
+{
+struct Vector3;
+
+namespace jolt
+{
+/** @note Functions in this file require an instance of JoltApi to be initialized somewhere in the executable. */
```
jaremieromer, line 22: Curious what the behavior is if there isn't an instance - and is it possible for there not to be? 

Also, can you explain more about why you put this comment - aka: what's different about these functions and all the other code you've written referencing jolt that made this comment required?

jaremieromer (approved):

source/ncjolt/ncjolt/ShapeUtility.h
McCallisterRomer: It would result in a crash/assert from within jolt. The main point of adding the ncjolt library was to add `JoltApi`, which does all of the globabl/static jolt initialization. `NcEngine` and `nc-convert` will handle initializing that guy, but that doesn't mean the instance exists for the entirety of the application.

I can imagine a future refactor to `nc-convert` that changes how the builder/converters are constructed which lands in a case where we call one of these functions prior to `JoltApi` being constructed - maybe asset inspection is pulled out of the builder (which makes the jolt instance) and can run without needing to construct the converters. Inspection of some assets still requires jolt, so we'd crash. Also tests.

Other code referencing jolt is 'private' within the physics module, so initialization is guaranteed to happen prior to that code being called. Comments are mostly just a reminder in case we run into confusing crashes in the future.

PR #872, opened 2025-01-12
finish convex hull support

McCallisterRomer:
part of #693 

Finish support for new `ConvexHull` format:
  - Added support/tracking for `JPH::ConvexHullShape` to `ShapeFactory`.
    - Asset loads are now routed from the manager to the factory via `Signal`. Only the ids need to be tracked in the manager.
  - Updated `Shape` and `RigidBody` ui/serialization to support `ConvexHull`s.
  - Added back support for inspecting hulls in `nc-convert`
  - Added some usage to `PhysicsTest`/`Benchmarks`/`SmokeTest` scenes

Found a few bugs along the way:
  - `Shape`s in `SmokeTest` were all defaulting to `Box`.
  - Changed `StringMap` to use our `Fnv1a` instead of `std::hash`. Seems that msvc uses fnv1a for strings, but not gcc. I'd rather be explicit about which algorithm is being used.
  - Removed an assert in `ComponentPool` that silences relevant diagnostics if triggered while an exception is in flight. I've never seen it catch anything important.

PR into feature branch, [manual build](https://github.com/NcStudios/NcEngine/actions/runs/12735721198)


sample/source/scenes/SmokeTest.cpp
```diff
@@ -207,26 +207,29 @@ void SmokeTest::Load(ecs::Ecs world, ModuleProvider modules)
...
+    const auto capsule2 = world.Emplace<Entity>({.position = Vector3{ 3.5f, 3.0f, 0.0f}});
+    const auto hull     = world.Emplace<Entity>({.position = Vector3{ 1.0f, 1.0f, 1.0f}});
+
+    world.Emplace<StaticMesh>(box1,     mesh::Cube,    material::Default);
+    world.Emplace<StaticMesh>(box2,     mesh::Cube,    material::Default);
+    world.Emplace<StaticMesh>(sphere1,  mesh::Sphere,  material::Default);
+    world.Emplace<StaticMesh>(sphere2,  mesh::Sphere,  material::Default);
     world.Emplace<StaticMesh>(capsule1, mesh::Capsule, material::Default);
     world.Emplace<StaticMesh>(capsule2, mesh::Capsule, material::Default);
+    world.Emplace<StaticMesh>(hull, mesh::Ramp, material::Default);
```
jaremieromer: Fix this lone hull to be column-aligned like the others

source/ncengine/physics/jolt/ShapeFactory.cpp
```diff
@@ -0,0 +1,115 @@
...
+#include "Conversion.h"
+
+#include "ncengine/asset/AssetData.h"
+#include "ncjolt/ShapeUtility.h"
+
+#include "Jolt/Physics/Collision/Shape/BoxShape.h"
+#include "Jolt/Physics/Collision/Shape/CapsuleShape.h"
+#include "Jolt/Physics/Collision/Shape/ScaledShape.h"
+#include "Jolt/Physics/Collision/Shape/RotatedTranslatedShape.h"
+#include "Jolt/Physics/Collision/Shape/ScaledShape.h"
```
jaremieromer, line 10: These headers could do with a bit of a tidy. ScaledShape is duped and they could use some alphabetizing.

source/ncengine/serialize/ComponentSerialization.cpp
```diff
@@ -179,9 +179,24 @@ void SerializeRigidBody(std::ostream& stream, const RigidBody& out, const Serial
...
+    {
+        case ShapeType::Box:
+        case ShapeType::Sphere:
+        case ShapeType::Capsule:
+            serialize::Serialize(stream, shape.GetLocalScale());
+            serialize::Serialize(stream, shape.GetLocalPosition());
+            break;
+        case ShapeType::ConvexHull:
+            serialize::Serialize(stream, shape.GetAssetId());
+            serialize::Serialize(stream, shape.GetLocalScale());
```
jaremieromer, line 195: Why don't we care about position for hulls?

jaremieromer (approved):

source/ncengine/serialize/ComponentSerialization.cpp
McCallisterRomer: If we really need it, it can be added in the future. I didn't do it now because translation/rotation/scaling of shapes in jolt is done via various `DecoratedShape` subclasses (you construct a base instance of the actual shape, from that you make a scaled shape instance, ...). For primitives, we can use scale directly in the construction of the base instance, resulting in only one 'wrapper' shape for translation. For hull/mesh shapes, we cannot, so we get sort of a nesting doll situation. Also, I figured that the main use of hull/mesh shapes is to match/approximate a rendered mesh. In those cases, position is handled by the local transform values.

PR #878, opened 2025-01-18
Add PhysicsSnapshot

### LINKED ISSUE #876: Add Physics Snapshots/Rollback

McCallisterRomer:
resolves #876

Adding rollback support for `NcPhysics`
 - Adding `PhysicsTick` and `PhysicsSnapshot` for state tracking.
 - Added functions to `NcPhysics` to facilitate manual updates/ticks.
   - Rustbelt has its own rollback strategy outside of what the engine really ought to care about at this point. With new functions, we can wrap `NcPhysics` in a custom physics module, giving us control over when/how many physics ticks happen.
   - Original functionality is still the default, with an opt-in in `PhysicsSettings`.


test/ncengine/physics/jolt/PhysicsSnapshot_unit_tests.cpp
```diff
@@ -0,0 +1,140 @@
...
+#include "ncengine/physics/PhysicsSnapshot.h"
+#include "ncutility/ScopeExit.h"
+
+#include <ranges>
+
+class PhysicsSnapshotTests : public JoltApiFixture
+{
+    protected:
+
+        // nc::Signal<const nc::asset::ConvexHullUpdateEventData&> convexHullSignal;
```
jaremieromer: What's going on here? Is this meant to be deleted, or meant as an example?

jaremieromer (approved):
Conditionally approving based on answer to this question:
The app crashes if enable_network_rollback is 1. I'm assuming this is because of the single player bug you're fixing now. If that's the case, I'm good with pushing in this state and waiting on the single player fix.

test/ncengine/physics/jolt/PhysicsSnapshot_unit_tests.cpp
McCallisterRomer: Oops, copy paste then realized only the base class was necessary.

PR #884, opened 2025-01-31
Allow Transform Interpolation in Physics

McCallisterRomer:
Rustbelt needs to hard reset `RigidBody` positions, so we want a way to smooth things out a bit. Adding an alternative code path in `NcPhysics` to interpolate when syncing transforms and animating wheels. This is never used in the default setup; its opt-in for the manual tick mode.


source/ncengine/physics/NcPhysicsImpl.cpp
```diff
@@ -52,6 +53,49 @@ class NcPhysicsStub : public nc::NcPhysics
...
+                 Func&& synchronize)
+{
+    for (auto& body : rigidBodyPool)
+    {
+        if (body.GetBodyType() == nc::BodyType::Static)
+        {
+            continue;
+        }
+
+        auto* apiBody = reinterpret_cast<JPH::Body*>(body.GetHandle());
```
jaremieromer, line 69: Can apiBody ever be null?

jaremieromer (approved):

source/ncengine/physics/NcPhysicsImpl.cpp
McCallisterRomer: No, `BodyManager` sets/clears the handle via `OnAdd`/`OnRemove` callbacks. The only time it can be null is in the constructor of `RigidBody`.

PR #886, opened 2025-02-05
Validation Error Fixes

### LINKED ISSUE #850: Address Validation Errors in CI
Fix issues and add back validation layer assertion in run_smoke_test.sh.

Excerpts:
```
VUID-vkCmdDrawIndexed-None-08114(ERROR / SPEC): msgNum: 1064294454 - Validation Error: [ VUID-vkCmdDrawIndexed-None-08114 ] Object 0: handle = 0x230000000023, name = PerFrameResourceSignature - static/mutable set, type = VK_OBJECT_TYPE_DESCRIPTOR_SET; | MessageID = 0x3f6fd836 | Descriptor set VkDescriptorSet 0x230000000023[PerFrameResourceSignature - static/mutable set] encountered the following validation error at vkCmdDrawIndexed time: Descriptor in binding #11 index 17 is being used in draw but has never been updated via vkUpdateDescriptorSets() or a similar call. The Vulkan spec states: Descriptors in each bound descriptor set, specified via vkCmdBindDescriptorSets, must be valid if they are statically used by the VkPipeline bound to the pipeline bind point used by this command and the bound VkPipeline was not created with VK_PIPELINE_CREATE_DESCRIPTOR_BUFFER_BIT_EXT (https://vulkan.lunarg.com/doc/view/1.3.261.1/linux/1.3-extensions/vkspec.html#VUID-vkCmdDrawIndexed-None-08114)
    Objects: 1
        [0] 0x230000000023, type: 23, name: PerFrameResourceSignature - static/mutable set
```

```
VUID-VkDescriptorImageInfo-imageLayout-00344(ERROR / SPEC): msgNum: -564812795 - Validation Error: [ VUID-VkDescriptorImageInfo-imageLayout-00344 ] Object 0: handle = 0x55d7250926d0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0x5d000000005d, name = Color Render Target: 0, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0xde55a405 | vkCmdDraw: Cannot use VkImage 0x5d000000005d[Color Render Target: 0] (layer=0 mip=0) with specific layout VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL that doesn't match the previous known layout VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL. The Vulkan spec states: imageLayout must match the actual VkImageLayout of each subresource accessible from imageView at the time this descriptor is accessed as defined by the image layout matching rules (https://vulkan.lunarg.com/doc/view/1.3.261.1/linux/1.3-extensions/vkspec.html#VUID-VkDescriptorImageInfo-imageLayout-00344)
    Objects: 2
        [0] 0x55d7250926d0, type: 6, name: NULL
        [1] 0x5d000000005d, type: 10, name: Color Render Target: 0
```

jaremieromer:
closes #850 

### Outline: 
Fix the validation errors introduced with the post process work.
There were two primary validation errors emitted every draw call:
1. The texture array resources had uninitialized image views bound to the pipeline.
2. The texture array resources had image views that were in different states within a single descriptor (SHADER_RESOURCE vs RENDER_TARGET)

### Details:
To fix 1, a 1x1 pixel texture was default initialized to fill the entire array for each texture array resource.
To fix 2, the root cause was that a texture array is considered a single descriptor. All elements in the descriptor must have the same state (SHADER_RESOURCE vs RENDER_TARGET). 

So, the fundamental design we were using for Post Process textures of:
-Bind the entire texture_array as SHADER_RESOURCE.
-Bind texture_array[0] as RENDER_TARGET.
-Render the post process pass [0] to texture_array[0].
-Bind the texture_array[0] AS SHADER_RESOURCE.
-Bind texture_array[1] as RENDER_TARGET.
-Use texture_array[1] as RENDER_TARGET while reading texture_array[0] as SHADER_RESOURCE.

Was causing this error, because you cannot have subresources in different states.

The fix was to create a SinkBufferResource for every post process pass. This is done dynamically by creating a single new SinkBufferDesc in ShaderBindings, and hijacking the Max Resources value. We create one SinkBufferResource (each with a single element) per count up to Max Resources. 
However, there is only one SinkBufferResource in the PipelineResourceSignature, because we will only every bind one SinkBufferResource as a SHADER_RESOURCE at any given time. The SinkBufferResource is a dynamic variable, so each time the post process pass requires a different SinkBufferResource as a SHADER_RESOURCE, we dynamically change the variable that the PipelineResourceSignature is pointing to.

The new design is similar to:
-Initialize the Color and Depth Sink arrays as RENDER_TARGET
-Do all the non-post process rendering
-Bind the entire Color and Depth Sink arrays as SHADER_RESOURCE
-Render the post process pass [0] to post process sink buffer 0.
-Bind post process sink buffer 0 as SHADER_RESOURCE
-Render the post process pass [1] to post process sink buffer 1 while reading post process sink buffer 0 as SHADER_RESOURCE.
(Not using element notation for post process sink buffer as they're not a single array,  but unique resources)

### Cleanup:
Since I was up in the passes anyway, I cleaned up and unified the implementations of all the pass creation backends. Eventually we can move these out to helper methods, but that was out of scope of this PR. For now, all the logic that is exactly the same across pipeline creation is organized next to each other so we can eventually cut these out into shared methods. Got rid of all the factory stuff for each different pass type, all is unified across passes so its all in the pass constructor.

I also changed how PassBackend was acquiring render target indices by extracting that logic out into PassManifest rather than having it be scattered throughout PassManifest and PassBackend.

I also changed the PassDesc interface to add a "useDepthTest" parameter, and changed the ColorTarget/DepthTarget/PostProcessTarget (new) specified in the PassDesc to be enums.

I also greatly simplified the BindRenderTarget and ClearRenderTarget interfaces.


jaremieromer:
Looks like the Linux Runner is broken, the https://sdk.lunarg.com/sdk/download/1.3.261.1/linux/vulkan-sdk-linux-x86_64-1.3.261.1.tar.gz?Human=true is a dead link.
I can take a look tonight

source/ncengine/graphics2/diligent/pass/PassManifest.cpp
```diff
@@ -118,24 +118,107 @@ void PassManifest::Clear()
...
-void PassManifest::SetMaxIndices(uint32_t colorRT, uint32_t depthRT, bool isMsaa)
+auto PassManifest::GetColorTargetIndex(ColorTarget colorTarget) const -> uint32_t
 {
-    auto& colorToModify = isMsaa ? m_colorSinkCountMsaa : m_colorSinkCount;
-    auto& depthToModify = isMsaa ? m_depthSinkCountMsaa : m_depthSinkCount;
+    switch (colorTarget)
+    {
+        case ColorTarget::None:
+        {
+            return std::numeric_limits<uint32_t>::max() - 1;
```
McCallisterRomer: I think you have constants defined for these that would be clearer.

source/ncengine/graphics2/diligent/pass/PassManifest.cpp
```diff
@@ -118,24 +118,107 @@ void PassManifest::Clear()
...
 }
+
+void PassManifest::RegisterTarget(ColorTarget colorTarget)
+{
+    if (colorTarget == ColorTarget::None || colorTarget == ColorTarget::Swapchain)
+    {
+        return;
+    }
+
+    if (std::ranges::find(m_colorSinkIndices, colorTarget) == m_colorSinkIndices.end())
```
McCallisterRomer: Its identical to this behind the hood, but there's `std::ranges::contains`.

source/ncengine/graphics2/diligent/pass/PassTypes.h
```diff
@@ -11,28 +11,47 @@
...
+    std::vector<uint32_t> color = std::vector<uint32_t>();
+    std::vector<uint32_t> depth = std::vector<uint32_t>();
+    uint32_t postProcess = 0u;
+};
 
-/* Depth Target Indices */
-constexpr auto MainDepth = 1u;
+struct Sinks
+{
+    uint32_t color;
```
McCallisterRomer: Default initializers on struct primitives are always a good idea.

source/ncengine/graphics2/diligent/pass/PassUtilities.cpp
```diff
@@ -153,9 +96,45 @@ auto CreateShaderFromSourceIfInitialized(ShaderFactory& shaderFactory, Diligent:
...
+        .depth = passManifest.GetDepthTargetIndex(passDesc.depthSink),
+        .postProcess = passManifest.GetPostProcessTargetIndex(passDesc.postProcessSink)
+    };
+}
+
+auto HasAnyColorSources(const Sources& sources) -> bool
+{
+    return std::ranges::any_of(sources.color, [](auto source)
+    {
+        return source != std::numeric_limits<uint32_t>::max() && source != std::numeric_limits<uint32_t>::max() - 1;
```
McCallisterRomer: Constants here again?

source/ncengine/graphics2/diligent/pass/PassUtilities.cpp
```diff
@@ -166,6 +145,51 @@ auto ToPassBaseId(const ShaderPaths& shaderPaths, std::string_view name) -> size
     return HashCombine(hashCode, shaderPaths.vertexShaderPath);
 }
 
-auto NoTargets() -> std::vector<uint32_t> { return std::vector<uint32_t>{}; }
+auto ToColorRenderTargetView(Diligent::ISwapChain& swapChain, SinkBufferResource& colorSinkBufferResource, uint32_t index, bool isMsaa) -> Diligent::ITextureView*
+{
+    if (index == std::numeric_limits<uint32_t>::max())
```
McCallisterRomer: constants a few times in this file

source/ncengine/graphics2/diligent/resource/PerPassResourceSignature.cpp
```diff
@@ -12,17 +12,23 @@ PerPassResourceSignature::PerPassResourceSignature(Diligent::IRenderDevice& devi
...
+    : m_postProcessSinkCount{postProcessSinksDesc.maxElementCount},
+      m_postProcessResourceKey{postProcessSinksDesc.resourceKey}
 {
-    const auto resources = std::array{
-        ToPipelineResourceDesc(colorSinkResourceDesc),
-        ToPipelineResourceDesc(depthSinkResourceDesc),
-        ToPipelineResourceDesc(sinkIndexResourceDesc)
+    auto postProcessTexDesc = SinkBufferDesc{m_postProcessResourceKey, Diligent::SHADER_TYPE_PIXEL, 1, true};
+
+    auto resources = std::vector{
```
McCallisterRomer: Why change to `vector` here?

source/ncengine/graphics2/diligent/resource/SinkIndexBufferResource.cpp
```diff
@@ -21,28 +21,35 @@ SinkIndexBufferResource::SinkIndexBufferResource(Diligent::IDeviceContext& conte
...
 
-    auto colorSourcesArray = std::array<uint32_t, 4u>{0u};
-    auto depthSourcesArray = std::array<uint32_t, 4u>{0u};
+    auto colorSourcesArray = std::array<int32_t, 4u>{-1};
+    auto depthSourcesArray = std::array<int32_t, 3u>{-1};
 
     for (auto i = 0u; i < colorSources.size(); i++)
     {
-        colorSourcesArray.at(i) = colorSources[i];
+        if (colorSources[i] != std::numeric_limits<uint32_t>::max())
```
McCallisterRomer: It was like this, but writing to the array based on the span size isn't safe, especially with `[]`. I don't know what the source object is, but span has a second template argument to declare a static size. Or you could just assert sizes are equal.

Constants would be nice here too.

Not sure on this one, but wasn't there another index type for max() - 1? Does that not come into play here? If it does, it would overflow the int, otherwise ignore me.

McCallisterRomer (approved):
Wowzer, nice job!

source/ncengine/graphics2/diligent/resource/PerPassResourceSignature.cpp
jaremieromer: Ah. I'll fix. At the time I changed it, I was going to be dynamically pushing back one desc per resource, rather than one total desc for all PP to share. Good eye!

source/ncengine/graphics2/diligent/resource/SinkIndexBufferResource.cpp
jaremieromer: I think we're safe here re: span not having a size. We default initialize the target array, so we don't have to worry about  the span being less than the array size, because those values will just remain the defaults.  And,  we already have a static assert that the span size must be equal to or less than the array size:

```
    NC_ASSERT(colorSources.size() <= 4u, "Only four color sources supported.");
    NC_ASSERT(depthSources.size() <= 3u, "Only three depth sources supported.");
```
LMK if I'm missing something though.

Re: Checking for max-1: I can't remember my reasoning on omitting this check, so I've added it back in.

PR #895, opened 2025-02-24
Jare style

### LINKED ISSUE #804: Investigate and fix banding problem revealed in adding lights

### LINKED ISSUE #805: Compute normal matrix CPU side

jaremieromer:
Lots to unpack here:

Major work done on artistic style shader for the game.

1. closes #805, closes #804 
2.  Refactored lighting algorithms for point, spot and directional lights.
3. Added option to allow hatching texture to materials
4. Added a Noise pass that applies noise to the scene. This noise can be dampened by a gradient, and is dampened by luminosity.
5. Alphabetized the sample manifest and `Prefabs.h/cpp` so I can find things easier.
6. Fixed the bug with the depth test for Particles I introduced when I refactored `PassBackend`
7. Moved the `PostProcessPropertyBuffer` to `PerPassResourceSignature` and changed the underlying buffer type to dynamic buffer. I also added a new buffer for the noise pass properties.
8. Added additional properties to MaterialData and the Post Process passes, and exposed editor UI for them.


include/ncengine/graphics/PostProcess.h
```diff
@@ -19,6 +20,7 @@ struct PostProcessPassFlag
     static constexpr auto None    = type{0};
     static constexpr auto Outline = type{1 << 0};
     static constexpr auto Fxaa    = type{2 << 0};
+    static constexpr auto Noise   = type{3 << 0};
```
McCallisterRomer: That's not a flag bit, use `1 << 2`. Fxaa, while technically correct, should be changed too.

sample/source/scenes/PhysicsTest.cpp
```diff
@@ -1296,13 +1298,18 @@ void PhysicsTest::Load(ecs::Ecs world, ModuleProvider modules)
...
+        Vector3{1.0f, 1.0f, 1.0f},
+        Vector3{1.0f, 1.0f, 1.0f},
+        1.0f
     );
 }
 
 void PhysicsTest::Unload()
 {
-    m_sampleUI->SetWidgetCallback(nullptr);
+    if (m_sampleUI)
```
McCallisterRomer: Nobody is able to null this pointer.

source/ncengine/graphics2/PostProcess.cpp
```diff
@@ -8,7 +8,7 @@ namespace nc
 {
 auto PassHasProperties(PostProcessPassFlag::type pass) -> bool
 {
-    return pass == PostProcessPassFlag::Outline;
+    return pass == PostProcessPassFlag::Outline || pass == PostProcessPassFlag::Noise;
```
McCallisterRomer: We should be able to check against a mask for these, but the pass id bug I commented about will cause it to fail, depending on the hash result:
```cpp
 constexpr auto mask = PostProcessPassFlag::Outline | PostProcessPassFlag::Noise;
 return pass & mask;
```

source/ncengine/graphics2/ShaderTypes.h
```diff
@@ -30,9 +30,21 @@ struct OutlinePassData
     Vector3 color = Vector3::Zero();
     float width = 1.0f;
     float depthThreshold = 0.8f;
+    float viewDirDepthThreshold = 0.4f;
```
McCallisterRomer: Was done previously (quite possibly by me), but the buffer for these is created based on `OutlinePassData` but is copied directly from `OutlinePassProperties`. They're identical, so its fine ATM, but easy to get out of sync. I feel like we can just use `using OutlinePassData = OutlinePassProperties` here to avoid having to do like an identity transformation between types.

source/ncengine/graphics2/ShaderTypes.h
```diff
@@ -30,9 +30,21 @@ struct OutlinePassData
...
     float normalThreshold = 0.4f;
 };
 
+// Object model for noise pass properties used by post processing effects (type: constant buffer)
+struct NoisePassData
```
McCallisterRomer, line 32: If we change `NoisePassProperties::noiseTex` from a `TextureView` to just the texture index, we can make this a using declaration too.

source/ncengine/graphics2/diligent/resource/PostProcessPropertyBufferResource.cpp
```diff
@@ -6,33 +6,58 @@
 #include <concepts>
 #include <ranges>
 
+namespace
+{
+auto ToNoisePassData(const nc::NoisePassProperties& properties) -> nc::graphics::NoisePassData
```
McCallisterRomer, line 11: Not needed, pending a previous comment.

source/ncengine/graphics2/frontend/subsystem/TransformCache.cpp
```diff
@@ -64,6 +65,7 @@ void TransformCache::UpdateMatrices(ecs::ExplicitEcs<Transform> ecs)
 
         const auto handle = static_cast<TransformDataHandle>(i);
         data[handle].modelMatrix = pool.Get(entity).TransformationMatrix();
+        data[handle].invModelMatrix = DirectX::XMMatrixTranspose(DirectX::XMMatrixInverse(nullptr, data[handle].modelMatrix));
```
McCallisterRomer: br00tal. This does result in a noticeable drop in performance. I don't think it will matter for the game though b/c we can pack all the statics up front.

McCallisterRomer (approved):

McCallisterRomer:
The shader looks really cool!

McCallisterRomer:
There's a bug in hiding with `PassManifest`/pass ids. I don't think it was introduced here, and I doubt it causing any problems with the current state of things. The member `PassDesc::id` in the final `PostProcessPass` is calculated based on a hash. All of the other pass ids are based on flags. This will cause any flag-based decisioning applied to the final pass to act on it randomly.

Also `PassManifest::m_ids` is referenced searched in `PassManifest::RegisterPass()`, but nothing ever sets those ids.

source/ncengine/graphics2/ShaderTypes.h
jaremieromer: Not doing, based on our conversation today. (We don't want to have an integer counter in the UI, but rather the texture interface.)

source/ncengine/graphics2/frontend/subsystem/TransformCache.cpp
jaremieromer: Removed

source/ncengine/graphics2/diligent/resource/PostProcessPropertyBufferResource.cpp
jaremieromer: Leaving it in - per conversation today reminding about needing the Texture View interface in the UI.

source/ncengine/graphics2/PostProcess.cpp
jaremieromer: Done

jaremieromer:
> There's a bug in hiding with `PassManifest`/pass ids. I don't think it was introduced here, and I doubt it causing any problems with the current state of things. The member `PassDesc::id` in the final `PostProcessPass` is calculated based on a hash. All of the other pass ids are based on flags. This will cause any flag-based decisioning applied to the final pass to act on it randomly.
> 
> Also `PassManifest::m_ids` is referenced searched in `PassManifest::RegisterPass()`, but nothing ever sets those ids.

Fixed. `m_finalPass` is never added to m_postProcessPasses, so we never do any checks against the pass flags during rendering. The only place we used the flag was `MakePostProcessPassInstances`, which I've fixed by circumventing that method for `m_finalPass`. Now, `m_finalPass` is completely separate from flags.

Great catch

include/ncengine/graphics/PostProcess.h
```diff
@@ -17,10 +17,10 @@ struct PostProcessPassFlag
 {
     using type = uint64_t;
 
-    static constexpr auto None    = type{0};
-    static constexpr auto Outline = type{1 << 0};
-    static constexpr auto Fxaa    = type{2 << 0};
-    static constexpr auto Noise   = type{3 << 0};
+    static constexpr auto None    = type{1 << 0};
```
McCallisterRomer: Decrement each of these - e.g. None should be 0

source/ncengine/graphics2/NcGraphicsImpl2.cpp
```diff
@@ -162,7 +162,7 @@ NcGraphicsImpl2::NcGraphicsImpl2(const config::GraphicsSettings& graphicsSetting
             std::vector<PassDesc>
             {
                 PassDesc{
-                    .id = MaterialPassFlag::Depth,
+                    .flag = MaterialPassFlag::Depth,
```
McCallisterRomer, line 165: sick

PR #904, opened 2025-03-12
Generate prefabs file from asset manifest

McCallisterRomer:
Adding functionality to `nc-convert` to generate source files with asset paths/ids/views.


source/ncconvert/Usage.h
```diff
@@ -0,0 +1,135 @@
+#pragma once
```
jaremieromer, line 1: This is nice in its own file

jaremieromer:
I'll jump on discord tomorrow - this is really cool and I'm assuming works great but I can't get it to do what I'm expecting it to do based on the documentation. Once we talk I can suggest updates to the docs that will help with what I'm misunderstanding today

docs/nc-convert.md
```diff
@@ -53,6 +55,11 @@ You can also use a JSON manifest file to convert many files at once:
 `nc-convert` will skip files that are already up-to-date when using a manifest.
 Relative paths within `globalOptions` are interpreted relative to the manifest.
 
+Source files enumerating assets in a manifest can be automatically generated with:
+```
+> nc-convert -g manifest.json -o out/dir -r out::namespace
```
jaremieromer, line 60: The `out::namespace` here is not really intuitive for me. I get what it's saying after seeing the comment on how to use in Prefabs, then trying it out. 

I wonder if we could add an example comment like:

`Example: nc-convert -g sample/assets/manifest.json -o sample/source/shared -r nc::sample`

jaremieromer (approved):
I tested this out and it's going to save me an extraordinary amount of time. NICE

PR #905, opened 2025-03-13
Shape Builder

McCallisterRomer:
Adding `ShapeType::Compound` and a `ShapeBuilder`.

Primary use case is for doing shape tests when placing multi-body objects in rustbelt - if you're trying to place a complicated object, we don't want to have to test against all of its shapes manually + make the constraints that hold it together. We can instead automatically build a bunch of runtime assets that represent each object to use during placement, and build the full configuration once placed.

Secondary use case is to replace `FixedConstraint`s in some of the multi body objects (sign attached to sign post). It is more stable in extreme cases to represent that as a compound set of primitives. Untested, but I also assume its less computationally intensive.

Future use case is to do offline cooking of static geometry. Serialization/asset support is still needed for that. Most static physics objects can be combined into a single shape asset and attached to one body. Mesh shapes and anything that requires specific physics materials would have to be kept separate. This would result in faster level loads, fewer total transforms, fewer bodies in the simulation, etc.


jaremieromer:
Seeing a crash when I load the PhysicsTest scene, but only the second load. AKA - if I load GraphicsTest or BenchmarksTest, then load PhysicsTest - we crash on trying to build the compound shape for the bridge.

`AssetId '42' already assigned to a runtime asset`
Looks like we need to do a clear of m_runtimeAssets on Scene Unload maybe?

I've attached the call stack - here are steps to reproduce.

1. Launch the sample
2. Click again on PhysicsTest to load it again

```
Sample.exe!_CxxThrowException(void * pExceptionObject, const _s__ThrowInfo * pThrowInfo) Line 81 (d:\a\_work\1\s\src\vctools\crt\vcruntime\src\eh\throw.cpp:81)
Sample.exe!nc::physics::ShapeFactory::AddRuntimeAsset(JPH::Ref<JPH::Shape> shape, unsigned __int64 id) Line 144 (c:\Source\NcEngine\source\ncengine\physics\jolt\ShapeFactory.cpp:144)
Sample.exe!nc::physics::NcPhysicsImpl::AddRuntimeCompoundShape(const nc::CookedShape & cookedShape, unsigned __int64 id) Line 357 (c:\Source\NcEngine\source\ncengine\physics\NcPhysicsImpl.cpp:357)
Sample.exe!nc::sample::BuildBridge(nc::ecs::EcsInterface<2> world, nc::NcPhysics & ncPhysics) Line 609 (c:\Source\NcEngine\sample\source\scenes\PhysicsTest.cpp:609)
Sample.exe!nc::sample::PhysicsTest::Load(nc::ecs::EcsInterface<2> world, nc::ModuleProvider modules) Line 1271 (c:\Source\NcEngine\sample\source\scenes\PhysicsTest.cpp:1271)
Sample.exe!nc::SceneManager::LoadQueuedScene(nc::ecs::EcsInterface<2> world, nc::ModuleRegistry & modules) Line 80 (c:\Source\NcEngine\source\ncengine\scene\NcSceneImpl.cpp:80)
Sample.exe!nc::NcEngineImpl::Run() Line 189 (c:\Source\NcEngine\source\ncengine\engine\NcEngineImpl.cpp:189)
Sample.exe!nc::NcEngineImpl::Start(std::unique_ptr<nc::Scene,std::default_delete<nc::Scene>> initialScene) Line 108 (c:\Source\NcEngine\source\ncengine\engine\NcEngineImpl.cpp:108)
Sample.exe!main(int argc, char * * argv) Line 69 (c:\Source\NcEngine\sample\source\SampleMain.cpp:69)
Sample.exe!invoke_main() Line 79 (d:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:79)
Sample.exe!__scrt_common_main_seh() Line 288 (d:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:288)
Sample.exe!__scrt_common_main() Line 331 (d:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:331)
Sample.exe!mainCRTStartup(void * __formal) Line 17 (d:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_main.cpp:17)
kernel32.dll!00007ffd54b6e8d7() (Unknown Source:0)
ntdll.dll!00007ffd5557bf6c() (Unknown Source:0)
```

McCallisterRomer:
> Seeing a crash when I load the PhysicsTest scene, but only the second load. AKA - if I load GraphicsTest or BenchmarksTest, then load PhysicsTest - we crash on trying to build the compound shape for the bridge.
> 
> `AssetId '42' already assigned to a runtime asset` Looks like we need to do a clear of m_runtimeAssets on Scene Unload maybe?
> ```

Oh, of course. Nice catch!

jaremieromer (approved):

PR #906, opened 2025-03-14
Update AudioSource

McCallisterRomer:
- Adding some `AudioSource` functions for TuneScape so we can pause/resume songs.
- Updating `AudioSource` to operate entirely with `AudioClipView`s instead of paths + views.
- Removing `audio` namespace from types in /include
- Fixed bug when removing clips from `AudioSource` in editor


include/ncengine/audio/AudioSource.h
```diff
@@ -55,42 +60,56 @@ class AudioSource : public ComponentBase
         auto IsPlaying() const noexcept -> bool { return m_properties.flags & AudioSourceFlags::Play; }
 
         /**
-         * @brief Get the index of the most recently played clip.
+         * @brief Preemptively set the next clip to be played. This can be started with Resume() instead of Play().
+         * @note Stops any playing clips and resets the play position.
```
jaremieromer, line 64: I don't understand why a `Queue` should stop the currently playing clip. Shouldn't it queue it up to be played after the currently playing clip completes naturally?

jaremieromer (approved):

PR #910, opened 2025-03-21
Asset/Asset Generation Cleanup

McCallisterRomer:
Several changes to assets and asset source generation. Big picture is default assets are included when generating a prefab files, and diffuse/normal textures can be combined in a bulk load. This allows the generated code to load all engine and project assets for a given type in one shot.

- `DefaultAssets.h` is now part of `NcAsset` to maintain build structure. The paths have been moved under a /default subdirectory.
- Replaced `AssetFlags` with `AssetSubtype` (previously an internal type in `nc-convert`).
  - Removed these from a bunch of functions that don't use them. 
- Added an additional load function to `AssetService` to take a `span<AssetSubtype>` along with the paths. Only used by textures currently, but it allows loading diffuse + normals in one bulk call.
- Removed the `isExternal` from asset load/manager functions. Its hasn't been used anywhere for a while.
- Tedious formatting to asset managers that probably wasn't worth the time spent.
- Found + fixed serialization bug with `AudioSource`.


include/ncasset/DefaultAssets.h
```diff
@@ -0,0 +1,74 @@
...
+constexpr auto WheelMesh = "default/wheel.nca";
+
+/** @brief Name of the default base color texture asset. */
+constexpr auto DefaultBaseColor = "default/color.nca";
+
+/** @brief Name of the default normal texture asset. */
+constexpr auto DefaultNormal = "default/normal.nca";
+
+/** @brief Name of the default roughness texture asset. */
+constexpr auto DefaultRoughness = "default/metallic.nca";
```
jaremieromer: We actually don't need DefaultRoughness and DefaultMetallic anymore until we do a different shader

resources/manifest.json
```diff
@@ -67,30 +67,30 @@
...
         "diffuse" : [
             {
                 "sourcePath": "raw/image/DefaultBaseColor.png",
-                "assetName": "texture/DefaultBaseColor"
+                "assetName": "texture/default/color"
             },
             {
                 "sourcePath": "raw/image/DefaultMetallic.png",
-                "assetName": "texture/DefaultMetallic"
+                "assetName": "texture/default/metallic"
```
jaremieromer: can delete

sample/source/shared/GeneratedAssets.cpp
```diff
@@ -156,40 +210,47 @@ const auto g_diffusePaths = std::array{
...
-nc::asset::TextureView diffuse_teal{};
-nc::asset::TextureView diffuse_yellow{};
-nc::asset::TextureView effect_linear_hatch{};
-nc::asset::TextureView effect_noise{};
-nc::asset::TextureView normal_cave{};
-nc::asset::TextureView normal_guy{};
-nc::asset::TextureView normal_ogre{};
-nc::asset::TextureView normal_skeleton{};
+const auto g_subtypes = std::array{
+    nc::asset::AssetSubtype::DiffuseTexture,
```
jaremieromer: No need to change now, but upon further thought 'Diffuse' isn't the right name for these. The difference between these and the normals is the texture format. I would actually call these either `ColorTexture` or `RgbaTexture`

source/ncconvert/builder/Manifest.cpp
```diff
@@ -59,13 +60,13 @@ auto IsUpToDate(const nc::convert::Target& target) -> bool
...
-auto ToAssetSubtype(std::string_view str) -> nc::convert::AssetSubtype
+auto ToAssetSubtype(std::string_view str) -> nc::asset::AssetSubtype
 {
-    using enum nc::convert::AssetSubtype;
-    if (str == "diffuse")  return DiffuseTexture;
+    using enum nc::asset::AssetSubtype;
     if (str == "normal")   return NormalTexture;
-    if (str == "particle") return ParticleTexture;
-    if (str == "effect")   return EffectTexture;
+    if (str == "diffuse")  return DiffuseTexture;
```
jaremieromer: 'diffuse' still works a a subdirectory for the textures that are not normal, particle or effect. But the commonality between diffuse, particle and effect is the texture format

source/ncengine/asset/manager/TextureAssetManager.cpp
```diff
@@ -37,7 +39,41 @@ bool TextureAssetManager::Load(const std::string& path, bool isExternal, asset_f
     return true;
 }
 
-bool TextureAssetManager::Load(std::span<const std::string> paths, bool isExternal, asset_flags_type flags)
+auto TextureAssetManager::Load(std::span<const std::string> paths, std::span<const AssetSubtype> subtypes) -> bool
+{
+    NC_ASSERT(paths.size() == subtypes.size(), "Number of subtypes does not match number of asset paths");
```
jaremieromer, line 44: Since we're introducing a requirement for folder structure - we should add that to documentation

jaremieromer (approved):

sample/source/shared/GeneratedAssets.cpp
McCallisterRomer: No better time to change than now.

include/ncasset/DefaultAssets.h
McCallisterRomer: Totally delete both?

source/ncengine/asset/manager/TextureAssetManager.cpp
McCallisterRomer: blanking for some reason, what specific requirement are you referring to?

PR #911, opened 2025-03-21
Remove Duplicate Transform Update

McCallisterRomer:
In #847 I moved the transform update to a new task but didn't remove the old call, so we've been doing it twice. Fixing.


jaremieromer (approved):
Great find!!!!

PR #912, opened 2025-03-23
Shape Support Functions

McCallisterRomer:
Adding functions to check the furthest vertex and half extents in a direction for `CookedShape`s. Used in rustbelt to snap objects to the outside of vehicles when placing them in the build phase - ray cast gives us a point and normal on the vehicle, we need to know how far to push the objects apart. This gets complicated for compound shapes, so we'd like these functions here where we can access internals.


include/ncengine/physics/SupportFunctions.h
```diff
@@ -0,0 +1,67 @@
...
+struct Vector3;
+
+/** @brief Wrapper around a scratch buffer used by support functions. */
+class SupportBuffer
+{
+    struct Impl;
+
+    public:
+        /** @brief Get the global SupportBuffer for use with single-threaded calls to support functions. */
+        static auto GetDefault() -> SupportBuffer&;
```
jaremieromer, line 23: Can we make 'return value is a reference' the exception to when we use trailing return types?
I really don't like seeing auto without & here.

jaremieromer (approved):

include/ncengine/physics/SupportFunctions.h
McCallisterRomer: I don't understand

McCallisterRomer: Do you mean do `static auto& GetDefault() -> SupportBuffer&`? I so, that isn't allowed.

McCallisterRomer: You can do `auto&` if trailing return is omitted, but we can't do that without the definition also being in the header.

jaremieromer: ah ok - must be looking at too much VB.net these days because what you're saying now makes sense. Feel free to ignore

PR #915, opened 2025-03-27
Point Shadows

### LINKED ISSUE #890: Add point light shadows
Add omnidirectional shadows

jaremieromer:
closes #890 

Implemented point shadow mapping.
I've added a new pass for the cube map shadow generation.
Extracted the view projection matrix out of `LightData` and into a separate Structured Buffer.

image


resources/shaders/PointShadowMapSkinned.vsh
```diff
@@ -0,0 +1,144 @@
...
+    uint lightCount;
+    float nearClip;
+    float farClip;
+};
+
+void main(in VSInput VSIn, uint InstanceID : SV_InstanceID, out PSInput PSIn)
+{
+    INSTANCE_DATA instance = INSTANCE_BUFFER[InstanceID];
+    float4 pos = float4(VSIn.Pos, 1.0);
+    float3 normal = VSIn.Normal;
```
McCallisterRomer: I can fix it when I merge this into my changes, but this is the thing glslangValidator catches that's somehow missed when we compile via Diligent. Should be a float4 since its used in a matrix multiply. Can't tell if it actually matters though.

source/ncengine/graphics2/NcGraphicsImpl2.cpp
```diff
@@ -162,23 +162,41 @@ NcGraphicsImpl2::NcGraphicsImpl2(const config::GraphicsSettings& graphicsSetting
...
+                    .shaderPaths = ShaderPaths{.vertexShaderPath = "UniShadowMap.vsh"},
                     .shadowMapSink = ShadowMapTarget::Uni,
                     .isMsaa = false,
                     .useDepthTest = true
                 },
                 PassDesc{
-                    .flag = MaterialPassFlag::Shadow,
-                    .name = "Shadow",
+                    .flag = MaterialPassFlag::UniShadow,
+                    .name = "UniShadow",
```
McCallisterRomer: Do we want unique names for skinned/non? We have them for toon.

McCallisterRomer (approved):

source/ncengine/graphics2/NcGraphicsImpl2.cpp
jaremieromer: Yes, fixed

PR #916, opened 2025-03-28
Shader Compilation

McCallisterRomer:
Updates to use pre-compiled shaders, since launch times are getting kind of long. As discussed, you will very likely need to update your vulkan sdk version in order to compile. Note that `DiligentEngine_tests` fails with validation errors (synchronization issue discussed) with the latest SDK. I have to wait to update the version in CI until we sort that out.

Currently, shaders can be compiled with `shaders.ps1`, just like before. I'm still planning to add this to `nc-convert` eventually.

I've added a simple `ShaderCache` since some of the shaders are reused several times. This is populated once before building the passes. Its deleted once all passes have been built.

A few minor shader issues popped up, like vec3/4 mismatches. Nothing impactful.


script/shaders.ps1
```diff
@@ -0,0 +1,89 @@
...
+        $FileName = "$($_.BaseName)Vertex.spv"
+    } else {
+        return
+    }
+
+    Write-Output "Compiling $_"
+    $InputFile = Join-Path $InputDir $_.Name
+    $OutputFile = Join-Path $TempDir $FileName
+    $Args = @(
+        '-V',
```
jaremieromer: Can you add comments to this file explaining what each of these do?

script/shaders.ps1
```diff
@@ -0,0 +1,89 @@
...
+    if ($LASTEXITCODE -eq 0) {
+        Write-Host "COMPILATION SUCCEEDED: $_ -> $OutputFile" -ForegroundColor Green
+    }
+    else {
+        Write-Host "COMPILATION FAILED: $_" -ForegroundColor Red
+        $FailureCount++
+    }
+}
+
+Write-Host "--Optimizing Shaders--"
```
jaremieromer: NICE!

script/shaders.ps1
```diff
@@ -0,0 +1,89 @@
...
+        '-D',
+        "-P$Preamble",
+        '-e', 'main',
+        '-fhlsl_functionality1',
+        "-S", $Stage,
+        "-o", $OutputFile,
+        $InputFile
+    )
+
+    & $Compiler @Args
```
jaremieromer, line 67: Can you add "-g" (compile in debug mode) to the possible parameters? I haven't been able to figure out how to do this since switching to Diligent and it would make a HUGE difference. I'm assuming we also don't want to optimize if running in this mode.

Debug allows you to see the actual shader source code when stepping through in Render Doc, and its hugely valuable

source/ncengine/graphics2/diligent/pass/MaterialPass.cpp
```diff
@@ -25,10 +25,8 @@ auto CreatePipeline(Diligent::IRenderDevice& device,
     ci.PSODesc.Name = passDesc.name.data();
     ci.PSODesc.ResourceLayout.DefaultVariableType = SHADER_RESOURCE_VARIABLE_TYPE_STATIC;
 
-    RefCntAutoPtr<IShader> pixelShader = CreateShaderFromSourceIfInitialized(shaderFactory, SHADER_TYPE_PIXEL, passDesc.shaderPaths);
-    RefCntAutoPtr<IShader> vertexShader = CreateShaderFromSourceIfInitialized(shaderFactory, SHADER_TYPE_VERTEX, passDesc.shaderPaths);
-    ci.pPS = pixelShader;
-    ci.pVS = vertexShader;
+    ci.pPS = shaders.pixelShader;
```
jaremieromer, line 28: nice

source/ncengine/graphics2/diligent/pass/MaterialPass.cpp
```diff
@@ -79,12 +77,16 @@ auto CreatePipeline(Diligent::IRenderDevice& device,
...
 {
 MaterialPass::MaterialPass(Diligent::IRenderDevice& device,
-                           ShaderFactory& shaderFactory,
+                           const PipelineShaders& shaders,
                            ShaderBindings& shaderBindings,
                            const PassManifest& passManifest,
                            const PassDesc& passDesc,
                            uint32_t numSamples)
-    : Pass{CreatePipeline(device, shaderFactory, shaderBindings, passDesc, numSamples), GetSinks(passManifest, passDesc), GetSources(passManifest, passDesc)},
+    : Pass{
```
jaremieromer, line 85: easier to read this way

source/ncengine/graphics2/diligent/pass/PipelineShaders.cpp
```diff
@@ -0,0 +1,71 @@
+#include "PipelineShaders.h"
+#include "graphics2/diligent/ShaderFactory.h"
+
+#include "ncutility/NcError.h"
+
+#include <array>
+
+namespace
+{
```
jaremieromer, line 9: `using namespace nc::graphics::shader` might go a long way here

script/shaders.ps1
```diff
@@ -0,0 +1,89 @@
...
+if (-not (Test-Path -Path $TempDir)) {
+    New-Item -Path $TempDir -ItemType Directory | Out-Null
+}
+
+Write-Host "--Compiling Shaders--"
+Get-ChildItem -Path $InputDir | Where-Object { $_.Extension -eq '.psh' -or $_.Extension -eq '.vsh' } |
+ForEach-Object {
+    if ($_.Extension -eq '.psh') {
+        $Stage = 'frag'
+        $FileName = "$($_.BaseName)Fragment.spv"
```
jaremieromer: It's an annoying change to make - but Diligent seems to use "Pixel" instead of "Fragment" everywhere. Can we change all uses of "Fragment" naming to "Pixel"? Then we match all throughout the codebase

jaremieromer (approved):
Great work! This is such an improvement. Just naming comments really, outside of adding a debug option to the script.

script/shaders.ps1
McCallisterRomer: Done

script/shaders.ps1
McCallisterRomer: Done

script/shaders.ps1
McCallisterRomer: Lets try to go over this today. Debug symbols exist, but spirv-opt may be generating a wildly different result. Not sure how that will affect your workflow. If we want to have two distinct build options, I'd prefer doing something like a debug/release directory that we can select from via config. Because the binaries are checked into source control, I don't want to leave open the possibility of accidentally checking in debug versions. I suppose we don't have to check in the binaries at all. Idk, worth a discussion.

PR #917, opened 2025-03-28
Factor Out Common Shader Types

McCallisterRomer:
- Extracting common shader types to `ShaderTypes.fxh`.
- Extracting common animation functions to `Animation.fxh`.
- Moved shader source files to souce subdirectory.
- Normalized order of input layout/bindings across all shaders.


jaremieromer (approved):
This is wonderful!

Issue #751, opened 2024-10-17
Investigate updating texture format + loading


McCallisterRomer:
Nca texture files are too large. We have three options:
1. Continue to use current nca format for textures, but add compression to the blob.
2. Update nca format so the texture blob is a known compressed format (like DXT).
3. Allow loading textures directly from any file type.

PR #918, opened 2025-03-30
Texture Compression

McCallisterRomer:
preliminary changes for #751 

Initial work for texture compression + mipmap generation. No changes here are visible in the engine, just a good stopping point. There will be more to get this fully integrated.


source/ncconvert/utility/Image.cpp
```diff
@@ -0,0 +1,168 @@
...
+{
+auto MakeRawImage(int width,
+                  int height,
+                  const unsigned char fillRGBA[4]) -> RawImage
+{
+    const auto numBytes = width * height * 4;
+    auto bytes = (unsigned char*)std::malloc(numBytes);
+    for (auto i = 0; i < numBytes; i += 4)
+    {
+        std::memcpy(bytes + i, fillRGBA, 4);
```
jaremieromer, line 56: Am I understanding this correctly - does this always set every texture's Alpha value to 1.0 (aka 255?)

source/ncconvert/utility/Image.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+    constexpr auto blockSize = 16;
+    const auto useAlpha = TextureFormatHasAlpha(format);
+    const auto numBlocks = (m_width / 4) * (m_height / 4);
+    auto compressedData = std::vector<unsigned char>(numBlocks * blockSize);
+    unsigned char block[blockSize * 4];
+
+    // compress in 4x4 RGBA blocks
+    const auto src = m_data.get();
+    for (auto y = 0; y < m_height; y += 4)
```
jaremieromer: This might be a case for more specific index variable names

jaremieromer (approved):

source/ncconvert/utility/Image.cpp
McCallisterRomer: No, this initializes a new texture by replicating one pixel across the whole thing. Used by tests and when we have to pad an image to power of two dimensions. For the latter, the original image is written on top of a larger black image.

source/ncconvert/utility/Image.cpp
McCallisterRomer: fixed

PR #919, opened 2025-03-31
Generate Mipmaps Offline

McCallisterRomer:
part of #751 

Updating mipmap generation to run as part of nc-convert texture conversion rather than generating at runtime. I've recooked all textures here to include mip levels, but in the next PR, I'll add an option to the manifest to toggle generation.


source/ncconvert/converters/TextureConverter.cpp
```diff
@@ -61,12 +62,35 @@ auto TextureConverter::ImportTexture(const std::filesystem::path& path) -> asset
         throw NcError("Invalid input file: ", path.string());
     }
 
+    /** @todo 751 Currently, texture format in the asset is not used and mips are always generated. Should:
```
jaremieromer, line 65: 👍

source/ncengine/graphics2/diligent/resource/ResourceTypes.cpp
```diff
@@ -141,54 +145,39 @@ void SetArrayRegion(Diligent::IShaderResourceVariable* variable, std::span<Dilig
...
+        .height = 1u,
+        .pixelData = std::vector<unsigned char>{0x1A, 0x2A, 0x3A, 0x4A}
     };
 
-    auto barriers = std::vector<Diligent::StateTransitionDesc>();
-    barriers.reserve(arraySize);
-
-    auto dummyTextures = std::vector<Diligent::RefCntAutoPtr<Diligent::ITexture>>();
-    auto dummyViews = std::vector<Diligent::IDeviceObject*>();
+    auto subResource = Diligent::TextureSubResData{dummyTexture.pixelData.data(), dummyTexture.width * 4u};
```
jaremieromer: Does this 4u need to be equivalent to whatever's returned by `GetMinimumDimension`?

jaremieromer (approved):

source/ncengine/graphics2/diligent/resource/ResourceTypes.cpp
McCallisterRomer: No, diligent wants the stride in bytes. Width * pixelSize. At some point we'll probably want other pixel sizes (many particles could be grayscale + alpha), but for now we force to 4 channels at conversion time.

McCallisterRomer: Although we have a constant for that. I'll update here to use it so we break if/when we change things.

PR #921, opened 2025-03-31
Skybox Support

### LINKED ISSUE #891: Add skybox support

jaremieromer:
closes #891 

image

image

Adds ability to render skyboxes from Cube Map assets. 
Added a new window in the editor to configure the skybox.


GeneratedAssets.cpp
```diff
@@ -0,0 +1,273 @@
+/** @note This file is auto generated. */
```
McCallisterRomer: This file and the header are not in the right spot. Should set the output to sample/source/shared. I don't see one there. Did anything change here?

source/ncengine/graphics/NcGraphicsImpl.cpp
```diff
@@ -157,6 +158,12 @@ namespace nc::graphics
         m_systemResources.environment.SetSkybox(path);
     }
 
+    auto NcGraphicsImpl::GetSkybox() const -> nc::asset::AssetId
```
McCallisterRomer, line 161: No need to change it, but this one shouldn't have to be updated, right? I don't think this file should be built by anything.

source/ncengine/graphics2/NcGraphicsImpl2.cpp
```diff
@@ -401,6 +418,7 @@ void NcGraphicsImpl2::SetPostProcessEffectProperties(PostProcessEffectId effectI
 void NcGraphicsImpl2::OnBeforeSceneLoad()
 {
     m_frontend.OnBeforeSceneLoad();
+    ClearEnvironment();
```
McCallisterRomer: Is this the right place, rather than `Clear()`?

source/ncengine/graphics2/diligent/pass/PassBackend.cpp
```diff
@@ -263,6 +266,62 @@ void PassBackend::RenderShadowPass(IDeviceContext& context,
...
+    if (!environmentRenderState.useSkybox || !m_skyboxPass)
+    {
+        return;
+    }
+
+    // We need to transition resource state manually for the color and depth target here.
+    auto* colorTargetTexture = perPassResourceSignature.GetColorSinksResource().GetTexture(m_skyboxPass->sinks.color);
+    auto* depthTargetTexture = perPassResourceSignature.GetDepthSinksResource().GetTexture(m_skyboxPass->sinks.depth);
+
+    auto barriers = std::vector<Diligent::StateTransitionDesc>();
```
McCallisterRomer: I think this could just be an array.

source/ncengine/graphics2/diligent/resource/CubeMapBufferResource.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+    auto barriers = std::vector<StateTransitionDesc>();
+    barriers.reserve(cubeMapCount);
+    m_cubeMaps.reserve(m_cubeMaps.size() + cubeMapCount);
+    m_shaderResourceViews.reserve(m_shaderResourceViews.size() + cubeMapCount);
+
+    for (const auto& [cubeMap, id] : cubeMaps)
+    {
+        // Assuming cubeMap.pixelData contains data for all 6 faces in order: +X, -X, +Y, -Y, +Z, -Z
+        auto faceSubResources = std::vector<Diligent::TextureSubResData>(6);
```
McCallisterRomer: array again?

source/ncengine/graphics2/diligent/resource/CubeMapBufferResource.cpp
```diff
@@ -0,0 +1,104 @@
...
+        const uint32_t faceSizeInBytes = cubeMap.faceSideLength * cubeMap.faceSideLength * 4; // Assuming RGBA format
```
McCallisterRomer: Will you do the same thing I did in my recent * 4 case, and use `CubeMap::numChannels`. In case we extend formats in the future, we'll get a compile time error because `numChannels` won't be static anymore.

source/ncengine/graphics2/diligent/resource/CubeMapBufferResource.cpp
```diff
@@ -0,0 +1,104 @@
...
+        
+        for (uint32_t face = 0; face < 6; ++face)
+        {
+            // Offset the data pointer for each face
+            const void* faceData = static_cast<const uint8_t*>(cubeMap.pixelData.data()) + (face * faceSizeInBytes);
+            faceSubResources[face] = TextureSubResData{
+                faceData,
+                cubeMap.faceSideLength * 4u // Stride per row
```
McCallisterRomer: same as above

source/ncengine/graphics2/diligent/resource/ResourceTypes.cpp
```diff
@@ -143,6 +186,62 @@ void SetArrayRegion(Diligent::IShaderResourceVariable* variable, std::span<Dilig
     );
 }
 
+void InitializeCubeArray(Diligent::IDeviceContext& context, Diligent::IRenderDevice& device, Diligent::IShaderResourceVariable* variable, uint32_t arraySize, bool transition)
```
McCallisterRomer, line 189: Will you clean this up the same way I did `InitializeArray` below. Only one texture need to be created, the barriers are only needed if transition is true, and there's no need to loop at all, just use the `vector(count, item)` constructor for the barriers/views.

source/ncengine/graphics2/frontend/subsystem/EnvironmentRenderState.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+#include "ncengine/asset/AssetViews.h"
+
+namespace nc::graphics
+{
+struct EnvironmentRenderState
+{
+    uint32_t skyboxIndex;
```
McCallisterRomer: Needs default initializers?

source/ncengine/graphics2/frontend/subsystem/EnvironmentSubsystem.h
```diff
@@ -0,0 +1,35 @@
...
+#include <cstdint>
+
+namespace nc::graphics
+{
+class EnvironmentSubsystem
+{
+    public:
+        void SetSkybox(const std::string& skyboxPath)
+        {
+            m_useSkybox = true;
```
McCallisterRomer: Small, but I think the bool is implied by the other members, no? Like `auto UseSkybox() const { return m_skyBoxId != NullAssetId; }`.

source/ncengine/graphics2/frontend/subsystem/EnvironmentSubsystem.h
```diff
@@ -0,0 +1,35 @@
...
+            m_skyboxIndex = asset::AssetService<asset::CubeMapView>::Get()->Acquire(skyboxPath).index;
```
McCallisterRomer: Just acquire once, compiler will not be able to optimize b/c it can't guarantee thread safety of `Acquire`.

McCallisterRomer: Or you could store the `TextureView` as a member.

McCallisterRomer (approved):

source/ncengine/graphics2/diligent/resource/ResourceTypes.cpp
McCallisterRomer: Double check my thinking there, does there even need to be more than one barrier if there's only one texture?

source/ncengine/graphics/NcGraphicsImpl.cpp
jaremieromer: It shouldn't, no. I'm not sure what our 'maintain vs toss' strategy is for NcGraphics. I know we're tossing it - but I figured I'll at least update it to satisfy interfaces until we scrap it completely.

source/ncengine/graphics2/NcGraphicsImpl2.cpp
jaremieromer: Ah. I'll move to `Clear()`. The existence of this method made me think that `Clear()` must be end of engine teardown stuff which now that I see we have a destructor doesn't actually make sense anyway.

PR #925, opened 2025-04-01
Finish Texture Compression

### LINKED ISSUE #751: Investigate updating texture format + loading

McCallisterRomer:
resolves #751

Finishing texture compression/mipmap generation
- Adding options to `nc-convert` to select texture format + enable/disable mip generation.
  - Adding defaults `TargetOptions` to the manifest to make things more manageable.
- Detecting format based on `Texture::TextureFormat` instead of `AssetSubtype`. (will remove in a separate PR b/c its a sprawling change).

Savings with this will always be significant, even with baking in mipmaps, since mips add a flat 33% and the compression ratio is guaranteed 4:1 or 8:1, depending on format:
uncompressed sample textures without mips: 37 mb
compressed samples textures with mips: 10 mb


resources/manifest.json
```diff
@@ -2,6 +2,25 @@
     "globalOptions": {
         "outputDirectory": "./nca"
     },
+    "defaultMeshOptions": {
```
jaremieromer, line 5: These are great

source/ncconvert/Usage.h
```diff
@@ -5,16 +5,22 @@ namespace nc::convert
...
+    -t <asset type>            Specify asset type for a single target.
+    -s <source>                Parse a single asset from <source>.
+    -n <name>                  Specify the asset name for a single target.
+    -o <dir>                   Output assets or source files to <dir>.
+    -m <manifest>              Perform conversions specified in <manifest>.
+    -g <manifest>              Generate source code based on <manifest>.
+    -i <assetPath>             Print details about an existing asset file.
+    -r <namespace>             Specify root namespace to use for generated source code.
+    -f <texture format>        For textures built in single target mode, specify texture
+                               format (default RGB8_UNORM_SRGB).
```
jaremieromer: Typo, should be `(default RGBA8_UNORM_SRGB)`

source/ncconvert/converters/TextureConverter.cpp
```diff
@@ -55,40 +112,21 @@ auto TextureConverter::ImportCubeMap(const std::filesystem::path& path) -> asset
...
+    if (generateMips)
     {
-        halfWidth = std::max(halfWidth / 2, minDimension);
-        halfHeight = std::max(halfHeight / 2, minDimension);
-        image.Resize(halfWidth, halfHeight);
-        texture.mipmaps.push_back(image.MakeTextureSubResource());
+        texture.mipmaps = GenerateMips(image, format, useCompression);
     }
 
     return texture;
```
jaremieromer, line 132: Very minor. Typo in closing namespace comment

jaremieromer (approved):
Curious what you're seeing with the normals - I ran it and it looks great.
This is excellent work

PR #927, opened 2025-04-02
Update Skybox Position

McCallisterRomer:
Fix skybox not tracking with the camera's position.


resources/shaders/source/Skybox.vsh
```diff
@@ -24,7 +23,8 @@ static float4x4 modelMatrix = float4x4(
 
 void main(in  VSInput VSIn, uint InstanceID : SV_InstanceID,  out PSInput PSIn)
 {
-    float4 TransformedPos = mul(float4(VSIn.Pos, 1.0), modelMatrix);
+    float4 TransformedPos = mul(float4(VSIn.Pos, 1.0), localModelMatrix);
+    TransformedPos += float4(cameraPosition, 0.0);
```
jaremieromer, line 27: Oh. you ADD it. I labored over this for a long time. NICE!

jaremieromer: You'd think linear algebra would sink in about now lol

jaremieromer (approved):

Issue #898, opened 2025-02-27
Failed to submit buffer to the command queue bug


jaremieromer:
![Image](https://github.com/user-attachments/assets/50b368c0-2e57-47b7-8e20-011fe58f4658)

Occasionally, we encounter a crash on submission of the queue. 

No steps to reproduce yet - its intermittent.

McCallisterRomer:
@jaremieromer I have only encountered this the one time. Have you seen it again?

jaremieromer:
@McCallisterRomer  I haven't seen it in quite a while, but 5-6 times total. Can you add the validation warning/error that vulkan produces to this ticket? I haven't updated Vulkan yet - but I can add it too when I do

McCallisterRomer:
VUID-vkAcquireNextImageKHR-semaphore-01779
                 vkAcquireNextImageKHR(): Semaphore must not have any pending operations.
The Vulkan spec states: If semaphore is not VK_NULL_HANDLE, it must not have any uncompleted signal or wait operations pending (https://vulkan.lunarg.com/doc/view/1.4.309.0/windows/antora/spec/latest/chapters/VK_KHR_surface/wsi.html#VUID-vkAcquireNextImageKHR-semaphore-01779)
                 Object[0] (semaphore): Handle 0x980f360000000011, Name: 'Swap chain image acquired semaphore 1'

PR #931, opened 2025-04-04
Update Diligent

McCallisterRomer:
presumably a fix for #898

Updating diligent to get a fix for the synchronization validation error. Also, had to do a simple to our imgui fork to make the sdl (1) backend header available. The version of imgui we're on has renamed it to sdl2, while diligent still tries to include the old name.


jaremieromer (approved):
Thank you so much for doing this!

PR #935, opened 2025-05-02
Added interface to specify the extents of a scene.

jaremieromer:
Added interface to specify the extents of a scene.
This is currently only read from by `LightSubsystem` which uses the information to calculate projection matrix bounds for lights. 
Directional lights look terrible now - as they were previously hacked to use a hard coded, smaller zNear and zFar and area for the projection matrix. 
I'll be picking up cascaded shadow mapping next to address this.


include/ncengine/scene/Scene.h
```diff
@@ -17,8 +18,29 @@ class Registry;
 class Scene
 {
     public:
+        Scene(Vector3 extents = Vector3{150.0f, 150.0f, 150.0f})
+        : m_extents{Vector3{std::max(1.0f, extents.x), std::max(1.0f, extents.y), std::max(1.0f, extents.z)}}
```
McCallisterRomer: I don't know if we decision off the minimum value elsewhere, but if so, it would be nice to have a `g_minSceneExtent` constant.

sample/source/shared/SampleUI.cpp
```diff
@@ -126,7 +126,7 @@ namespace nc::sample
 
             if (ImGui::Button("GraphicsTest", buttonSize))
             {
-                m_ncScene->Queue(std::make_unique<GraphicsTest>(this));
+                m_ncScene->Queue(std::make_unique<GraphicsTest>(this, nc::Vector3{20.0f, 20.0f, 40.0f}));
```
McCallisterRomer: It makes more sense to me for this to be a constant defined in `GraphicsTest` (either the object or just in the cpp). `SampleUI` can't really make an informed decision here.

source/ncengine/graphics2/frontend/GraphicsFrontend.h
```diff
@@ -68,16 +69,18 @@ class GraphicsFrontend
               m_particleSystem{
                 world,
                 maxParticles
-              }
+              },
+              m_modules{modules}
```
McCallisterRomer: I think this is leftover. Can delete?

test/ncengine/scene/NcScene_unit_tests.cpp
```diff
@@ -18,6 +18,22 @@ struct TestScene : public nc::Scene
     }
 };
 
+struct TestSceneWithExtents : public nc::Scene
```
McCallisterRomer: Can delete

McCallisterRomer (approved):

PR #943, opened 2025-08-12
Support for images in ImGui

jaremieromer:
Took your advice @McCallisterRomer and instead added a method to `NcGraphics` to return the `IDeviceObject*` as a void ptr, to allow callers to do with it as they will.

It works great! We can render texture assets, shadow maps, the normal buffer, the depth buffer, the post process effect-applied scene, etc.

image


source/ncengine/graphics2/NcGraphicsImpl2.cpp
```diff
@@ -435,6 +436,47 @@ void NcGraphicsImpl2::SetPostProcessEffectProperties(PostProcessEffectId effectI
...
 }
 
+auto NcGraphicsImpl2::GetTextureView(TextureViewType type, uint32_t index) -> void*
+{
+    switch(type)
+    {
+        case TextureViewType::Asset:
+        {
+            return static_cast<void*>(m_shaderBindings.GetPerFrameSignature().GetTextureBuffer().GetTextureView(index));
+            break;
```
McCallisterRomer: They're not hurting anything, but you don't need a return + break statement.

McCallisterRomer (approved):

PR #945, opened 2025-08-14
SampleUI Bug

### LINKED ISSUE #939: Accessing SampleUI in scene unload crashes instead of logging nicely
The order of the creation of variables on the stack in SampleMain is:
```
auto config = nc::config::Load(args.configPath);
auto engine = nc::InitializeNcEngine(config);
auto ui = nc::sample::InitializeSampleUI(engine.get());
```
So on exception throw in engine->Start, the stack unwinds and the ui gets destroyed before the engine.
So in engine teardown, m_sampleUI is already destroyed at this point:
```
void PhysicsTest::Unload()
{
    m_sampleUI->SetWidgetCallback(std::function<void()>{});
    CharacterEntities.clear();
}
```

I would imagine a shared_ptr would fix this

McCallisterRomer:
resolves #939

Exceptions can cause scenes to call into SampleUI after its lifetime has ended. Moved it so its guaranteed to outlive the engine.

### DIFF
--- a/sample/source/SampleMain.cpp
+++ b/sample/source/SampleMain.cpp
@@ -54,6 +54,7 @@ int main(int argc, char** argv)
     try
     {
         auto config = nc::config::Load(args.configPath);
+        auto ui = std::unique_ptr<nc::sample::SampleUI>{};
         auto engine = nc::InitializeNcEngine(config);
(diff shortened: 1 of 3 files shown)


jaremieromer (approved):
That was a clever fix!

PR #946, opened 2025-08-19
Update Audio Device Handling + Buffers

### LINKED ISSUE #374: Output channel testing
Verify we're well-behaved when using audio devices with various output channels. The current expectation is that we should have channel compensation when a device supports (or requires) more channels than we write, but we'll encounter a critical error for the inverse case.

### LINKED ISSUE #375: Handle dynamic buffer size + sample rate
Buffer size in `NcAudioImpl` needs to be be based on the actual buffer frames reported when opening a stream. This may need to be recreated on device change.

The call to `openStream` needs to consider what is supported and preferred by the device.

Its probably worth adding (preferred) options for both of these to the config.

### LINKED ISSUE #376: Handle audio device failure and disconnect

### LINKED ISSUE #378: Update audio buffer handling
We want to use something more like a ring buffer/spsc queue for the buffer that sits between NcAudio and the driver callback. Goals would be:
- Change mutex locking to atomic ops in audio thread (Haven't timed anything, just assuming this is faster. Might be worth looking into what is conventional).
- Missing samples in an underrun would be minimized to the number of samples we're actually missing, rather than an entire buffer.
- The thread running NcAudio would be able to do work based on how many samples are available for writing rather than skipping execution if we don't have a stale buffer.

McCallisterRomer:
resolves #374, resolves #375, resolves #376, resolves #378

- Changing audio buffers to use a more typical ring buffer + lock-free queue.
- Fixed bug where buffer size wasn't updated on device change.
- Adding a recovery attempt on `DeviceStream` failure. If recovery fails, we try to fallback to another valid device. If this fails, audio support is abandoned (and NcAudio trivially returns).
  - There are cases where a device can fail and not be immediately recoverable, but may become recoverable after a short delay. The engine defers to the game here by sending a 'Null' device via the `OnOutputDeviceChanged` signal. Client applications are free to attempt additional recovery attempts as/when they see fit.
    - These cases may be rare. not sure yet. Tweaking settings directly through my device's control application causes Windows to lose my device temporarily. Tweaking the same settings through Windows is well-behaved.

### DIFF
--- a/include/ncutility/SpscQueue.h
+++ b/include/ncutility/SpscQueue.h
@@ -4,16 +4,10 @@
  */
 #pragma once
 
-#include "ncutility/platform/Platform.h"
+#include "ncutility/platform/New.h"
 
 #include <array>
 #include <atomic>
-#include <new>
-
-#ifndef __cpp_lib_hardware_interference_size
-#error "Destructive interference not defined"
-#endif
-
 
 namespace nc
 {
@@ -83,7 +77,7 @@ class spsc_queue
 
         static_assert(N >= 1, "Queue capacity must be >= 1");
         static_assert(std::atomic<size_t>::is_always_lock_free);
-        static constexpr auto CacheLine = std::hardware_destructive_interference_size;
+        static constexpr auto CacheLine = nc::hardware_destructive_interference_size;
 
         alignas(CacheLine) std::array<T, N> m_buffer = {};
         alignas(CacheLine) std::atomic<size_t> m_head = 0;
--- a/include/ncutility/SpscQueue.h
+++ b/include/ncutility/SpscQueue.h
@@ -1,0 +1,110 @@
+/**
+ * @file SpscQueue.h
+ * @copyright Jaremie Romer and McCallister Romer 2025
+ */
+#pragma once
+
+#include "ncutility/platform/Platform.h"
+
+#include <array>
+#include <atomic>
+#include <new>
+
+namespace nc
+{
+NC_DISABLE_WARNING_PUSH
+NC_DISABLE_WARNING_MSVC(4324)
+
+/** @brief Fixed-size single-producer/single-consumer queue. */
+template <class T, size_t N>
+class spsc_queue
+{
+    public:
+        [[nodiscard]] auto push(const T& v) -> bool
+        {
+            return emplace(v);
+        }
+
+        [[nodiscard]] auto push(T&& v) -> bool
+        {
+            return emplace(std::move(v));
+        }
+
+        [[nodiscard]] auto pop(T& out) -> bool
+        {
+            if (is_empty(m_consumer))
+            {
+                refresh_consumer_cache();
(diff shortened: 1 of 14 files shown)


source/ncengine/audio/DeviceStream.cpp
```diff
@@ -153,6 +153,12 @@ void DeviceStream::CloseStream() noexcept
     }
 }
 
+void DeviceStream::AbandomStream() noexcept
```
jaremieromer: Typo in method name

source/ncengine/audio/NcAudioImpl.cpp
```diff
@@ -277,5 +204,51 @@ void NcAudioImpl::MixToBuffer(double* buffer)
...
+    }
+
+    m_outputDeviceChanged.Emit(m_deviceStream.GetDevice());
+}
+
+auto NcAudioImpl::TryFailureRecovery() -> bool
+{
+    const auto lastKnownDevice = m_deviceStream.GetDevice().id;
+    const auto preferredDevice = lastKnownDevice == InvalidAudioDeviceId
+        ? DefaultAudioDeviceId
```
jaremieromer, line 233: Isn't the typical case going to be that the lastKnownDevice is the DefaultAudioDeviceId, so this failure recovery would be a no-op?

include/ncutility/SpscQueue.h
```diff
@@ -0,0 +1,109 @@
...
+            if (is_full(m_producer))
+            {
+                refresh_producer_cache();
+                if (is_full(m_producer))
+                {
+                    return false;
+                }
+            }
+
+            m_buffer[index(m_producer.tail)] = T(std::forward<Args>(args)...);
```
jaremieromer, line 60: What happens to the T that was in `m_buffer[index(m_producer.tail)]` here? Is this guaranteed to only run if `m_buffer[index(m_producer.tail)]` is empty so it doesn't overwrite the previous T?

jaremieromer (approved):
From what I can understand, it looks good! Just a typo in a method name and a question about a potential leak - though it may not actually be a problem.

source/ncengine/audio/DeviceStream.cpp
McCallisterRomer: nice

source/ncengine/audio/NcAudioImpl.cpp
McCallisterRomer: Yes, the typical case is DefaultAudioDeviceId, but that's ok. We'd just attempt to reopen the stream on whatever device window's says we should use; not a no-op (line 237).

include/ncutility/SpscQueue.h
McCallisterRomer: The T that was previously here must have been already popped before we got here. AKA `pop()` is guaranteed to move from this buffer location before we can write back to it.

PR #962, opened 2026-08-17
fix quats + tests

### LINKED ISSUE #336: [NcCommon] Quaternion unit tests
Add unit tests for Quaternion.

McCallisterRomer:
resolves #336

- In editor, we now cache euler angles on object selection so we aren't converting to/from quaternions repeatedly.
- `Quaternion` `ToEulerAngles()` and `FromEulerAngles()` were using different rotation conventions. Both are valid, but incompatible Anything using 'to' AND 'from' was certainly broken. Its possible anything using the 'to' was also broken b/c that should be incompatible with other DX stuff, but there is no other usage in the engine I see. Anyways, updated to operate in DX's terms (roll -> pitch -> yaw, or Z -> X -> Y).
- Added quaternion tests
- General cleanup of Quaternion.cpp

### DIFF
--- a/include/ncengine/ui/editor/EditorContext.h
+++ b/include/ncengine/ui/editor/EditorContext.h
@@ -48,6 +48,7 @@ struct EditorContext
 
     // Mutable state (for internal use)
     Entity selectedEntity = Entity::Null();
+    Vector3 eulerRotation = Vector3::Zero();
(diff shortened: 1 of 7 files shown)


jaremieromer:
magnifico!! Works great, tysm!!

jaremieromer (approved):

PR #963, opened 2026-09-03
Added a BoneSnapper component

jaremieromer:
Added a new component `BoneSnapper` that ties the transform of an entity to that of a bone in a SkinnedMesh.


source/ncengine/graphics2/frontend/subsystem/animation/BoneSnapperSubsystem.cpp
```diff
@@ -0,0 +1,44 @@
...
+        if (!m_skeletalAnimationSubsystem.ContainsBone(skinnedMesh.GetMeshId(), boneSnapper.boneName))
+        {
+            continue;
+        }
+
+        const auto& targetBoneTransform = m_skeletalAnimationSubsystem.GetBoneSnapperOffset(boneSnapper.target);
+        auto& sourceTransform = ecs.Get<Transform>(sourceEntity);
+        const auto sourceLocalScale = sourceTransform.LocalScale();
+
+        sourceTransform.SetTransformationMatrix(DirectX::XMMatrixIdentity());
```
McCallisterRomer: This line shouldn't have to be here because its overwritten on line 40. Also, in line 40, `sourceTransform.LocalTransformationMatrix()` is just identity after this call. So the transform becomes the bone, but retaining its old scale. Is that correct?

McCallisterRomer (approved):

source/ncengine/graphics2/frontend/subsystem/animation/BoneSnapperSubsystem.cpp
jaremieromer: Oh, you're totally right. It can just be:

```
        sourceTransform.SetTransformationMatrix(targetBoneTransform);
        sourceTransform.SetScale(sourceLocalScale);
```

